PR #7, opened 2019-10-10
Ban smart pointers, ban #define

andrewbriand:


save-buffer:
You can still use #define lol. sometimes there's a legitimate need for a macro. You can change it to something like "prefer constexpr for constants".

PR #12, opened 2019-11-13
Create first working version of AR Tag detection

evan3334:
Detection is somewhat finicky but working. More improvements to the process will be coming soon but the main parts of it are in place.

**Note that once this is merged, all team members will need to have OpenCV installed either in WSL if they're using that, or directly on their computers if they're not using WSL.**


README.md
```diff
@@ -1,31 +1,93 @@
...
+
+### On Mac
+There is a Homebrew package available for OpenCV on Mac OS. Open the Terminal
+and run these commands:
+1. `/usr/bin/ruby -e "$(curl -fsSL
+   https://raw.githubusercontent.com/Homebrew/install/master/install)"` (If you
+   already have Homebrew installed, you can skip this step.)
+2. `brew install opencv@3`
+
+## Install Catch2 (for unit testing)
```
evilnose: I thought we were using the single header version of catch2 for easy setup?

src/ar/ARTester.cpp
```diff
@@ -0,0 +1,94 @@
+#include "Detector.h"
+
+#include <opencv2/core.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+
+#include <iostream>
+
+using namespace std;
```
evilnose: This is bad practice. Also even though this is used, there is still a mix of "cout" and "std::cout" below.

src/ar/Detector.cpp
```diff
@@ -0,0 +1,92 @@
+#include "Detector.h"
+
+namespace AR
+{
+	const int CONTOUR_AREA_THRESH = 1000;
```
evilnose: Inconsistent indentation here and below

src/ar/Tag.h
```diff
@@ -0,0 +1,131 @@
...
+	/*
+	  Camera parameters
+	  --------------------
+	  These are obtained from camera calibration and are unique to and constant for
+	  every camera. Change the macro above to use a different set of camera parameters.
+	*/
+
+#ifndef USE_CAMERA_PARAMS
+#error "USE_CAMERA_PARAMS macro is not defined"
+#elif USE_CAMERA_PARAMS == EVAN_LAPTOP_CAMERA_PARAMS
```
evilnose: File needs to be formatted

src/ar/Tag.cpp
```diff
@@ -0,0 +1,151 @@
...
+		else
+		{
+			x = atan2(-R.at<double>(1, 2), R.at<double>(1, 1));
+			y = atan2(-R.at<double>(2, 0), sy);
+			z = 0;
+		}
+		return cv::Vec3f(x, y, z);
+	}
+
+	const char* InvalidCornerException::what() const throw()
```
evilnose: We should use error codes

src/ar/Detector.cpp
```diff
@@ -0,0 +1,92 @@
...
+		{
+			std::vector<cv::Point2f> approx;
+			cv::approxPolyDP(contours[i], approx, arcLength(contours[i], true)*0.02, true);
+			if(approx.size() == 4
+			   && cv::isContourConvex(approx)
+			   && cv::contourArea(approx) > CONTOUR_AREA_THRESH)
+			{
+				//sort points into the correct order (top left, top right, bottom right, bottom
+				//  left)
+				struct sortY {
```
evilnose: define this outside of the loop and use lambdas instead

src/ar/Tag.cpp
evilnose: I will add stuff to slite about this later

README.md
evan3334: Unfortunately if we want the CMake integration we need to use the CMake package from their git repo.

src/ar/Tag.cpp
evan3334: Okay. Unsure about how that will work because this exception is thrown from a constructor, in the event that the corners are in the wrong order. In this case we thought it was better to fail rather than sort the coordinates into the correct order without telling the user.

Maybe we could have an error code parameter passed by reference or something.

evan3334:
I just added a commit that fixes most of the style issues, except for the error code one. Will fix that one sometime soon, as soon as I can find some documentation explaining it

src/ar/Tag.cpp
evilnose: Umm how do you guys plan to handle this exception when one is thrown? Currently this exception is not caught.

src/ar/Tag.cpp
```diff
@@ -0,0 +1,153 @@
...
+	double c_mag = sqrt((c.x * c.x) + (c.y * c.y));
+	return acos(dot_product / (a_mag * c_mag));
+}
+
+cv::Vec3d rotationMatrixToEulerAngles(cv::Mat &R)
+{
+	double sy = sqrt(R.at<double>(0, 0) * R.at<double>(0, 0) +
+	                 R.at<double>(1, 0) * R.at<double>(1, 0));
+
+	bool singular = sy < 1e-6;
```
evilnose: can we create an epsilon constant, or a function if applicable, for the fuzzy comparison

src/ar/Tag.cpp
```diff
@@ -0,0 +1,153 @@
...
+		corners.push_back(Corner{internal_angle, points[i]});
+	}
+	auto t1 = std::chrono::high_resolution_clock::now();
+	orientation = calcOrientation();
+	auto t2 = std::chrono::high_resolution_clock::now();
+	auto duration = std::chrono::duration_cast<std::chrono::microseconds>(t2 - t1).count();
+	std::cout << "calcOrientation took " << duration << "us" << std::endl;
+}
+
+cv::Point getTriCentr(cv::Point pt1, cv::Point pt2, cv::Point pt3)
```
evilnose: Abbreviations are fine but I feel that omitting only one 'e' doesn't shorten the name that much and may cause misspelling

PR #37, opened 2020-07-11
Merge lidar-mapping into master

nhatch:
I figure we should just use master as the main branch. Makes it easier for people to know where to look for things.


nhatch:
Huh, it seems I'm not able to review my own PR. Guess that makes sense.

The differences here are obviously way too big to review (which is another reason we should just develop off of master with small feature branches). So I'm just going to force-merge this.

PR #39, opened 2020-07-17
Pathing auto

assafvayner:
connects pathfinding files to autonomous class
obstacle

 (successfully avoids obstacle)


src/Autonomous.cpp
```diff
@@ -54,9 +86,40 @@ void Autonomous::autonomyIter() {
...
+        speed = 5.0;
+    }
+    setCmdVel(dtheta, speed);
+  }
+}
+
+double Autonomous::pathDirection(points_t lidar, pose_t gpsPose) {
+    double dtheta;
+    if(lidar.empty()) {
+        dtheta = angleToTarget(gpsPose);
```
nhatch, line 112: `angleToTarget` returns an angle difference, whereas `setCmdVel` expects an angular _rate_. I think in this case the rover will still behave correctly, since this setting basically says we should turn so that we reach the target angle in one second. So this is just an FYI.

src/Pathfinding/ObstacleMap.cpp
```diff
@@ -34,18 +20,19 @@ inline bool inBounds(int coordinate)
 }
 
 //rebuilds ObstacleMap with given Obstacles
-void ObstacleMap::update(std::vector<PointXY>& obstacles)
+void ObstacleMap::update(const std::vector<PointXY>& obstacles, float robotX, float robotY)
 {
-    updateRobotPosition();//updates values in robotX and robotY fields
+    this->robotX = robotX;
+    this->robotY = robotY;
```
nhatch, line 26: Minor quibble: it's slightly inconsistent that we use `this->robotX` here and just `robotX` below

src/Pathfinding/Pather2.cpp
```diff
@@ -31,76 +31,75 @@ void print_queue(std::queue<Pather2::queueNode> q) {
...
+                std::queue<PointXY> addAdjPointToPath = curr.path;
+                addAdjPointToPath.push(test);
+                Pather2::queueNode adjNode = {PointXY{(float)row, (float)col}, curr.dist + 1, addAdjPointToPath};
+                obsMap.obstacle_map[(int)adjNode.pt.x][(int)adjNode.pt.y] = true;
+                q.push(adjNode);
+            }
+        }
+        
+    }
+    return error;
```
nhatch, line 95: Oddly, nowhere in this method is `error` ever updated. So this function will always return (5,5). Seems like there are probably a lot of bugs in here that we should take a look at in a follow-up pull request. We should also take a look at PatherTest to figure out why it hasn't caught any of these bugs.

src/Autonomous.cpp
```diff
@@ -54,9 +86,40 @@ void Autonomous::autonomyIter() {
...
+    } else {
+        const std::vector<PointXY>& ref = points_tToPointXYs(lidar);
+        PointXY direction = pather.getPath(ref, (float) gpsPose(0), (float) gpsPose(1), target);
```
nhatch, line 115: Does `getPath` return a direction or a location? E.g. if the robot is at the point (1,1) and the next spot on the path is (2,1), then will `getPath` return (1,0) or (2,1)? If the latter, then we would need to subtract the current robot position from the return value of `getPath` in order to get a direction. But again, let's wait until a future diff to dig into the problems with the Pather2 class.

nhatch:
Ok I uploaded a video to Slack: https://huskyrobotics19-20.slack.com/archives/CNP0C7P9C/p1595549688222300

Current behavior is:

- Go straight up at 45 degrees until out of lidar range of obstacles
- Then speed up and go straight toward the current goal based on GPS estimate. Since the GPS estimate is quite noisy this involves a lot of zigzagging but we often do get close to the goal
- But then we drive right past it because we're not checking the "visual system" to figure out where the goal is
- Once we're facing away from the goal, we actually have trouble turning around to head back towards it. Due to noise in the heading estimate, the angle difference switches back and forth rapidly between pi and -pi, with the result that the rover never finishes turning around

Still, this is a huge improvement! There is some intelligence here! I think we should just merge it

PR #40, opened 2020-07-18
Get Visualizer.cpp to compile

nhatch:
I took over from Manasi to try out the pathing visualizer. This PR gets the visualizer to compile and fixes a few bugs, but there are still many exciting problems to solve. The current version:

* Doesn't visualize the goal
* Displays the obstacles as rotated 90 degrees clockwise relative to what they should be (as shown in the console output)
* Seems to have the wrong start and end points for the plan (sometimes it doesn't start on the robot, and it ends in a different spot than it should, assuming the goal and the obstacles are given in the same reference frame)

I thought I would just merge this one, which should make it easier for someone familiar with the pathing code to track down these remaining problems.

Relatively good case:
Screenshot from 2020-07-18 14-36-50

Something weird is going on:
Screenshot from 2020-07-18 14-37-01


nhatch:
I think we decided that they should be given in the map frame. I'll update this PR to remove those changes.

nhatch:
@assafvayner Updated.

nhatch:
@assafvayner I fixed the merge conflicts. Feel free to merge this if it looks good to you.

PR #44, opened 2020-07-24
Extract Eigen from repository

evan3334:
Noticed that we have a local checkout of Eigen tracked by Git. Ideally, I think this is probably something we want to have installed on people's systems through system package managers instead of tracking it in the source tree. 

This pull request adds Eigen3 as a dependency in CMake, and removes the local copy. It also replaces `#include` statements referring to the local copy to be system include statements that would reference the system-installed copy of Eigen. On Ubuntu, it's pretty easy to just use Eigen from system package managers, `sudo apt install libeigen3-dev` and then CMake will take it from there. 

Some future changes might be required in this; I'm mostly just opening this right now as a request for comments. Please let me know if there's something this breaks or if we should change documentation or something and we can try to fix that. Primarily, it would be nice if someone who has a Mac can test this to make sure that CMake can find Eigen properly if it's installed from Homebrew; and if not, what kind of steps we need to take to work around that.


nhatch:
A good first step might be to figure out how to get the CI tests to build. Looks like this was the error from the last run:

```
CMake Error at CMakeLists.txt:8 (find_package):
  Could not find a configuration file for package "Eigen3" that is compatible
  with requested version "3.3".

  The following configuration files were considered but not accepted:

    /usr/lib/cmake/eigen3/Eigen3Config.cmake, version: unknown
```

evan3334:
Okay, so I have a solution to fix the CI, it's a little convoluted, but it seems to work.

Our current CI testing system appears to be using Docker, and is based on an image called `spmallick/opencv-docker` which contains Ubuntu with OpenCV installed on top. A few problems arise from this:
1. The Ubuntu installed in this image is very old, and as such it has old versions of CMake that do not properly detect the Eigen installation.
2. The OpenCV version installed in this image is also old, and doesn't match the version we're building against, nor does it include the ARUco contrib module that future AR tag detection code is going to need.
3. It doesn't include any of the other dependencies of our codebase like Catch2 and the URG library, which means we have to spend time building them every time we run the CI, even though these things should change very infrequently.

What I've done is make a new Docker image, which is based off Ubuntu 18.04 LTS, and has OpenCV 3.4.7 built with the contrib module, Catch2, Eigen 3, SFML, and Andrew's URG lidar library installed. I've then changed the CI config to be based off this image instead of the `spmallick/opencv-docker` image, in commit d28e671. Instead of taking a really long time to build everything, all it needs to do is just pull in all the prebuilt stuff from Dockerhub, and then build our repository. The CI shows that all the tests have passed, so I think this solution _works_, it's just a question of whether or not this is a _good_ solution or kind of hacky.

**Future changes that are probably necessary:**
- We probably should host the Dockerfile for this image either in a subdirectory of our PY2020 repository, or as a separate repository under the huskyroboticsteam account, just to make sure that the team can edit it if necessary. I currently have it hosted at [evan3334/PY2020-test-deps](https://github.com/evan3334/PY2020-test-deps), but this isn't a good long-term place for it to live because only I have access to this.
- Similarly, we should probably host the image on a Dockerhub account that everyone, or at least others besides me can access; it currently lives on `estraw/rover-test-deps`, but likewise, no one but me has access to this, which isn't good if we need to change any of the dependencies for the CI process.

Please let me know what you think.

nhatch:
Good investigation. That sounds like a tricky problem to debug; I'm glad you got to the bottom of it.

I agree that we shouldn't base our CI builds off of an old / unmaintained docker image. A custom docker image that we maintain ourselves could be a great solution if either of the following is true:

* We're planning to deploy the exact same docker image on the rover itself during the competition
* Automatically installing our dependencies from a more "standard" docker image takes too long or is too complicated

The main alternative to this, I think, would be to use a more up-to-date image available elsewhere on the internet, e.g. https://hub.docker.com/_/ubuntu. The challenge there, it seems, is that OpenCV can only be installed from source, and it takes a very long time to compile. Installing our other dependencies is pretty fast, I think.

What about this docker image instead? https://hub.docker.com/r/jjanzic/docker-python3-opencv (I'm not sure if this is what we need or not; I just Google'd "opencv docker image".)

PR #45, opened 2020-08-01
Update to new simulator master

nhatch:
This new version of the simulator avoids the dependency on the `X11` library and does not try to launch windows from worker threads (which apparently is not allowed on OSX). However, I think it requires updating to SFML 2.5, which may require building from source.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,7 +117,7 @@ add_executable(visualizer_test
   )
 
 list(APPEND simulator_libs
-  sfml-graphics sfml-window sfml-system X11 pthread)
+  sfml-graphics sfml-window sfml-system pthread)
 
 target_link_libraries(lidar_vis liburg_c.a ${OpenCV_LIBS})
 target_link_libraries(lidarOnlyTest liburg_c.a ${OpenCV_LIBS})


benfrancis314:
TL;DR: Simulator program works and logs coordinates, but window doesn't show
----
Build was succesful (!), though for future Mac user reference I did have to tell it where sfml was by adding `-L/usr/local/lib` in: 

`list(APPEND simulator_libs
  sfml-graphics sfml-window sfml-system pthread -L/usr/local/lib)`


When running ./RoverSim now, I got the following log: 

`Base station connect failed: Connection refused
Cannot create a window from a worker thread. (OS X limitation)
arrived at gate
x: 3.02531 y: 2.88853 theta: 0.000248095
arrived at gate
x: 2.93905 y: 2.87511 theta: 0.526885
arrived at gate
x: 3.17812 y: 2.3167 theta: 0.363176
arrived at gate
x: 2.64729 y: 2.53623 theta: 0.226305`

So it appears it's just a problem with the window, as I would assume the `arrived at gate` and `x:...` outputs are from the simulator logs while it is running. 

The old error was: 
`Base station connect failed: Connection refused
Cannot create a window from a worker thread. (OS X limitation)`

So while there is still some funkiness, this is a strong improvement since now I can see the coordinates.

nhatch:
Darn, it seems like this didn't really do anything to address the no-window-visible problem. But it's good to remove the X11 dependency in any case.

I think if we eventually want the simulator window to be visible on OSX, we should get someone on OSX (@benfrancis314, if you're interested) to go through one of the SFML tutorials (https://www.sfml-dev.org/tutorials/2.5/) and just get a basic program working. Then I could go through the code for that and try to find out what are the differences between that and how the simulator currently works.

Another thing you could try is building SFML 2.5 from source.

But if the current setup is good enough to test your code, we can just stick with it.

benfrancis314:
Yes I am in! The current setup is probably at least good enough for me to experiment with the SFML tutorial and try to get the window up; it may be good enough to test Kalman filter, but it's a little hard for me to tell right now. 

This update did also seem to allow the program to run, whereas before I didn't get any output from it

PR #49, opened 2020-10-28
Use Compatible Catch2 Version

arnavthareja:
The default branch of the Catch2 GitHub repository is now on a version 3 preview which has a different file organization and naming scheme that causes errors when trying to find catch2/catch.hpp when building this repository (on both local machines and the Docker CI). This pull request ensures that we are using a compatible version of catch2 that includes a catch.hpp file by checking out to the latest release on the v2.x branch of the Catch2 GitHub repository. The README and docker entrypoint have been updated to include this extra step.

This could be a temporary fix if we plan to move to v3 of Catch2, as that would require some more effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,9 @@ CMake, or Homebrew for Mac users.
 Open the WSL terminal and run the following commands:
 1. `git clone https://github.com/catchorg/Catch2.git`
 2. `cd Catch2`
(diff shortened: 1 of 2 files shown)


nhatch:
Probably a good idea to update to v3 at some point soon, but this is a good fix for now.

PR #50, opened 2020-11-01
Navigate to target without obstacles

abhaybd:
This also adds a Kalman Filter implementation, along with some other code for state estimation.


abhaybd:
The CI is still broken, so I think #49 needs to be merged first.

nhatch:
I haven't looked through all the code yet, but I tried building this branch and running it. How have you been testing this code? I used the `RoverSim` executable, which is what we've used in the past. FYI, I'm not seeing it go towards the first post. I'll take a closer look at the code tomorrow and try to figure out what's going on.

abhaybd:
In `Rover.cpp`, the target is being set to `(3.14,2.17)`, which is very close to the robot starting position. I assume these were test values since those are roughly pi and e, respectively. Changing those values should cause the robot to navigate to different places.

abhaybd:
I realized there's a weird bug, so this branch probably shouldn't be merged before I figure that out.

I can't figure out where the problem is, but for some reason both the linear and extended kalman filters move the state estimation way too much during the predict step, which causes the state estimate to be too far ahead of the robot.

Weirdly enough, I dropped the files into the coding challenge project and it works perfectly fine over there, so I'm not sure what's going on.

I'll spend some time debugging it and get back to you.

nhatch:
@abhaybd What might be happening is there's no rate-limiting in the PY2020 control loop like there is in the code challenge---the control loop runs just as fast as the processor can execute it, rather than sleeping each loop so it only runs at like 30 Hz. I'll write some code to add rate limiting

src/filters/KalmanFilter.h
```diff
@@ -0,0 +1,278 @@
...
+			   const Eigen::Matrix<double, numStates, numStates> &outputMat,
+			   const Eigen::Matrix<double, numStates, 1> &stateStdDevs,
+			   const Eigen::Matrix<double, numStates, 1> &measurementStdDevs, double dt)
+	{
+		matrix stateCovarianceCont = createCovarianceMatrix(stateStdDevs);
+		matrix measurementCovarianceCont = createCovarianceMatrix(measurementStdDevs);
+
+		matrix contA = systemMat;
+		matrix contB = inputMat;
+		discreteToContinuous(contA, contB, dt);
```
nhatch, line 82: I'm a little confused why we have one constructor that takes a discrete system and makes it continuous, and another that goes the other direction, but no constructor that takes a discrete system and _keeps_ it discrete.

src/filters/PoseEstimator.h
```diff
@@ -0,0 +1,83 @@
...
+		reset(Eigen::Vector3d::Zero());
+	}
+
+	/**
+	 * Sets the state estimate to the supplied vector and resets the estimate covariance
+	 * matrix.
+	 *
+	 * @param pose The pose to which the state estimate will be set.
+	 */
+	void reset(const pose_t &pose);
```
nhatch, line 68: It seems odd to have some of these functions implemented in the `.h` file, but others implemented in the `.cpp` file. I will admit I'm not completely an expert in C++ so maybe there's a good reason for this, but would you mind explaining it to me? By default, I think we should stick with the standard header/implementation between `.h` and `.cpp`.

src/Autonomous.cpp
```diff
@@ -1,227 +1,308 @@
...
+double dist(const PointXY &p1, const PointXY &p2)
+{
+	return std::hypot(p1.x - p2.x, p1.y - p2.y);
+}
+
+void filterLandmarks(points_t &landmarks)
+{
+	landmarks.erase(std::remove_if(landmarks.begin(), landmarks.end(),
+								   [](point_t p) { return p[2] == 0; }),
+					landmarks.end());
```
nhatch, line 109: We might not want to do this, because the index in the array indicates which landmark it is. So e.g. if only one of the landmarks is visible, we'll erase all the others and then we won't know which landmark we're looking at.

src/Autonomous.cpp
```diff
@@ -1,227 +1,308 @@
...
-        dtheta = angleToTarget(gpsPose);
-    } else {
-        const std::vector<PointXY>& ref = points_tToPointXYs(lidar);
-        PointXY direction = pather.getPath(ref, (float) gpsPose(0), (float) gpsPose(1), target);
-        float theta = atan2(direction.y, direction.x);
-        dtheta = static_cast<double>(theta) - gpsPose(2);
-    }
-    return dtheta;
+	if (!Globals::AUTONOMOUS)
+		return;
```
nhatch, line 115: If we're going to put this on a separate line, I'd suggest adding `{}` around it. Otherwise it's easy to mistakenly add extra indented lines here without realizing that only the first one is conditioned.

src/filters/PoseEstimator.h
abhaybd: There wasn't really a good reason for this, I'll move all the method implementations to the cpp file.

src/filters/KalmanFilter.h
abhaybd: The `createDiscrete()` method only stores the passed in discrete system and input matrices. The reason that we need to convert the system matrix into continuous form temporarily is because the passed in process covariance matrix is continuous. Discretizing the process covariance matrix requires a continuous system matrix, which requires us to temporarily convert the system matrix into continuous form.

src/filters/KalmanFilter.h
```diff
@@ -0,0 +1,278 @@
+#pragma once
+
+#include <Eigen/Core>
+#include <Eigen/LU>
+#include <unsupported/Eigen/MatrixFunctions>
```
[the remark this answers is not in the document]
abhaybd: The unsupported refers to the fact that it's a "contribution from various users", but it still ships with the Eigen installation, so no additional configuration is required. It's used because in order to convert the system from continuous to discrete and back, we need to take the matrix logarithm and exponential, which are implemented in the `MatrixFunctions` module. Is that ok?

src/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@ list(APPEND RoverCommonTest
   Autonomous.cpp
   Util.cpp
   simulator/utils.cpp
+  filters/PoseEstimator.cpp
+  filters/PoseEstimator.h
+  filters/RollingAvgFilter.h
+  filters/KalmanFilter.h
```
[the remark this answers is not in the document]
abhaybd: Ok, I'll fix that. I'm not too familiar with cmake so I just put everything into the makelists file, sorry!

As part of a bug fix I modified RollingAvgFilter to use templates so it's more generic. In that case, would it be ok to leave the implementation in the header file? I'm not super familiar with the build system so I could be wrong but don't templates require implementation in the header files?

src/Autonomous.cpp
abhaybd: Oh, I didn't realize that's how the landmarks worked, I'll fix that. I assume the landmark for the first leg is at index 0, right? Is there one landmark per leg?

abhaybd:
> @abhaybd What might be happening is there's no rate-limiting in the PY2020 control loop like there is in the code challenge---the control loop runs just as fast as the processor can execute it, rather than sleeping each loop so it only runs at like 30 Hz. I'll write some code to add rate limiting

@nhatch I definitely missed that! That helped solve most of the problem, but the estimation still moves ahead too fast. (although not as much, and it settles a few meters ahead of the robot) I double and triple checked that the same code works perfectly fine in the coding challenge project, so I suspect my bug is somewhere in `Autonomous.cpp`.

abhaybd:
Solved it! Turns out it wasn't a bug in any of the autonomous or filter code. In `world.cpp` the `spinSim()` method always sleeps for 1/30 seconds, but it assumes that the code runs in 0ms, which is not the case. This caused the average loop rate to be less than 30Hz, so the simulation wasn't updating fast enough, which caused the robot to move slower than the model predicted it should, since its position wasn't getting updated frequently enough.

By measuring the elapsed time and subtracting that from the time to sleep this bug got fixed. I saw that in the coding challenge there was code to do this but that code is missing from the simulation code in this repo. Can we get that fixed in the simulator repo?

src/CMakeLists.txt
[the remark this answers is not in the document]
nhatch: Actually no, templates are "templates" in the sense that they can't be compiled without specifying all the template arguments. So I believe they're compiled separately for every `.cpp` file that includes them.

That said, I've made mistakes on the template compilation process many, many times and I could be wrong here. I'd say just try to compile this without putting the .h files here. If for some reason that doesn't work, you can put them back I guess.

src/Autonomous.cpp
nhatch: No, some legs have gates instead of posts, which means there are two landmarks instead of one. Best just to think about landmarks as AR tags, and the mapping from legs to AR tags is done using the `URCLeg` objects and the `getLeg()` method.

nhatch:
Ok, #52 makes these rate-limit changes. Feel free to merge that one (#52) if it looks good to you.

src/CMakeLists.txt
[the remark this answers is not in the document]
nhatch: It looks like you already did this, so I'll mark this as resolved.

src/Autonomous.cpp
nhatch: But anyway, doing this properly would require passing in the URCLeg object instead of just a PointXY, which is exactly what #51 does. So we'll deal with it in #51 after this is merged. The updated code looks good.

nhatch:
And it looks like all my comments have been addressed in this PR; great! So this is good to merge as soon as #52 is merged.

abhaybd:
@nhatch I added a little tidbit to accept the control frequency as a parameter, but it's pretty small so I'm not sure if it requires a review.

Additionally, for larger PRs that have lots of commits (this one has 29, for example) should I do a regular merge, or a squash & merge? Is there one that's preferred?

nhatch:
Good question. I generally like squash and merge, unless the commit history on the branch is particularly clean

abhaybd:
I noticed changing the signature on the constructor broke the build due to tests, so I'm fixing the test cases to add a control_hz parameter.

PR #51, opened 2020-11-07
Control loop rate limit 10 Hz; switch target from PointXY -> URCLeg

nhatch:
Using URCLeg instead is important, because we also need to know when the
AR tag is visible to the cameras.

### DIFF
--- a/src/Rover.cpp
+++ b/src/Rover.cpp
@@ -14,6 +14,8 @@ void InitializeRover()
     InitializeBaseStationSocket();
 }
 
+const double CONTROL_HZ = 10.0;
+
 int main(int argc, char **argv)
 {
     world_interface_init();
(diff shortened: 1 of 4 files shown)


nhatch:
This is going to cause merge conflicts with #50 actually, so let's wait to merge this until #50 is in.

src/Rover.cpp
```diff
@@ -35,6 +37,7 @@ int main(int argc, char **argv)
             ParseBaseStationPacket(buffer);
         }
         autonomous.autonomyIter();
+        usleep(1000 * 1000 / CONTROL_HZ);
```
abhaybd, line 40: Instead of assuming that the code executes in 0 time, do you think we can measure elapsed time and subtract that from the time to sleep? The controllers and filters would probably perform better with a more accurate dt.

nhatch: Yeah, that's a good idea. I always write it this way first because I'm lazy, but then I always end up having to change it later

nhatch:
I made a new PR instead of rebasing this one

PR #53, opened 2020-11-13
Fixed bug in KF math

abhaybd:
I was looking over my math and I realized I had an extra set of parenthesis that changed the meaning of the equation.

It should be solving the equation S^T * K^T = A * P^T for the matrix K, but it was instead solving S^T * K = (A * P^T)^T for K, which is incorrect.


nhatch:
I'll take your word for it :) Did this change affect performance at all?

abhaybd:
By coincidence the matrices were all symmetric and/or diagonal, so the equations worked out to the same thing. In more complicated scenarios with different matrices, it would result in undesired behavior.

PR #55, opened 2020-11-28
Replace PointXY with pose_t

nhatch:
We want our navigation goals to include orientation, so we need to
specify them as a pose_t instead of PointXY.

This commit also removes a lot of code from the previous attempt at
autonomous navigation, which I don't really understand and which didn't
quite seem to work.

This commit ALSO makes "calibrated" default to true, so we skip the
calibration phase at the beginning of navigation. (This saves time, and
it doesn't seem to harm navigation performance.)


src/Autonomous.cpp
```diff
@@ -87,20 +79,21 @@ double transformAngle(double currAngle, double targetAngle)
 	return currAngle + ((absDist > (range / 2)) ? -sign * (range - absDist) : dist);
 }
 
-double getThetaVel(const PointXY &target, const pose_t &pose, double &thetaErr)
+double getThetaVel(const point_t &target, const pose_t &pose, double &thetaErr)
 {
-	double dx = target.x - pose[0];
-	double dy = target.y - pose[1];
+	double dx = target(0) - pose(0);
```
nhatch, line 84: I think with Eigen the "correct" way to access vector elements is with parentheses rather than brackets. Not sure why it worked before

src/Autonomous.cpp
```diff
@@ -142,34 +135,32 @@ void Autonomous::autonomyIter()
...
 		std::cout << "x: " << pose(0) << " y: " << pose(1) << " theta: " << pose(2)
-				  << std::endl;
+					<< std::endl;
 		landmarkFilter.reset(); // clear the cached data points
 		setCmdVel(0, 0);
 	}
 	else
 	{
-		PointXY driveTarget = getTarget();
-		if (dist(driveTarget, point_tToPointXY(pose)) < 25)
```
nhatch, line 152: I removed this check. If we see the AR tag from farther away than 25 meters, we still might as well use that as our navigation target.

src/Autonomous.h
```diff
@@ -21,38 +20,26 @@ class Autonomous
...
+	pose_t getTarget();
 	void autonomyIter();
 
 private:
 	URCLeg target;
 	PoseEstimator poseEstimator;
-	RollingAvgFilter<5,3> landmarkFilter;
-	bool calibrated = false;
-	std::vector<pose_t> calibrationPoses{};
+	int state;	// 1 is move forwards, 0 is turning, -1 is back up
```
nhatch, line 30: I reordered these private variables to match the order used by the constructor

abhaybd:
@nhatch I think defaulting `calibrated` to true works by coincidence, since the starting pose of the robot happens to be (0,0,0). If that weren't true, then it would take a while for the pose estimator to converge at the correct position.

Especially when the pose estimator algorithm gets switched to an EKF, since EKFs are a little more sensitive to initialization, this approach might not work. If you want the calibration to be faster, maybe the number of poses used for calibration could be drastically reduced instead. Reducing it from 62 to about 5 (line 54 of Autonomous.cpp) should get around 2m of accuracy for initialization, and it should only take half a second instead of 6 seconds. After initialization, the pose estimator should converge on the true position fairly quickly.

I've tested this with my EKF implementation, and initializing with only 5 samples yields decent results.

src/Autonomous.cpp
```diff
@@ -183,127 +174,21 @@ void Autonomous::autonomyIter()
 			poseEstimator.predict(thetaVel, driveSpeed);
 
 			std::cout << "ThetaVel: " << thetaVel << " DriveVel: " << driveSpeed
-					  << " thetaErr: " << thetaErr << " targetX: " << driveTarget.x
-					  << " targetY: " << driveTarget.y << std::endl;
+						<< " thetaErr: " << thetaErr << " targetX: " << driveTarget(0)
+						<< " targetY: " << driveTarget(0) << std::endl;
```
abhaybd, line 178: I think the second `driveTarget(0)` is supposed to be `driveTarget(1)`, right?

src/Autonomous.cpp
abhaybd: `point_t` is an `Eigen::Vector3d`, and Eigen supports `operator[]` for vectors. However, I agree that using parenthesis instead of brackets is better for consistency.

nhatch:
Alright, I switched it to calibrate for half a second. Thanks for the feedback.

While we're on the subject, one thing that concerns me about this "calibration" procedure is that I'm not sure that's how GPS readings will work on the real robot. I suspect there won't be much difference between a single GPS reading and five GPS readings spaced over half a second, in terms of the accuracy of our initial pose estimate. Still, for the time being this simulator is all we have, so calibration makes sense for now.

abhaybd:
That makes sense, since calibrating five times on the same data doesn't really do anything. A possible fix would be to calibrate with only a single data point. That way, it should happen pretty quickly and it should work in real life. I'm including a fix in my EKF branch since I'm redoing most of the pose estimation anyway, but the new calibration procedure should be able to accept inaccurate calibration as long as the inaccuracy is quantified.

The current implementation assumes perfect calibration, but the new one should be able to calibrate decently with as little as one data point. Obviously the performance would be better with more data, but it seems to run okay in the sim with just one sample.

Additionally, could the simulator be changed to reflect the irl update rate of the gps? If the gps updates slower than the robot loop, then the pose estimation code should take that into account, since it won't have fresh data every iteration.

nhatch:
Yeah, once we know more about the real GPS sensor we'll be using, we will probably want to update the simulator to better match that behavior. In the meantime, I think the current behavior / calibration procedure is fine.

src/Autonomous.cpp
```diff
@@ -142,34 +135,32 @@ void Autonomous::autonomyIter()
...
+				driveTarget = landmarkFilter.get();
+			}
+			else
+			{
+				// transform and add the new data to the filter
+				transform_t invTransform = toTransform(pose).inverse();
+				point_t landmarkMapSpace = invTransform * leftPostLandmark;
+				// the filtering is done on the target in map space to reduce any phase lag
+				// caused by filtering
+				driveTarget = landmarkFilter.get(landmarkMapSpace);
```
abhaybd, line 163: The `landmarkFilter` filters on `point_t`s, which means it has data of the form `[x, y, 1]`. However, `driveTarget` is a `pose_t`, which means it has data of the form `[x, y, theta]`. If we set output of the filter to the drive target, then the target theta will be misinterpreted to be 1rad, which may not be the case.

Granted, the theta value in `driveTarget` is never used in any calculations so it doesn't affect anything right now, so I'll defer to your judgement on how to fix this.

abhaybd (approved):

PR #56, opened 2020-12-03
Implement navigation state and target orientations

nhatch:
Once the rover gets close enough to the goal, we put it into a state
that indicates it should use the final target orientation, rather than
trying to aim the rover towards the goal.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -11,16 +11,14 @@ constexpr float PI = M_PI;
 constexpr double KP_ANGLE = 2;
 constexpr double DRIVE_SPEED = 8;
 
+
 Autonomous::Autonomous(const URCLeg &_target, double controlHz)
 	: target(_target),
 		poseEstimator({0.8, 0.8, 0.6}, {2, 2, PI / 24}, 1.0 / controlHz),
-		state(0),
-		targetHeading(-1),
-		forwardCount(-1),
-		rightTurn(false),
 		calibrated(false),
 		calibrationPoses({}),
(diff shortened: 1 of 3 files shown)


src/Autonomous.h
```diff
@@ -21,25 +26,25 @@ class Autonomous
...
 	RollingAvgFilter<5,3> landmarkFilter;
+	NavState state;
 
 	// determine direction for robot at any given iteration
 	double pathDirection(const points_t &lidar, const pose_t &gpsPose);
 	double angleToTarget(const pose_t &gpsPose) const;
 	bool arrived(const pose_t &pose) const;
 
+	double getLinearVel(const pose_t &target, const pose_t &pose, double &thetaErr);
+	double getThetaVel(const pose_t &target, const pose_t &pose, double &thetaErr);
```
abhaybd, line 46: Both of these should be `const`, right?

arnavthareja (approved):
Looks good to me

src/Autonomous.cpp
```diff
@@ -79,23 +81,38 @@ double transformAngle(double currAngle, double targetAngle)
 	return currAngle + ((absDist > (range / 2)) ? -sign * (range - absDist) : dist);
 }
 
-double getThetaVel(const point_t &target, const pose_t &pose, double &thetaErr)
+double Autonomous::getLinearVel(const pose_t &target, const pose_t &pose, double &thetaErr) {
```
abhaybd, line 84: I think `thetaErr` could just be a `double` instead of a `double &`

PR #58, opened 2020-12-05
EKF Pose Estimator

abhaybd:
Adds an EKF implementation. This had some overlap with the existing KF, so I created a KalmanFilterBase class, which implements the common functionality, which KF and EKF both extend.

Added a pose estimator using the new EKF. The old pose estimator using a KF was renamed to LinearPoseEstimator. I've kept it instead of deleting it because it's never been tested on real robots.

I moved the static utility methods from KalmanFilter to StateSpaceUtil.h, since they're not specific to kalman filters.

Added a wheel base width constant to Constants.h

I also changed the calibration procedure. Now, the estimator is initialized with a single gps reading, which reduces the start-up time. The filter estimates and accounts for the noise in this measurement, so it can tolerate the inaccuracy from a single data point.

It's difficult to gauge how well it'll do on a real robot but it works pretty well in simulations. It's significantly more accurate than either gps or odometry alone.


abhaybd:
I realized I wasn't very specific in my original PR message, so I've edited it to more specifically list what I've done. Let me know if I should add more info.

src/filters/ExtendedKalmanFilter.h
```diff
@@ -0,0 +1,284 @@
...
+	std::function<Eigen::Matrix<double, size, size>(const state_t &, const param_t &)> func;
+};
+
+/**
+ * Implements a discrete-time EKF. This implements the more general system form described here:
+ * https://en.wikipedia.org/wiki/Extended_Kalman_filter#Non-additive_noise_formulation_and_equations
+ *
+ * @tparam numStates The number of states in the system.
+ * @tparam numInputs The number of inputs in the system.
+ * @tparam numOutputs The number of outputs in the system.
```
nhatch, line 81: It might be more accurate to call this e.g. the dimension of the state space of the system. (Technically there are an infinite number of states.)

src/filters/ExtendedKalmanFilter.h
```diff
@@ -0,0 +1,284 @@
...
+		output_t y = measurement -
+					 outputFunc(this->xHat, outputnoise_t::Zero()); // measurement residual
+
+		this->xHat = this->xHat + K * y;
+		this->P = (Eigen::Matrix<double, numStates, numStates>::Identity() - K * H) * this->P;
+	}
+
+	/**
+	 * Set this to provide an analytic solution to df/dx.
+	 * If this is null, it will be numerically approximated.
```
nhatch, line 157: This is a neat way of doing it.

Some people might think it's a little odd that we provide the state function in the constructor, but specify its derivative by setting a public member variable, but I think this is good enough.

src/filters/KalmanFilter.h
```diff
@@ -2,9 +2,12 @@
 
 #include <Eigen/Core>
 #include <Eigen/LU>
-#include <unsupported/Eigen/MatrixFunctions>
 
-template <int numStates, int numInputs> class KalmanFilter
+#include "KalmanFilterBase.h"
```
nhatch, line 6: I'm having some trouble understanding what KalmanFilter.h is for. It looks like it's used by PoseEstimatorLinear, the deprecated old version of the pose estimator. Is it used by anything else?

Also, I would think that a Kalman Filter is just a special case of an extended Kalman Filter, where the derivatives are all constant. Is the idea here just to keep the old code in case there's a bug in the new, more general implementation?

src/filters/PoseEstimator.h
```diff
@@ -46,24 +53,35 @@ class PoseEstimator
...
-	}
+	void reset();
 
 	/**
-	 * Sets the state estimate to the supplied vector and resets the estimate covariance
-	 * matrix.
+	 * Sets the state estimate to the supplied vector and resets the estimate covariance matrix
+	 * to a diagonal matrix with large values to reflect complete uncertainty in the current
+	 * pose. If you have any information about the current pose, use reset(const pose_t &,
+	 * const pose_t &)
```
nhatch, line 70: These comments are very helpful, thanks.

src/filters/ExtendedKalmanFilter.h
nhatch: To make sure I'm understanding correctly, in this PR the state func jacobians use the default behavior of numerical approximation?

src/filters/PoseEstimator.cpp
```diff
@@ -1,49 +1,124 @@
...
+	}
+	else
+	{
+		double sinDTheta = sin(dTheta);
+		double cosDTheta = cos(dTheta);
+		B << sinDTheta, cosDTheta - 1, 0, 1 - cosDTheta, sinDTheta, 0, 0, 0, dTheta;
+		B /= dTheta;
+	}
+
+	return x + A * B * localUpdate;
```
nhatch, line 47: I haven't gone through this math in detail, so I guess I'll just run the code and see if it behaves reasonably. I'm not too familiar with analytic integration of differential drive kinematics. Sounds interesting though. . . .

Math like this might be a good case for adding some automated tests. E.g. if we start at pose (0,0,0) and apply velocity 1m/s to both wheels for half a second, we should end at pose (0.5,0,0). And some more complicated tests that involve turning, going backward, etc.

src/filters/ExtendedKalmanFilter.h
abhaybd: Yeah, unless the user specifies analytic solutions to the jacobians of the state and output functions, it will default to numerically differentiating the function itself. The user can specify any or all of the 4 jacobian functions.

src/filters/KalmanFilter.h
abhaybd: Yeah, the KalmanFilter class is only used by the PoseEstimatorLinear.

During development/testing I was using the KF pose estimator as a baseline to compare the EKF against, so I needed the PoseEstimatorLinear to still exist. My reasoning for keeping it going forward is that it's working estimation code that's simpler than the EKF, so if the EKF turns out to be terrible on real robots, we can test it against the KF estimator.

Now that you mention it though, since the KF is just a special case of the EKF there might not be a point in keeping it, assuming the EKF implementation is lacking significant bugs. The KF is probably computationally cheaper, but I'm not sure if that's reason enough to keep it.

nhatch:
Two overall high-level points of feedback:

1. To get a high-quality review of your code, it is often useful to split large changes into multiple smaller PRs. For example, this PR could have been split into a refactor (factoring out StateSpaceUtils and KalmanFilterBase), followed by a separate PR adding the EKF implementation. Even if this wasn't the order in which the code was written, it is easier to review that way, which will make it more likely that we'll catch any errors.

2. One thing that would be nice to have, especially for code that implements sophisticated calculus or linear algebra, is some automated tests. We don't necessarily need to add those in this PR since it's already very large, but the sooner the better---it's often easier to catch implementation errors that way, rather than trying to read through the code. (In the comments above, I mentioned one particular piece of code that would be nice to test. It also might be nice to test the EKF predict/correct functions somehow.) If you add these tests using Catch2 (see around `add_executable(tests` in `CMakeLists.txt`) then they will be run automatically everytime someone pushes code to GitHub.

abhaybd:
@nhatch Ok, that makes sense! I'll do that for my future prs.

Regarding the tests, should I start working on that once this gets finalized/merged? It'll take some time to figure out how catch2 works, but I should be able to do it.

nhatch:
Yeah, doing tests after this PR sounds good.

I just ran this branch on my laptop and it seems to be working well. So this seems OK to merge. (Thanks for renaming those parameters.)

PR #59, opened 2020-12-05
Implement A* planning for obstacle avoidance

nhatch:
This also implements showing a visualization window so we can visualize
    what the rover is planning to do.
    
    The current code replans once every two seconds and in the mean time
    does a hacky kind of pure pursuit.
    
    Due to the infrequent replanning, sometimes the rover crashes before
    replanning is triggered. In a future PR, we should fix this by verifying
    the plan at each timestep using the most recent lidar scans.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -6,19 +6,26 @@
 
 #include "Globals.h"
 #include "simulator/world_interface.h"
+#include "simulator/constants.h"
 
 constexpr float PI = M_PI;
 constexpr double KP_ANGLE = 2;
 constexpr double DRIVE_SPEED = 8;
 
+const transform_t VIZ_BASE_TF = toTransform({NavSim::DEFAULT_WINDOW_CENTER_X,NavSim::DEFAULT_WINDOW_CENTER_Y,M_PI/2});
 
 Autonomous::Autonomous(const URCLeg &_target, double controlHz)
-	: target(_target),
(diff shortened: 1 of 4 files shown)


src/Autonomous.cpp
```diff
@@ -183,6 +198,52 @@ void Autonomous::autonomyIter()
...
+			point_t point_t_goal;
+			point_t_goal.topRows(2) = driveTarget.topRows(2);
+			point_t_goal(2) = 1.0;
+			point_t_goal = toTransform(pose) * point_t_goal;
+			double goal_radius = 2.0;
+			plan = getPlan(lidar_scan, point_t_goal, goal_radius);
+		}
+
+		int c;
+		while ((c = viz_window.pollWindowEvent()) != -1) {}
```
abhaybd, line 215: Since `c` isn't used anywhere, I think this can be simplified to `while (viz_window.pollWindowEvent() != -1)`

src/Autonomous.cpp
```diff
@@ -183,6 +198,52 @@ void Autonomous::autonomyIter()
...
+		transform_t curr_odom = readOdom();
```
abhaybd, line 216: I don't think `curr_odom` is used anywhere, right? I think it could be deleted.

c0deface (approved):
These look good!

arnavthareja (approved):
Looks good to me!

PR #61, opened 2020-12-14
Add AR subdirectory back to CMakeLists.txt

evan3334:
When we originally set up the AR detection code, we had a few demo/tester programs that needed to be built within the context of the rest of the project, so we added this directory as a subdirectory in CMake. At some point, the CMakeLists file changed to explicitly define all executables and source files in the top-level CMakeLists.txt, which is fine, but somewhere the `add_subdirectory(ar)` line got removed and we can't build our test program now, since the CMakeLists.txt file that exists in the `ar` subdirectory isn't a full CMake file; it's intended to be included in a full CMake file.

This pull request adds the `add_subdirectory(ar)` line back to the main CMakeLists.txt file. In order to avoid cluttering the build directory and building tester programs that might not be necessary, the subdirectory CMakeLists file sets the property `EXCLUDE_FROM_ALL` on all tester programs so that they aren't built or considered when building the project; they have to be specified by name to be built. This allows us to still build our tester programs if desired while keeping them out of a "production" build. Hopefully this is an acceptable compromise.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -115,6 +115,8 @@ add_executable(visualizer_test
   Pathfinding/ObstacleMap.cpp
   )
 
+add_subdirectory(ar)
+
 list(APPEND sfml_libs
   sfml-graphics sfml-window sfml-system pthread Eigen3::Eigen)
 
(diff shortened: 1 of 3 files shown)


nhatch:
LGTM. Not sure whether any of the other AR people have opinions about this. I'll just merge this, and if objections come up (which seems unlikely) we can fix it in a follow-up.

PR #62, opened 2020-12-16
Differential Drive Kinematics

abhaybd:
In preparation of writing tests for the kinematics code, I refactored the pose estimator to put the differential drive kinematics math into a separate class (`DiffDriveKinematics`), which will make it cleaner and easier to test.

I modified the `PoseEstimator` class to use the new `DiffDriveKinematics` class as well.

### DIFF
--- a/src/filters/PoseEstimator.cpp
+++ b/src/filters/PoseEstimator.cpp
@@ -5,46 +5,12 @@
 
 namespace
 {
-statevec_t stateFunc(double wheelBase, double dt, const statevec_t &x, const Eigen::Vector2d &u,
-					 const Eigen::Vector2d &w)
+statevec_t stateFunc(const DiffDriveKinematics &kinematics, double dt, const statevec_t &x,
+					 const Eigen::Vector2d &u, const Eigen::Vector2d &w)
 {
-	// Instead of using euler integration, we can represent the pose update as a twist
-	// and then apply that twist to the current pose. Derived from differential drive kinematics.
-	// This will be more accurate than euler integration at slower update rates.
 	Eigen::Vector2d input = u + w; // noise is applied to the input vector
-	double theta = x(2);
-
-	Eigen::Matrix<double, 3, 2> trf;
-	trf << 0.5, 0.5, 0, 0, -1 / wheelBase, 1 / wheelBase;
-
-	// This represents [deltaX, 0, deltaTheta]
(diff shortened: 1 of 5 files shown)


src/kinematics/DiffDriveKinematics.h
```diff
@@ -0,0 +1,95 @@
...
+	/**
+	 * Calculate the pose update in the local reference frame assuming the left and right
+	 * wheels have constant velocity.
+	 *
+	 * @param wheelVel The velocities of the wheels.
+	 * @param dt The elapsed time in seconds for which the update is calculated. The robot
+	 * has been moving at the specified velocity for this much time.
+	 * @return The pose update in the robot's local reference frame.
+	 */
+	pose_t getLocalPoseUpdate(const wheelvel_t &wheelVel, double dt) const;
```
nhatch, line 68: Do we need both of these `getLocalPoseUpdate` functions? Seems like they're almost identical, so it might be better to keep the interface simple.

src/filters/PoseEstimator.cpp
```diff
@@ -5,46 +5,12 @@
...
 namespace
 {
-statevec_t stateFunc(double wheelBase, double dt, const statevec_t &x, const Eigen::Vector2d &u,
-					 const Eigen::Vector2d &w)
+statevec_t stateFunc(const DiffDriveKinematics &kinematics, double dt, const statevec_t &x,
+					 const Eigen::Vector2d &u, const Eigen::Vector2d &w)
 {
-	// Instead of using euler integration, we can represent the pose update as a twist
-	// and then apply that twist to the current pose. Derived from differential drive kinematics.
-	// This will be more accurate than euler integration at slower update rates.
```
nhatch, line 13: I kind of liked this comment. Might be worth keeping?

nhatch (commented):
This is great! Very easy to review, and pulling out this code seems like it will make the overall pipeline easier to understand.

I have a couple of small inline comments, but nothing major.

src/kinematics/DiffDriveKinematics.h
abhaybd: I had just defined one as an overload for the other so it would accept a wider variety of parameters, but I see your point about simplicity, so I removed one of them.

abhaybd:
@nhatch I've addressed the two comments you left. Is there anything else I should change, or am I good to merge?

nhatch:
LGTM! Merged.

PR #63, opened 2020-12-29
Fix bug in RollingAvgFilter

abhaybd:
While writing unit tests, I found this bug in the rolling average filter. The size and index fields are not initialized on startup, so they have arbitrary values at the beginning, causing the filter to break.

This is pretty easily fixed by initializing them to 0.


nhatch:
LGTM

PR #64, opened 2020-12-30
hopefully the last fix to RollingAvgFilter

abhaybd:
It seems that my environment *usually* zero-initialized vectors, but this isn't guaranteed or enforced, so it would randomly and infrequently fail unit tests.

Explicitly initializing the vector used to sum the data points to zero fixed this. I've tested it many times since, so hopefully this should be ironed out.

### DIFF
--- a/src/filters/RollingAvgFilter.h
+++ b/src/filters/RollingAvgFilter.h
@@ -17,7 +17,7 @@ public:
 	 */
 	Eigen::Matrix<double, numDims, 1> get() const
 	{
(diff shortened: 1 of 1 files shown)


nhatch:
Yay! Glad to hear unit tests have already come in useful!

And yeah Eigen is a very shoot-yourself-in-the-foot kind of library. It will do no superfluous computations unless you explicitly tell it to (e.g. zeroing vectors, checking dimensions when using Eigen::Map, etc.).

PR #67, opened 2021-01-03
Explicitly zero-init vector when calibrating

abhaybd:
More of the same mistake where I didn't explicitly zero-init a vector. It seems to work most of the time since it was usually zero-initialized but that's not guaranteed.

I've grepped through the rest of the project and I'm pretty sure there are no other occurrences of this mistake.


nhatch:
LGTM

Now might also be a good time to mention that @arnavthareja @c0deface should also feel free to review and approve PRs if they feel so inclined!

PR #69, opened 2021-01-30
Each timestep, check whether the plan collides, and replan if necessary

nhatch:
Unfortunately, in testing, this still doesn't avoid collisions all of
the time (or even most of the time), but at least it tries a bit harder.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -10,7 +10,7 @@
 
 constexpr float PI = M_PI;
 constexpr double KP_ANGLE = 2;
-constexpr double DRIVE_SPEED = 8;
+constexpr double DRIVE_SPEED = 3;
 const Eigen::Vector3d gpsStdDev = {2, 2, PI / 24};
 constexpr int numSamples = 1;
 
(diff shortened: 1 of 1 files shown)


src/Autonomous.cpp
```diff
@@ -278,6 +278,13 @@ void Autonomous::autonomyIter()
 				plan_pose(2) += action(0);
 				plan_pose(0) += action(1) * cos(plan_pose(2));
 				plan_pose(1) += action(1) * sin(plan_pose(2));
+				transform_t tf_lidar_base = toTransform(pose);
+				transform_t tf_plan_pose = toTransform(plan_pose) * tf_lidar_base.inverse();
```
abhaybd, line 282: Would it be more efficient to calculate the inverse of the transformation matrix outside of the for loop? Since tf_lidar_base has the same value every iteration it would save time to just compute it once before the loop.

nhatch: Fixed!

abhaybd (approved):
Lgtm!

PR #70, opened 2021-02-06
Improve collision avoidance for planning

nhatch:
(1) We only switch to a new plan if its cost is significantly lower than
the current plan's cost.

(2) We stop the robot if the plan collides with an object near the
robot.

(3) As a refactor, the "should_replan" boolean is removed in favor of
just setting the plan cost to INFINITE_COST.

Screencap demo of current capabilities: https://huskyrobotics20-21.slack.com/files/U01B55M6DNV/F01MM8KTEM7/no_collisions.mp4


arnavthareja (approved):
Looks good to me!

PR #71, opened 2021-02-11
ROS2 Plan Visualization

arnavthareja:
This pull request fixes the issue of two SFML windows not being supported on some OSs, causing the simulator and plan visualization not being able to run. The simulator now publishes information for visualization to a ROS topic, and the new PlanViz executable reads information from the topic and displays it. The simulator and plan visualization are now different executables (RoverSim and PlanViz), so a simulator window can be run without visualization and vice versa.

I have also updated the Docker CI to include ROS2. It is now based on Ubuntu 20.04, as support for the ROS2 version that targets Ubuntu 18.04 is ending in May. I have also added a reference version of the Dockerfile used to create the CI image so it is easier to access if it needs to be updated in the future. It is based on the Dockerfile from the previous version of the CI, which can be found at [evan3334/PY2020-test-deps](https://www.github.com/evan3334/PY2020-test-deps). Right now, the reference version is located at `tests-docker-action/rover-test-deps-ros-Dockerfile`; please let me know if it should be stored somewhere else.

A few important things to note:
- The simulator and plan visualization are now different executables (RoverSim and PlanViz)
- As such, the simulator and plan visualization must be run from different terminals if you want to run both simultaneously
- Trying to visualize plans of multiple simulator windows at once will result in unexpected visualization behavior
- However, using multiple plan visualization windows with one simulator window will work as expected
- You can use the same plan visualization window for multiple non-overlapping simulator runs
- The current implementation requires ROS2

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -3,8 +3,7 @@
 #include <Eigen/LU>
 #include <cmath>
 #include <iostream>
-#include <rclcpp/rclcpp.hpp>
-#include <geometry_msgs/msg/point.hpp>
+#include <csignal>
 
 #include "Globals.h"
 #include "simulator/world_interface.h"
@@ -15,13 +14,14 @@ constexpr double KP_ANGLE = 2;
 constexpr double DRIVE_SPEED = 8;
 const Eigen::Vector3d gpsStdDev = {2, 2, PI / 24};
 constexpr int numSamples = 1;
-const rclcpp::Node::SharedPtr node = std::make_shared<rclcpp::Node>("autonomous.cpp");
-////const rclcpp::Publisher<geometry_msgs::msg::Point>::SharedPtr plan_pub = node->create_publisher<geometry_msgs::msg::Point>("plan_viz", 10);
-const rclcpp::Publisher<geometry_msgs::msg::Point>::SharedPtr curr_pose_pub = node->create_publisher<geometry_msgs::msg::Point>("current_pose", 10);
-const rclcpp::Publisher<geometry_msgs::msg::Point>::SharedPtr next_pose_pub = node->create_publisher<geometry_msgs::msg::Point>("next_pose", 10);
-////const rclcpp::Publisher<geometry_msgs::msg::Point>::SharedPtr lidar_pub = node->create_publisher<geometry_msgs::msg::Point>("lidar_scan", 10);
 
 const transform_t VIZ_BASE_TF = toTransform({NavSim::DEFAULT_WINDOW_CENTER_X,NavSim::DEFAULT_WINDOW_CENTER_Y,M_PI/2});
+
+void stop(int signum)
+{
+	rclcpp::shutdown();
+	raise(SIGTERM);
+}
 
 Autonomous::Autonomous(const URCLeg &_target, double controlHz)
 	: //viz_window("Planning visualization"),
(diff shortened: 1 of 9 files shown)


src/Autonomous.cpp
```diff
@@ -269,9 +279,16 @@ void Autonomous::autonomyIter()
...
 		}
 
-		while (viz_window.pollWindowEvent() != -1) {}
-		viz_window.drawPoints(transformReadings(lidar_scan, VIZ_BASE_TF), sf::Color::Red, 3);
-		drawPose(pose, pose, sf::Color::Black);
+		// Send lidar points to visualization
+		const points_t lidar_scan_transformed = transformReadings(lidar_scan, VIZ_BASE_TF);
+		for (point_t point : lidar_scan_transformed)
+		{
+			publish(point, lidar_pub);
```
nhatch, line 286: I've actually never used ROS2 before, but in ROS 1 there are different message types that you can use to publish e.g. a list of points, rather than a single point at a time. We might eventually want to switch to use those---maybe it would be more computationally efficient, or play better with other ROS tools like Rviz.

tests-docker-action/rover-test-deps-ros-Dockerfile
```diff
@@ -0,0 +1,116 @@
...
+RUN apt-get install -y locales
+RUN locale-gen en_US en_US.UTF-8 && update-locale LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8
+ENV LANG=en_US.UTF-8
+
+# Add ROS GPG keys
+RUN curl -s https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | apt-key add -
+RUN sh -c 'echo "deb [arch=$(dpkg --print-architecture)] http://packages.ros.org/ros2/ubuntu $(lsb_release -cs) main" > /etc/apt/sources.list.d/ros2-latest.list'
+
+# Install ROS
+RUN apt-get update && apt-get install -y ros-foxy-ros-base
```
nhatch, line 116: Good idea to add this to our main repo. This seems like a fine place for it.

src/PlanViz.cpp
```diff
@@ -0,0 +1,75 @@
...
+			viz_window.display();
+		}
+		else
+		{
+			viz_window.drawRobot(toTransform({message->x, message->y, message->z}), sf::Color::Red);
+		}
+	}
+
+	points_t lidar_scan;
+	MyWindow viz_window;
```
nhatch, line 57: Eventually it might be nice just to use Rviz instead of needing to write custom code for this. But that would require starting to think about how to publish TF (transform) information, which might be tough, so this is great for now.

nhatch (commented):
I'm trying to build and run this on my machine and I'm getting some linker errors. I can try again to debug this this weekend, but on the off chance, did you see / solve this error at any point while working on this PR?

```undefined reference to `rosidl_message_type_support_t const* rosidl_typesupport_cpp::get_message_type_support_handle<geometry_msgs::msg::Point_<std::allocator<void> > >()'
collect2: error: ld returned 1 exit status```

nhatch:
I forgot to say, THIS IS AWESOME!!! I'm really looking forward to merging this once I get it working locally.

nhatch:
Ah I figured it out! It was a problem caused by my already having another version of ROS installed on my system. Apparently if your source devel/setup.sh for two versions of ROS simultaneously, you can have problems....

I was able to test it. Looks good, I'll merge

PR #72, opened 2021-02-13
Added working global lidar map - will improve

abhaybd:
This adds a global lidar map, which I've tested and it works okayish with the robot sim. This PR doesn't hook it up to the autonomous code since I still need to add in a few performance enhancements, as well some accuracy improvements. I wanted to merge it in before continuing because otherwise the PR might get too big.

It adds the code for global mapping, as well as some unit tests that verify behavior.

After merging, I'll continue work on improving it.


arnavthareja (approved):
Looks good to me!

PR #74, opened 2021-02-19
Faster compile time (50% speedup)

nhatch:
Compiling things on the Jetson was getting to be a bit of a bottleneck. Also, I've been wanting to clean up our CMakeLists.txt for a while. This PR attempts to speed up things by:

* removing old files (ObstacleMap, Pather2) which are no longer used
* making liberal use of `add_library` so we compile most files into libraries which can be linked by multiple targets
* add more networking stubs so that we can compile `Rover.cpp` into our `tests` executable

Other changes:

* bump CMAKE_CXX_STANDARD to 14 for all targets (not a speed improvement, just a simplification)
* remove some Eigen3::Eigen references from link_libraries --- doesn't make sense since Eigen is header-only

On my laptop, on master, removing the build directory and building everything from scratch currently takes 3 min 58 seconds. On this branch, it takes 1 min 39 seconds. There may be further improvements possible; for instance, I'm not sure why it takes so long to compile Rover.cpp and Autonomous.cpp.


abhaybd:
One thing that sped up compilation time for me was changing the `CMakeLists.txt` to use the [lld linker](https://lld.llvm.org/). It required me to install an extra package, but it drastically sped up the linking times.

I didn't ever push this since I'm not really sure how well it transfers between platforms, but maybe you could take a look at it and add it to this PR if it's good?

nhatch:
Interesting idea @abhaybd . I hadn't noticed the linking times being the bottleneck, but LLD sounds intriguing. I'll keep this in mind!

nhatch:
@abhaybd I tried installing LLD. I'm running into an issue that LLD doesn't seem to use the LD_LIBRARY_PATH environment variable to check where to look for libraries (so it can't find the SFML libraries). How did you specify the directories where LLD should look for libraries?

abhaybd:
@nhatch how did you go about setting it up? For me, it was as easy as `sudo apt install llvm`, and then I just added this line:

```
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=lld")
```

To the CMakeLists.txt. I'm not sure if it matters, but I put it at the top.

After that, it didn't really require any further configuration, since it's supposed to be a drop-in replacement for LD.

nhatch:
Thanks, that worked! I also had to add `-L/usr/local/lib` to `CMAKE_EXE_LINKER_FLAGS` because apparently `lld` doesn't check the environment variable `$LD_LIBRARY_PATH`.

In a brief test it seems like, on my laptop, for some reason, `lld` isn't perceptibly faster than `ld`. Maybe given the potential compatibility issues, I won't suggest we add this to a PR right away. Still, it's good to have this conversation to point to for other people who are experiencing slow linker times.

PR #75, opened 2021-02-21
Visualize landmarks in PlanViz

nhatch:
This commit also demonstrates how we can use the PoseArray message type
to send a long list of points all at once. It also starts to think about
how we'll eventually use transform frames.

When we move to 3D, we're going to have to come up with a better way to
associate landmark location information with the ID of the landmark.
Right now we use that the z-value is zero, which might not work well
once we have landmarks where the z-value is important.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -19,6 +19,8 @@ constexpr double PLAN_COLLISION_STOP_DIST = 2.0;
 constexpr double INFINITE_COST = 1e10;
 constexpr int REPLAN_PERIOD = 20;
 
+constexpr const char * ROVER_FRAME = "rover";
+
 const transform_t VIZ_BASE_TF = toTransform({NavSim::DEFAULT_WINDOW_CENTER_X,NavSim::DEFAULT_WINDOW_CENTER_Y,M_PI/2});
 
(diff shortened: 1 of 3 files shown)


arnavthareja (commented):
I'm seeing some weird behavior when the rover gets close to the landmark: sometimes, it will try to go back to its previous plan pose instead of setting the landmark as the drive target, and it will end up driving in circles a bit before it reaches the target. I think this is caused by the change in `FOLLOW_DIST`, as I don't see this behavior when I change it back to 5. I'm not sure if this is something that needs to be fixed in this PR or if we can address it later, as it is infrequent and the rover still reaches the goal in the end. Other than that, everything looks good to me.

nhatch:
Yeah, maybe I shouldn't have changed FOLLOW_DIST in this PR, as it's a bit orthogonal to the ROS stuff.

Decreasing FOLLOW_DIST to 2.0 solves a different issue, where the rover would sometimes crash into obstacles by cutting corners too aggressively. I think . . . let's just leave it at 2.0 for this PR.

nhatch:
Made a clickup task for fixing the driving in circles bug: https://app.clickup.com/t/nq0rky

PR #78, opened 2021-02-25
Change base station communication protocol to specify packet lengths

nhatch:
The new protocol expects four bytes, little endian, specifying the
length of the (JSON) packet to follow. Then the JSON packet must be
exactly that length, and there must be no intervening bytes before or
afterward, until the next four-byte length code.

This ensures that if the base station sends two packets (nearly)
simultaneously, we can separate them and process them individually.

Requires a corresponding PR on MissionControl: https://github.com/huskyroboticsteam/MissionControl/pull/16

Addresses https://app.clickup.com/t/kt181t . I tested this out with the base station. Without this fix, the program crashes. With it, it seems to run fine, no matter how much I mash the keyboard.

Also robustifies the connection between the rover and the base station: if the base station disconnects, the rover will periodically try to reconnect.


nhatch:
Mission control PR has been merged, so we should probs just merge this now.

PR #79, opened 2021-02-27
Added QuadTree data structure

abhaybd:
It's not hooked up to anything yet, which I'll do in a future PR.

It implements the quadtree data structure for fast lookup times of points in a 2d plane, which the lidar map will use for fast nearest neighbor calculations.

I also added unit tests for the quad tree implementation.


tests/worldmap/QuadTreeTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+	for (int i = 0; i < 50; i++)
+	{
+		// make the size smaller, so we can guarantee that some points are not in the tree
+		tree.add(randPoint(5));
+	}
+
+	auto size = tree.getSize();
+	REQUIRE_FALSE(tree.remove({53, 1, 1}));
+	REQUIRE_FALSE(tree.remove({0, 0, 0}));
```
nhatch, line 78: I'm not suuuper concerned about this, but is there a chance that this point might actually be in the tree?

tests/worldmap/QuadTreeTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+		REQUIRE(tree.add(point));
+		REQUIRE(tree.getSize() == i + 1);
+	}
+
+	REQUIRE_FALSE(tree.empty());
+}
+
+TEST_CASE("QuadTree - RemovePoints", "[QuadTree]")
+{
+	srand(time(nullptr)); // NOLINT(cert-msc51-cpp)
```
nhatch, line 40: Does this NOLINT comment have some effect on our testing suite?

tests/worldmap/QuadTreeTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+namespace
+{
+double randNum(double low, double high)
+{
+	return low + (rand() / (RAND_MAX / (high - low))); // NOLINT(cert-msc50-cpp)
+}
+
+point_t randPoint(double size)
+{
+	return {randNum(-size / 2, size / 2), randNum(-size / 2, size / 2), 1};
```
nhatch, line 16: It might be worth thinking about what will happen when we move to 3D. We can always probably project obstacle points down into a 2D map, but maybe at some point we will want to build a 3D map. No changes needed in this PR, but just food for thought.

tests/worldmap/QuadTreeTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+		auto itr = std::find(closestToRight.begin(), closestToRight.end(), point);
+		REQUIRE(itr != closestToRight.end());
+		REQUIRE(point == *itr);
+	}
+}
+
+TEST_CASE("QuadTree - GetClosestPoint", "[QuadTree]")
+{
+	srand(time(nullptr)); // NOLINT(cert-msc51-cpp)
+	QuadTree tree(100);
```
nhatch, line 177: Oddly, I cannot find a QuadTree constructor that takes just one argument. Does this compile?

nhatch (commented):
I only looked at the tests, but overall this seems good to me.

Out of curiosity, did you try looking for existing QuadTree implementations?

tests/worldmap/QuadTreeTest.cpp
abhaybd: The second one has a default value of 4, so invoking it with one parameter compiles fine I think.

https://github.com/huskyroboticsteam/PY2020/blob/ac1b87d04da18e0e709774037b455f913651b02b/src/worldmap/QuadTree.h#L25

tests/worldmap/QuadTreeTest.cpp
abhaybd: The main reason for putting it was my IDE was just yelling at me since clang-tidy doesn't like it when you use srand and rand instead of the newer random functions. I don't think it has any other side effects besides silencing warnings, right?

tests/worldmap/QuadTreeTest.cpp
abhaybd: https://github.com/huskyroboticsteam/PY2020/blob/d0c4c596cfb738e6b03d628f134dd6e8b1c587af/tests/worldmap/QuadTreeTest.cpp#L70-L74

All the points added fall in the bounding box centered at (0,0,1) with a height and width of 5, so all of these points are guaranteed to fall outside of this range. (and {0,0,0} has an invalid z-coord, so that one doesn't work either)

tests/worldmap/QuadTreeTest.cpp
abhaybd: Hmm, in that case I'd either have to overhaul all the lidar map code to work in 3d, or there's probably a library out there well suited for this. Could we use this for now, and I'll do some research into 3d?

abhaybd:
I think I just found a bug in the getClosest() method, so this probably shouldn't be merged yet. I'll let you know when its fixed.

tests/worldmap/QuadTreeTest.cpp
nhatch: Oh right! I forgot that 0 is an invalid z coordinate.

tests/worldmap/QuadTreeTest.cpp
nhatch: Yeah, of course this is fine for now. 3D research might be a bit premature at this point, but feel free to take a look if that sounds interesting

PR #81, opened 2021-02-27
Added QuadTree data structure - for real this time

abhaybd:
Ok, the blocking bug has been reproduced and fixed, so this PR should be final.

======================

It's not hooked up to anything yet, which I'll do in a future PR.

It implements the quadtree data structure for fast lookup times of points in a 2d plane, which the lidar map will use for fast nearest neighbor calculations.

I also added unit tests for the quad tree implementation.


nhatch:
For posterity: the previous PR was https://github.com/huskyroboticsteam/PY2020/pull/79, see comments there.

LGTM!

PR #82, opened 2021-03-04
Implement the new AR tag detection system

evan3334:
This pull request implements the new AR tag detection system as documented here: https://app.clickup.com/2200143/v/dc/234jf-1778/234jf-3157

The notable changes are ability to use both ARUco tags (used in CIRC) and ALVAR tags (used in URC), and the use of a "MarkerSet" system to map the IDs of the tags to useful values. Additionally, the system is better and much faster at detecting tags since it uses the built-in OpenCV ARUco module for detection. 

This is kind of a first draft of the system; we have tested it on our webcams at home with both types of tags and it seems to work fine, but we will probably open pull requests later to add more tests or fix bugs as needed.


src/ar/ARTester.cpp
```diff
@@ -1,105 +1,326 @@
...
-const std::string THRESH_TRACKBAR_NAME = "Threshold";
-const std::string THRESH2_TRACKBAR_NAME = "Threshold 2";
-const std::string BLUR_TRACKBAR_NAME = "Blur";
 
 constexpr bool EXTRA_WINDOWS = true;
 
+// Set to whichever camera params should be used
+AR::CameraParams PARAMS = AR::getCameraParams(AR::Params::ROBOT_TOP_WEBCAM_480);
+
+// Set to whichever MarkerSet should be used (CIRC/URC)
```
nhatch, line 26: Maybe we could mention that this can be overridden by command line params

src/ar/ARTester.cpp
```diff
@@ -1,105 +1,326 @@
...
+				grid_points.push_back(cv::Point2f(x, -y) + center);
+			}
+		}
+	}
+	cv::undistortPoints(grid_points, projected_points, PARAMS.getCameraParams(),
+						PARAMS.getDistCoeff());
+	return projected_points;
+}
+
 int main(int argc, char *argv[])
```
nhatch, line 84: This is a REALLY long function. In my experience, it's easier to maintain code when it is broken into multiple smaller functions. We don't have to fix that for this PR, but it's something to think about for the future.

src/ar/Marker.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <memory>
+#include <opencv2/core.hpp>
+#include <cstdint>
+
+/**
+   A class to represent the pattern that may appear on a physical AR tag, NOT a physical
+   instance of a tag itself; those are represented by the Tag class. Markers are square
```
nhatch, line 9: I'm a little unclear on the difference here. Is it just that an image may have multiple Tags corresponding to the same Marker?

src/ar/Tag.cpp
```diff
@@ -1,175 +1,29 @@
...
-
-#include <opencv2/calib3d.hpp>
-
-#include <cassert>
-#include <iostream>
-#include <cmath>
 
 namespace AR
 {
-double findAngle(cv::Point a, cv::Point b, cv::Point c)
```
nhatch, line 12: I guess all this code is deleted because you switched to the OpenCV implementation?

nhatch (commented):
Ok, I _skimmed_ this PR and it looks fine to me. I'm hoping Winston may be able to provide more comments about the specific implementation details. But since this is a pretty big PR I think we can't necessarily expect to catch all the possible bugs here. We'll just fix stuff as it comes up later.

src/ar/Marker.h
evan3334: Yeah, I think in the competition they will have these triangular prisms mounted on the posts with each face being a tag that has the same marker pattern, so it could be possible that there are multiple tags corresponding to the same marker. It also seemed to make sense to me when I first introduced this class to keep the concept of the pattern on the tag and the physical tag separate in a way. Now that you mention it though, I agree the naming is a little confusing; maybe `MarkerPattern` for the pattern, keeping `Tag` for the physical occurrence?

src/ar/Tag.cpp
evan3334: Yeah, partly that and partly because we wanted the Tag to be a simpler data structure that just holds a position, rotation, etc. Previously we had the Tag computing its own orientation when you constructed it, which meant that you needed to give it image coordinates of the corners to construct it. That meant we couldn't really use it in the simulator or in testing because we required that you actually detect a tag in an image to make a Tag object. We decided that was more of a job for the detector and that the Tag should just be a simple data structure for returning results from the detector, so all that code was removed; it was replaced in the detector by OpenCV's implementation.

src/ar/ARTester.cpp
evan3334: It's actually given a default value by the command line parser so on second thought it doesn't make too much sense to assign it a default value there. I think that comment is left over from before we had a command line parser and we would just hardcode those values in constants instead, so I'll remove that.

src/ar/ARTester.cpp
evan3334: Yeah, agreed. I can still fix it for the PR or later, it would be good to have it be more maintainable in the future

src/ar/Marker.h
nhatch: The name `MarkerPattern` does seem a bit clearer, good idea! Also just updating this comment so it says something like

"A class to represent an AR tag pattern that may appear in an image. Each individual instance of the pattern is represented by an instance of the Tag object; there may be multiple Tags corresponding to the same MarkerPattern."

evan3334:
Added some more commits to this PR to address the issues that were brought up about naming/documentation, as well as to add some unit tests and fix some bugs that arose.

nhatch:
@WinstonBullen are you still planning to look over this PR? If not, I think this is good to merge.

WinstonBullen:
Sorry, I have been a bit busy recently. I took a few skims over the PR and I think it looks good. We did a good job of maintaining and refactoring the code as we went along. I think it is good to merge. @nhatch

PR #84, opened 2021-04-01
Update OpenCV install instructions

evan3334:
Our install instructions are currently to do a custom source build of OpenCV 3 with certain contrib modules mixed in; however, this is pretty unhygienic and wasteful to do on every new setup, and we probably ought to be using the version that comes with the system package manager. On most current Ubuntu installations OpenCV 4 is packaged; we've tested and it seems to be compatible (at least, the parts we used in our code) out of the box with OpenCV 3, so no code changes should be required to move to OpenCV 4 installed from APT. 

All this pull request really does is update the install instructions and change the Dockerfile that defines the image the CI runs tests on to also use those install instructions instead of building from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,10 @@ access.
 
 ## Install OpenCV
 OpenCV is a computer vision library that's used for the AR tag detection code
-and as such is needed to compile the project.
-
-The setup process takes some time and will use a lot of CPU, but it's fairly
-straightforward and should only need to be done once.
-
-We need an extra module (ARUco) for OpenCV for the AR tag detection code which
-is not bundled with OpenCV, so you will have to install that module before
-building OpenCV. The below instructions will cover this.
+and as such is needed to compile the project. We will also need the OpenCV
+contrib modules for the AR tag detection code. Either OpenCV 3 or OpenCV 4
+should work, but we are developing and testing against OpenCV 4 so you should
+try to get that version for best results.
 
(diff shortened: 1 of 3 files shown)


nhatch:
README changes look good, but ideally we should update the Docker image now so we don't forget to do so later. I believe @arnavthareja  was the last person to update the Docker image. You can check this PR to see what changed: https://github.com/huskyroboticsteam/PY2020/pull/71/files

evan3334:
Yeah, sounds good. What would it entail to update the Docker image? Are you just using that file we've checked in to build the Docker image manually then sending it to Dockerhub on the Husky Robotics account, or is there some other way we're doing it?

nhatch:
I'm not sure I've ever done this myself. @arnavthareja is that (what Evan said) how you updated the Docker image last time?

arnavthareja:
Yeah, I used the file in this repo to build the image and then pushed it to our Dockerhub.

evan3334:
Added one last fix, was just a CMake thing. For some reason in `lidar_base` it wasn't finding the OpenCV header files; no idea if this is because we're building it as a separate library or something, but I had to add a `target_include_directories` command to explicitly add the OpenCV headers to the `lidar_base` include path.

evan3334:
Going to go ahead and merge this as the changes to the CI docker image are causing testing on other branches to fail without the change to the CMake config that's in this PR.

PR #86, opened 2021-04-09
Add Doxygen for generating source code documentation

evan3334:
We should probably be documenting our source code going forward, and many of our source files already have documentation but nothing really makes use of that. This pull request adds Doxygen as a documentation system and a Github CI action to generate the docs when we push to master, and publish them on Github Pages. We shouldn't have to install anything locally unless someone wants to generate the docs locally. The only thing this means for members of the software team is that going forward we should try to include documentation comments in Doxygen format in our source code, and we may want to go back and document significant portions of our code.

As for where the docs are published, and the configuration for Doxygen (or if we should even use Doxygen as opposed to something else), if anyone has any comments or suggestions for something we should do differently, please let me know - this change is pretty low priority and I'd like to get people's thoughts first.


nhatch:
Here's an example of what the generated documentation looks like: https://evan3334.github.io/PY2020-private/classDiffDriveKinematics.html#aec54118f4bab8a82f40cf4a21c417500

LGTM!

PR #88, opened 2021-04-21
Update calibration program to latest version, from OpenCV 4

evan3334:
This pull request updates the calibration sample program to the latest version that is available in OpenCV 4. Notable changes include the use of the "releasing-object" method of camera calibration, which is more accurate for less precise calibration targets (like the printed paper targets we will most likely be using). We probably ought to use this when we calibrate the cameras on the rover. I'll update the documentation on camera calibration to include info on this.

Note that this new program uses functions that have been introduced in OpenCV 4, so as such OpenCV 4 will be *required* to build it. I don't actually think it will affect the entire project, you have to explicitly specify that you want to build the calibration program so people using OpenCV 3 should still be fine for the time being - however, we probably want to have everyone update at some point.

### DIFF
--- a/src/ar/calibration.cpp
+++ b/src/ar/calibration.cpp
@@ -1,12 +1,8 @@
 ///////////////////////////////////////////////////////////////////////////////////////
 //
-//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
-//
-//  By downloading, copying, installing or using the software you agree to this license.
-//  If you do not agree to this license, do not download, install,
-//  copy or use the software.
-//
-//
+// Program to calibrate a camera. Taken from OpenCV sample code, original
+// license agreement is below:
+// 
 //                           License Agreement
 //                For Open Source Computer Vision Library
 //
@@ -54,6 +50,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <time.h>
+#include <iostream>
 
 using namespace cv;
 using namespace std;
@@ -90,10 +87,10 @@ const char* liveCaptureHelp =
         "  'g' - start capturing images\n"
         "  'u' - switch undistortion on/off\n";
 
-static void help(char** argv)
+static void help()
 {
     printf( "This is a camera calibration sample.\n"
-        "Usage: %s\n"
+        "Usage: calibration\n"
         "     -w=<board_width>         # the number of inner corners per one of board dimension\n"
         "     -h=<board_height>        # the number of inner corners per another board dimension\n"
         "     [-pt=<pattern>]          # the type of pattern: chessboard or circles' grid\n"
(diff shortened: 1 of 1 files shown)


src/ar/calibration.cpp
```diff
@@ -106,19 +103,25 @@ static void help(char** argv)
...
+        "     [-dt=<distance>]         # actual distance between top-left and top-right corners of\n"
+        "                              # the calibration grid. If this parameter is specified, a more\n"
+        "                              # accurate calibration method will be used which may be better\n"
+        "                              # with inaccurate, roughly planar target.\n"
         "     [input_data]             # input data, one of the following:\n"
         "                              #  - text file with a list of the images of the board\n"
         "                              #    the text file can be generated with imagelist_creator\n"
         "                              #  - name of video file with a video of the board\n"
         "                              # if input_data not specified, a live view from the camera is used\n"
-        "\n", argv[0] );
```
nhatch, line 121: Why this change? Seems like a fine practice to use argv for this, in case we ever change the filename or name of the executable

src/ar/calibration.cpp
```diff
@@ -194,16 +199,32 @@ static bool runCalibration( vector<vector<Point2f> > imagePoints,
...
     objectPoints.resize(imagePoints.size(),objectPoints[0]);
 
-    double rms = calibrateCamera(objectPoints, imagePoints, imageSize, cameraMatrix,
-                    distCoeffs, rvecs, tvecs, flags|CALIB_FIX_K4|CALIB_FIX_K5);
-                    ///*|CALIB_FIX_K3*/|CALIB_FIX_K4|CALIB_FIX_K5);
+    double rms;
+    int iFixedPoint = -1;
+    if (release_object)
+        iFixedPoint = boardSize.width - 1;
+    rms = calibrateCameraRO(objectPoints, imagePoints, imageSize, iFixedPoint,
```
nhatch, line 211: Ok so I guess this is the new OpenCV4 API call? (Not actually a question, just sort of trying to prove I'm paying attention to what's going on.)

Small suggestion: instead of `using namespace cv` above, just use `cv::calibrateCameraRO` here. That makes it clear that this is an OpenCV method. We don't have to do that in this PR, but just as a suggestion for the future. "using namespace" usually makes code a bit less clear to unfamiliar readers.

nhatch (approved):
For posterity, here's the documentation on camera calibration: https://app.clickup.com/2200143/v/dc/234jf-1778/234jf-3969

Looks good to me.

src/ar/calibration.cpp
evan3334: No idea, I just copied in their sample program basically. Not sure why they made that change but if it doesn't make sense we can just exclude it

src/ar/calibration.cpp
evan3334: Yeah, that makes sense. OpenCV seems pretty eager to use `using namespace` in their sample code which isn't great practice, I agree. I can just modify that now.

evan3334:
Made the recommended changes, so I think I'm just going to go ahead and merge this.

PR #90, opened 2021-04-30
Add code to Rover.cpp to initialize encoders & add PID tuning script

nhatch:
There are three CAN messages we need to send to the motor boards in order to initialize encoder sensor measurements. This PR makes it so we send those messages automatically when starting the `Rover` executable.

This PR also adds a script to help tune PID coefficients. The script sends a trajectory of target angles (currently a sine wave of amplitude 15 degrees and period 3 seconds) to the motor boards and measures the actual angles as reported by the encoders. The idea is that a human operator will run this script several times for each motor, tweaking the PID coefficients, until we get good trajectory tracking performance. I have not actually tested this script yet, because we haven't yet gotten good data from the encoders.

In the course of compiling this script, I reorganized our CMakeLists.txt.

### DIFF
--- a/src/Rover.cpp
+++ b/src/Rover.cpp
@@ -16,31 +16,52 @@
 extern "C"
 {
     #include "HindsightCAN/CANMotorUnit.h"
+    #include "HindsightCAN/CANSerialNumbers.h"
+}
+
+void initEncoders()
+{
+    CANPacket p;
+    for (uint8_t serial = DEVICE_SERIAL_MOTOR_BASE;
+        serial < DEVICE_SERIAL_MOTOR_CHASSIS_BR;
+        serial ++ ) {
+      AssembleEncoderInitializePacket(&p, DEVICE_GROUP_MOTOR_CONTROL, serial,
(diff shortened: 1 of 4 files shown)


src/Rover.cpp
```diff
@@ -16,31 +17,64 @@
...
 }
 
-void InitializeRover()
+void InitializeRover(uint8_t arm_mode)
 {
     InitializeCANSocket();
 
     // Set all wheel motors to mode PWM
     CANPacket p;
     uint8_t mode_PWM = 0x0;
```
evan3334, line 64: This is a super minor thing but is this variable `mode_PWM` being used anymore? Noticed it was previously but now a constant `MOTOR_UNIT_MODE_PWM` is being used instead

evan3334 (approved):
Looks pretty good to me. I think there was like one minor unused variable or something but it's not a huge deal.

Just to make sure I have it right, all the motor addresses are consecutive, right? I noticed we were using a `for` loop to send a package to all the motors so I assumed that was the case.

evan3334 (approved):

PR #91, opened 2021-05-01
Add test program for getting GPS data from gpsd

nhatch:
Installation instructions are found in gps/README.txt. I tested this on
my laptop and was able to get GPS data. Next steps:
- try installing this on the Jetson
- use the test program to get GPS data within our main Rover program


tests-docker-action/entrypoint.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 git submodule update --init --recursive
 source /opt/ros/foxy/setup.bash
+
+apt-get update
+apt-get install -y gpsd gpsd-clients libgps-dev
```
evan3334, line 6: If `gpsd`, `gpsd-clients`, and `libgps-dev` are dependencies now, I think we should probably put them in the README, and also include them in the Docker image rather than installing them at build time on the CI.

src/CMakeLists.txt
```diff
@@ -187,6 +187,10 @@ target_link_libraries(simulator_navigation_demo
   simulator_base simulator_nav simulator_world_interface)
 target_link_libraries(simulator_navigation_demo ${sfml_libs})
 
+add_executable(gpsd_test
+  gps/gpsd_test.cpp)
+target_link_libraries(gpsd_test gps)
+
```
evan3334, line 193: For a lot of the testing programs I've been excluding them from the main target, so when we build the Rover program it isn't building all the test programs as well. Do you think this is a good policy, and should we implement it for this one? I also forgot to bring this up on the last pull request with the PID tuning program. 

If we want to implement that here, we can just add `set_property(TARGET gpsd_test PROPERTY EXCLUDE_FROM_ALL true)`.

src/gps/README.txt
```diff
@@ -0,0 +1,57 @@
+
```
evan3334, line 1: This file seems like it has Markdown formatting, so should this be a .md file instead of .txt?

src/gps/gpsd_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * This software is distributed under a BSD-style license. See the
+ * file "COPYING" in the top-level directory of the distribution for details.
+ *
+ */
+
+/* This simple program shows the basic functionality of the C++ wrapper class */
+#include <iostream>
+
+#include "libgpsmm.h"
```
evan3334, line 19: CMake seems to handle it fine either way but just for the sake of conventions I think this should use angle brackets to make it clear this is a system header and not a local one.

src/gps/gpsd_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+ * scons has a bug; we can't get it to add -lgps to the link line,
+ * apparently because it doesn't honor parse_flags on a Program()
+ * build of a C++ file.
+ */
+static void libgps_dump_state(struct gps_data_t *collect)
+{
+    /* no need to dump the entire state, this is a sanity check */
+#ifndef USE_QT
+    /* will fail on a 32-bit macine */
+    (void)fprintf(stdout, "flags: (0x%04x) %s\n",
```
evan3334, line 34: Is there a reason these use `fprintf(stdout, ...)` instead of just `printf`? Not saying it has to change, just curious.

evan3334 (changes requested):
Looks pretty good, I just had a few minor comments. As for the CI thing, I can update the image pretty easily if you add the change to the Dockerfile (or I could do that part as well, either way works).

tests-docker-action/entrypoint.sh
nhatch: Personally, I would rather do it this way. Updating the Docker image is kind of difficult, so I think it makes sense for the normal workflow to just edit this CI script. Installing stuff doesn't take much time (since we're not building from source). What do you think?

I'll add this line to the README though, that's a good idea.

src/CMakeLists.txt
nhatch: Hmm. I think we can potentially adopt the policy that people usually want to build individual targets (e.g. `make Rover`). So it's probably okay for `make` to by default build all of the targets---since most of the time people won't be doing that. In fact, we probably want cmake to build everything by default, because that's what we use for CI:

```
cmake --build ./build -j$(nproc)
./build/tests
```

We want to make sure CI is trying to build everything, so that we catch build errors when they're introduced.

src/gps/README.txt
nhatch: Yeah, I'll change that.

tests-docker-action/entrypoint.sh
evan3334: Yeah, that sounds good. Maybe if we end up having a lot of extra dependencies installed there we can move them to the Docker image all at once? We can cross that bridge when we come to it though.

evan3334 (approved):

PR #92, opened 2021-05-01
Implement driving through gate posts

arnavthareja:
This PR implements the behavior for driving through gates. The rover sets two targets, one in front of the gate and one behind, which are visualized in green in PlanViz. Behavior is not perfect, as the rover sometimes drives through a post. This might be fixed if we consider posts to be similar to obstacles while planning, although I'm not sure if this would result in incorrect behavior in non-gate legs.


src/Autonomous.cpp
```diff
@@ -243,23 +264,99 @@ void Autonomous::autonomyIter()
...
+					// Align to target to improve planning
+					state = NavState::GATE_ALIGN;
+				}
+
+				// we have already seen the first post and have set gate targets
+				if (!std::isnan(gate_targets.first(2)))
+				{
+					// Update gate targets
+					if (!std::isinf(gate_targets.first(2)) &&
+						dist(pose, gate_targets.first, 0.0) < 0.2)
```
nhatch, line 333: Here you should probably use a nonzero theta weight (e.g. `dist(blah, blah, 1.0)`). We want to make sure that we've actually achieved the target orientation as well. Otherwise who knows what will happen while the robot turns to face the second target---it might crash into a gatepost.

src/Autonomous.cpp
```diff
@@ -243,23 +264,99 @@ void Autonomous::autonomyIter()
...
+					{
+						// Arrived at first gate target, mark as complete
+						gate_targets.first = pose_t(INFINITY, INFINITY, INFINITY);
+
+						// Update target angle of second target to account for inaccuracies
+						gate_targets.second(2) = std::atan2(gate_targets.second(1) - pose(1), gate_targets.second(0) - pose(0));
```
nhatch, line 339: Again, I think we should handle this just by constantly recomputing both gate targets (earlier in the control loop, where I left that other comment).

src/Autonomous.cpp
nhatch: Actually, here's an idea. There is some duplicated code here in the sense that we often want to check how close we are to the target, and if so, advance the navigation state and re-plan. Maybe we can pull all that into a single section of code to make sure that we're doing these things consistently.

To facilitate that, we might want to clarify how we're using various local variables such as `driveTarget`. We should probably use `driveTarget` just for local control targets, and introduce a different local variable `planTarget` for the pose we're planning to reach. (Probably also a third local variable `navigationTarget` for the URC GPS target, because that's often different from the planning target, e.g. during search patterns.)

That simplifies some of the logic here, because we just need to test whether the current pose is close to the `planTarget`. If so, we should advance to the next state, compute a new `planTarget`, and trigger a replan by setting `plan_cost` to `INFINITE_COST`. The navigation states could be for instance, FOLLOWING_GPS, SEARCH_PATTERN, POST_VISIBLE, GATE_ALIGN, GATE_TRAVERSE or something like that.

src/Autonomous.cpp
```diff
@@ -243,23 +264,99 @@ void Autonomous::autonomyIter()
...
+					{
+						// Arrived at second gate target
+						gate_targets.second = pose_t(INFINITY, INFINITY, INFINITY);
+
+						// No further action needed since goal has been reached
+						already_arrived = true;
+						return;
+					}
+					driveTarget = std::isinf(gate_targets.first(2)) ? gate_targets.second : gate_targets.first;
+					// Avoids weird planning behavior when close to the second target
```
nhatch, line 354: This is a good catch. As mentioned above, we should re-plan whenever we reach a key waypoint (such as the first gate target).

nhatch (changes requested):
Thanks for the PR, @arnavthareja ! Great to see that this is working.

I'm requesting changes because the autonomy code is getting high-key confusing. Not your fault! Just a side effect of all the features we've been adding over time. I think we need to refactor things (and this refactor might address some of the issues you mentioned with crashing into gateposts).

Let me know if my comments below make sense to you. If you don't feel comfortable doing these refactors, let me know. In that case, we can probably just merge this and I'll make a follow-up PR to clean things up.

evan3334 (changes requested):
I'm also glad to see this is working! I think Nathan pretty much covered most of the concerns with this pull request, so it looks good to me as long as you make the changes he suggested. I'm just going to mark "Request changes" for now and I can undo it or hit "Approve" later if needed.

src/Autonomous.cpp
```diff
@@ -243,23 +264,99 @@ void Autonomous::autonomyIter()
...
 			// caused by filtering
-			driveTarget.topRows(2) = landmarkFilter.get(landmarkMapSpace).topRows(2);
-			// clear the filter so the second post is removed from it
-			landmarkFilter.reset();
-		}
-		// If we are in a search pattern, set the drive target to the next search point
-		else if (state == NavState::SEARCH_PATTERN)
-		{
-			driveTarget = search_target;
+			point_t post_2 = gate_filter.get(landmarkMapSpace);
```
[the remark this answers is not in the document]
arnavthareja: I agree, `leftPostFilter` and `rightPostFilter` make more sense

nhatch:
@arnavthareja I went ahead and made some of those larger refactors that I suggested. I figured it wasn't really fair asking you to do that since the mess was there long before you arrived :sweat_smile: 

Here's a video of the latest version doing a gate traversal: https://huskyrobotics20-21.slack.com/files/U01B55M6DNV/F023KB1QARX/gate_traverse.mp4

@evan3334 or @arnavthareja I basically rewrote the entire Autonomous.cpp file. This may be too large a change to review in detail, but maybe someone should give a once-over to see if the new structure makes sense. Basically I tried to break out things into (slightly) smaller functions.

src/Rover.cpp
```diff
@@ -129,6 +129,7 @@ const double CONTROL_HZ = 10.0;
 int rover_loop(int argc, char **argv)
 {
     LOG_LEVEL = LOG_INFO;
+    Globals::AUTONOMOUS = true;
```
evan3334, line 132: Just curious, do we have a way to change this while the program is running?

evan3334 (approved):
Had a look through the new Autonomous.cpp. I'm not too familiar with the planning stuff yet, but I saw the simulation video in Slack and it seems like it works, so it seems good to me. Just had a few clarifying questions about things, but I didn't see any issues I think we need to change.

nhatch (approved):
Switching my review to "approved" since I made a bunch of code changes. Someone else should take a look at this though

EDIT: Oh, Evan already did!

src/Rover.cpp
nhatch: Not yet. We do need to implement that (very) soon

src/Autonomous.h
```diff
@@ -16,51 +16,77 @@
...
 
+enum NavState {
+	GPS,
+	POST_VISIBLE,
+	SEARCH_PATTERN,
+	SEARCH_PATTERN_SECOND_POST,
+	GATE_ALIGN,
+	GATE_TRAVERSE,
+	DONE
+};
```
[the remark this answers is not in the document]
nhatch: NavState is high-level navigation behavior. ControlState deals with short term, lower-level details of how we're trying to reach the driveTarget. Basically `NavState : plan_target :: ControlState : driveTarget`. I pinky swear I will document this stuff after vURC...

PR #94, opened 2021-05-11
Fix some bugs for PID control

nhatch:
Changes TunePID script to allow starting at a nonzero angle (which is
important to avoid possible issues with negative underflow).

Also makes it possible to set a different PPJR for each arm joint.

Also adds a simpler script (in FakeCANBoard.cpp) for testing PID
control with constant angles rather than time-varying angles.


evan3334 (approved):
Looks good to me!

PR #95, opened 2021-05-13
More changes to PID testing code

nhatch:
A bunch of code changes that I found useful while doing on-rover testing today.

- Implement sending actual e-stop packets
- Easier-to-use FakeCANBoard script
- Additional logging messages in the Rover.cpp main loop
- some other small changes

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,7 +155,7 @@ include(CTest)
 include(Catch)
 target_link_libraries(tests ${sfml_libs} ${rclcpp_LIBRARIES} ${geometry_msgs_LIBRARIES})
 target_link_libraries(tests Catch2::Catch2)
-target_link_libraries(tests ${OpenCV_LIBS})
+target_link_libraries(tests ${OpenCV_LIBS} opencv_aruco)
 catch_discover_tests(tests)
 
 add_library(lidar_base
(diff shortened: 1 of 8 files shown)


src/CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ include(CTest)
 include(Catch)
 target_link_libraries(tests ${sfml_libs} ${rclcpp_LIBRARIES} ${geometry_msgs_LIBRARIES})
 target_link_libraries(tests Catch2::Catch2)
-target_link_libraries(tests ${OpenCV_LIBS})
+target_link_libraries(tests ${OpenCV_LIBS} opencv_aruco)
```
evan3334, line 158: Interesting, I would have thought `opencv_aruco` would be included in `${OpenCV_LIBS}`. Did the linking error only happen on the Jetson, or on other machines as well?

evan3334 (approved):
Looks good to me. That linker error thing is interesting, I never ran into that problem before. As long as the change fixes it, it should be good.

src/CMakeLists.txt
nhatch: Oddly, it happens on my laptop. This may be an artifact of some weird way I installed OpenCV in eons past. I'm not sure about the Jetson though---I haven't tried building `tests` there.

PR #96, opened 2021-05-14
Added planning files, updated submodule reference

abhaybd:
Added removed files from the submodule, updated references, updated CMakeLists, etc.

`navigation.cpp`, `search.cpp`, and `search.h` aren't directly used by our code anywhere, but they're referenced in the CMakeLists and they're added to executables. Does that mean we still want them around, or should I delete them?

### DIFF
--- a/src/Autonomous.h
+++ b/src/Autonomous.h
@@ -14,7 +14,7 @@
 #include "lidar/PointCloudProcessing.h"
 #include "simulator/graphics.h"
 #include "simulator/utils.h"
-#include "simulator/plan.h"
+#include "planning/plan.h"
 
 enum NavState {
 	INIT,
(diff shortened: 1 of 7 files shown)


evan3334:
I would say just leave them for now; if they're added to targets in CMakeLists there's a chance they're still used for **something**, although I'm not sure what. @nhatch do you know what these files are for?

evan3334 (approved):

nhatch:
Those three files are just used for `simulator_navigation_demo`, which was something I originally wrote totally separately from Husky Robotics. I just added it to the CMakeLists because I wanted to make it easy for other team members to build it and try things out. I think it's safe to remove that target from the CMakeLists (and remove the corresponding files) now that we've implemented our own search behavior in this repo.

PR #97, opened 2021-05-15
Remove unused planning simulator files

evan3334:
Removes `search.cpp`, `search.h`, and `navigation.cpp` as per https://github.com/huskyroboticsteam/PY2020/issues/96#issuecomment-841572082.


nhatch:
There's something weird about this PR. It seems to include Abhay's commit and I don't actually see any files here being removed. Did you pull master before writing this?

evan3334:
Hmm, I thought I did, but I think I might have made a mistake. I'll fix it in a few minutes.

PR #98, opened 2021-05-15
Change style guide/clang-format config

evan3334:
I notice we often use a different brace convention by default than what is in the style guide/auto-formatter configuration, so I wanted to propose this change to the style that changes the brace configuration from
```c++
void function()
{
    // blah blah
}
```
to
```c++
void function() {
    // blah blah
}
```

This is only a draft at this point, so if anyone has any objections to this change or any other parts of the style they want to change, please mention them.


StyleGuide.md
```diff
@@ -35,8 +32,8 @@ Bad:
 ```
 - ALWAYS guard a header file with `#pragma once`
 - Avoid #define, prefer constexpr
-- Avoid raw pointers, prefer smart pointers (i.e shared_ptr) or references  
+- Avoid raw pointers, prefer smart pointers (i.e shared_ptr) or references
 
 ### Indentation
-Use tabs to indent, and use additional spaces to align multi-line code (i.e. when lines are too long and
```
nhatch, line 41: Ironic

abhaybd:
Could we also indent cases in a switch-case?
```
IndentCaseLabels: true
```

PR #99, opened 2021-05-15
Implement real-world readLidarScan

nhatch:
This also changes some of the logic in autonomyIter so that we can test this by running `Rover` (i.e. with real world interface) and not worry about the rover starting to move the wheels. I'll test this on the Rover next week, with the real sensor, but I think this scaffolding should be fine to merge before that test.


src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -0,0 +1,37 @@
+
+#include "PointCloudProcessing.h"
+#include "PointGenerator.h"
+#include "URGLidar.h"
+#include "../simulator/utils.h"
+
+bool lidar_initialized = false;
+URGLidar urg_lidar;
+
+points_t readLidarScan() {
```
evan3334, line 10: If this is the implementation of `readLidarScan()` in `world_interface.h`, then `world_interface.h` should probably be included in this file.

evan3334 (changes requested):

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "PointCloudProcessing.h"
+#include "PointGenerator.h"
+#include "URGLidar.h"
+#include "../simulator/world_interface.h"
+#include "../simulator/utils.h"
+
+bool lidar_initialized = false;
+URGLidar urg_lidar;
+
+points_t readLidarScan() {
```
abhaybd, line 11: I suppose this goes for all the sensors, but could we include an out parameter, like a pointer or reference, that indicates if the returned data is fresh? Otherwise we could read the same data in consecutive loops and mistake it to be different samples.

If we want to keep the method signature the same for this method we could make the parameter have a default value like `bool *fresh=nullptr`.

nhatch: That's an interesting point. I don't actually know yet how to check whether the URG sensor is returning fresh data or not. I'll add this as a Clickup task so we remember to look into it after virtual URC: https://app.clickup.com/t/vn739w

nhatch:
I tested this with the real sensor today and fixed a few bugs. Here's a video showing it working: https://photos.app.goo.gl/cBh7xGvB7LViQESg6

evan3334:
Looks like we can probably go ahead and merge this if everyone's ready. For the sake of time we should probably look into @abhaybd's change after virtual URC. It is a good point though, I did something similar in the camera system to make sure we're getting fresh data.

PR #100, opened 2021-05-15
Implement real-world readGPS

nhatch:
I tested this on the hardware; see comment below.

### DIFF
--- a/src/gps/read_usb_gps.cpp
+++ b/src/gps/read_usb_gps.cpp
@@ -68,13 +68,15 @@ void gps_loop() {
       log(LOG_ERROR, "GPS read error.\n");
       continue;
     } else {
-      log(LOG_DEBUG, "Received fresh GPS data.\n");
       if (newdata->status == STATUS_NO_FIX) {
         log(LOG_WARN, "No GPS fix.\n");
         continue;
       } else {
         double lat = newdata->fix.latitude;
         double lon = newdata->fix.longitude;
+        point_t p = gpsToMeters__private(lon, lat);
+        log(LOG_DEBUG, "Received fresh GPS data: %.2f %.2f (lat %.6f, lon %.6f).\n",
+            p(0), p(1), lat, lon);
         gps_mutex.lock();
         if (first_fix_lat == 0.0) {
           // This is our first fix
@@ -84,7 +86,7 @@ void gps_loop() {
 
         fresh_data = true;
         // TODO no heading information
-        most_recent_tf = toTransform(gpsToMeters__private(lon, lat));
+        most_recent_tf = toTransform(p);
         gps_mutex.unlock();
       }
     }
--- a/src/gps/read_usb_gps.cpp
+++ b/src/gps/read_usb_gps.cpp
@@ -1,5 +1,7 @@
 
 #include <libgpsmm.h>
+#include <thread>
+#include <mutex>
 
 #include "../log.h"
 #include "../simulator/world_interface.h"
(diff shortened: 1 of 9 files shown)


src/gps/read_usb_gps.cpp
```diff
@@ -0,0 +1,63 @@
+
+#include <libgpsmm.h>
+
+#include "../log.h"
+#include "../simulator/world_interface.h"
+#include "../simulator/utils.h"
+
+// UW lat/long is 47.653116, -122.305619
+// source: http://www.csgnetwork.com/degreelenllavcalc.html
+const double METERS_PER_DEG_NS = 111183.53599983045;
```
evan3334, line 10: Should we put these in `Constants.h`?

evan3334 (commented):
Looks good to me, I think I just had one comment about whether we should put the constants in `Constants.h`.

evan3334:
Merged #99; we can go ahead and merge this if it's ready.

src/gps/read_usb_gps.cpp
evan3334: Also, it might be a better idea to use constexpr for these since they're primitive types.

nhatch: Unless we're planning to use these constants elsewhere in the codebase, I think it's easier to keep the constants here, close to where they're used. I've been meaning to move some of our other constants _out_ of `Constants.h`, actually.

nhatch:
I tried walking around the block with the GPS sensor connected to the Rover program. This is the path the GPS thinks I took:

Screenshot from 2021-05-22 13-19-07

It doesn't really make sense, because I think my block should be north/south axis-aligned. Also there were some `nan` readings which I need to investigate---not sure what caused those. But we're getting data of some kind! I think this is ready for review / merge.

Raw data attached
[gps_data.txt](https://github.com/huskyroboticsteam/PY2020/files/6527070/gps_data.txt)

src/gps/read_usb_gps.cpp
```diff
@@ -0,0 +1,107 @@
...
+bool startGPSThread() {
+  if (gps_rec.stream(WATCH_ENABLE|WATCH_JSON) == NULL) {
+    log(LOG_ERROR, "No GPSD running.\n");
+    return false;
+  }
+  if (!gps_rec.waiting(5000000)) {
+    log(LOG_ERROR, "No GPS hardware detected.\n");
+    return false;
+  }
+  gps_thread = std::thread(gps_loop);
```
evan3334, line 105: Does the GPS thread ever get stopped before we exit the program? One issue I ran into with the threading in the camera system was that threads will throw an exception if they're forcefully terminated instead of gracefully stopped. Maybe we should have some `running` variable that's used as the `while` loop condition in the thread, and then have a `stopGPSThread()` function that sets it to false and calls `join()` on the thread just to make sure it stops properly.

evan3334 (changes requested):
Looks good to me, just had a note regarding the GPS thread.

src/gps/read_usb_gps.cpp
nhatch: I agree there may be some advantages to explicitly joining this thread. Maybe at some point we will have a program that could call startGPSThread multiple times, so it would need to clean up after itself? However, at least on my computer, it doesn't throw any errors if I interrupt the program to kill it. In this case it might be kinder to future maintainers not to add extra code if it doesn't have some concrete benefit. What do you think?

evan3334: Oh weird. Maybe I was mistaken about the error thing? Regardless, if it works okay then it's probably fine. I agree it's best to keep the code simple for now; if it's a problem later we can cross that bridge when we come to it.

evan3334 (approved):

PR #101, opened 2021-05-21
Implement camera access system

evan3334:
Implemented a system for opening and sharing access to cameras on the rover, as well as being able to define them in a configuration file that includes information like camera parameters. The idea is that a `Camera` object, when first constructed (or when `open()` is called) will open a camera, and then share access with all other instances created using a copy constructor. 

I don't think it's 100% complete yet but it might have to do for now. In particular, one big thing that needs to be added in the future is some kind of central class/object to open all the cameras in the first place (potentially using another config file) from which other parts of the rover software can request access. For the sake of time before virtual URC I just opened one camera in `real_world_interface.cpp` manually, assuming the rest are being used for video streaming via Motion. This class could potentially publish a ROS topic or something but that's a discussion for another time probably. I also tried to write documentation where I could but it might still be lacking in some areas.

I tried not to let this branch live too long but it's still a pretty big set of changes, sorry. The shared access thing also means there's a lot of messy work with shared pointers, but I made the `CameraTester.cpp` program for testing it and it seems to work fine, as does the reworked `ARTester.cpp`.


src/camera/Camera.h
```diff
@@ -0,0 +1,295 @@
...
+
+#include "CameraParams.h"
+
+/**
+   @defgroup camera Camera Access
+   @brief This module provides an interface for defining and accessing the
+   different cameras attached to the rover.
+
+   Cameras should be defined in a YAML file with a format that will be
+   documented soon, and the location of this file should be able to be
```
nhatch, line 19: Looks like this was already documented?

src/camera/Camera.h
```diff
@@ -0,0 +1,295 @@
...
+	last one the client retrieved. The frame number will be different if two
+	frames (within a reasonable time interval) are different.
+
+	Additionally, if copied, Camera objects share ownership over a camera; when
+	the last Camera object is destroyed, the underlying camera will be
+	closed. This means multiple Camera objects may exist at different places in
+	the code that refer to the same underlying camera and share access to it.
+
+	@warning Note that intrinsic and extrinsic parameters are optional for a
+	Camera object; such Cameras may not be suitable for some computer vision
```
nhatch, line 95: "such Cameras" meaning, cameras without one or both of those parameters?

src/camera/Camera.h
```diff
@@ -0,0 +1,295 @@
...
+class Camera
+{
+private:
+	std::shared_ptr<cv::Mat> _frame;
+	std::shared_ptr<uint32_t> _frame_num;
+	std::shared_ptr<cv::VideoCapture> _capture;
+	std::string _name;
+	std::string _description;
+	std::shared_ptr<std::mutex> _frame_lock;
+	std::shared_ptr<std::mutex> _cap_lock;
```
nhatch, line 108: Is "cap" short for "capture"? Might as well use the full name `_capture_lock` I think

src/camera/Camera.h
evan3334: Woops, good catch. I wrote that way before writing the config file format docs and forgot to change it.

src/camera/Camera.h
evan3334: Yeah, that is confusing wording on second thought.

src/camera/Camera.h
evan3334: Fixed in 759bbaa2e85f16fe33c7aff486c34636fcb857bd.

src/camera/Camera.h
evan3334: Fixed in 759bbaa2e85f16fe33c7aff486c34636fcb857bd.

src/camera/Camera.h
evan3334: Good call. Fixed in eb5e340e687bcb78a0331f70a126407458b01463.

src/camera/Camera.h
```diff
@@ -0,0 +1,295 @@
...
+	   @brief Opens the camera using the given configuration file.
+
+	   @param filename The path to the configuration file to open and
+	   read. Configuration file should be formatted as described in @ref
+	   cameraconfig.
+
+	   @throws invalid_camera_config If the configuration is invalid for any
+	   reason.
+	*/
+	bool openFromConfigFile(std::string filename);
```
nhatch, line 200: I wonder if this should just be another overloaded `open()` method? I.e. `bool open(std::string config_filename);`

src/camera/Camera.h
```diff
@@ -0,0 +1,295 @@
...
+
+	   @returns true if there is a new frame (i.e. the current frame number is
+	   different than the given one) and false if not.
+	 */
+	bool hasNext(uint32_t last_frame_num) const;
+	/**
+	   @brief Retrieves the next frame.
+
+	   @param[out] frame An output parameter for the frame. Whatever is passed
+	   in will be overwritten so you do not need to worry about the passed-in
```
nhatch, line 221: I was briefly confused by this wording. Maybe instead of "Whatever is passed in will be overwritten" just say "Everything will be overwritten"

nhatch:
Okay, I looked over the header files. The API looks clean and easy to understand! Nice work. Feel free to merge whenever.

src/camera/Camera.h
evan3334: Agreed, seems better.

src/camera/Camera.h
evan3334: Yeah, perhaps. Not sure how to handle the conflict with `open(std::string filename)` though, since you can also open a camera with a filename or URI as a string, so I just made it a separate method. It was originally just a public function that lived in the `cam` namespace, but when I refactored the Camera class to add a default constructor I thought it would work better as a method of `Camera`.

nhatch: Oh I see, I didn't notice there was already an `open` method with a string argument. In that case maybe the best way is just to put those two functions next to each other in the header file / implementation file. Not super important though, obviously

PR #102, opened 2021-05-21
Added scoped timer class

abhaybd:
I'm using it to measure performance in my other branch, so I figured it would be useful to add to the repo in its own PR.


evan3334 (approved):
Looks good to me!

PR #103, opened 2021-05-21
Add PID coefficients for elbow and shoulder motors

nhatch:
We have PID control working for these two motors! The fixes were some firmware tweaks and a couple of software-side changes (such as making sure we only send the mode set packet once).

Tracking for the TunePID script has an RMSE of about 4 degrees for the shoulder and about 2 degrees for the elbow. (This is just tracking latency; there's no steady-state error.) We might be able to improve this later, but it's good enough for now.


evan3334 (changes requested):
Looks good to me; I just had one comment on the arrays of PID coefficients. We probably ought to use regular arrays instead of C++ vectors unless we really need them to be vectors.

src/Rover.cpp
```diff
@@ -33,14 +33,31 @@ const std::vector<uint32_t> arm_PPJRs = {
...
+const std::vector<int32_t> arm_Ps = {
+   0,    100,    500,     0,       0,         0
+};
+const std::vector<int32_t> arm_Is = {
+   0,      0,    300,     0,       0,         0
+};
+const std::vector<int32_t> arm_Ds = {
+   0,   1000,  10000,     0,       0,         0
+};
+const std::vector<uint8_t> arm_encoder_signs = {
```
evan3334, line 48:
```suggestion
constexpr int32_t arm_Ps[] = {
   0,    100,    500,     0,       0,         0
};
constexpr int32_t arm_Is[] = {
   0,      0,    300,     0,       0,         0
};
constexpr int32_t arm_Ds[] = {
   0,   1000,  10000,     0,       0,         0
};
constexpr uint8_t arm_encoder_signs[] = {
```

Jacob6819:
I've read over all of the changed files and they look good to me.

evan3334 (approved):

PR #104, opened 2021-05-22
Implement real-world readLandmarks()

evan3334:
This PR implements `readLandmarks()` in `real_world_interface.cpp` using the AR tag detection system. Tested on my camera in my room and it seems to work, and give reasonable estimates for distance. 

In the future we should probably have some kind of proper system (like a central "CameraManager" class) for controlling access to all the cameras on the rover but for now the world interface just opens one predefined camera (whose configuration file is defined in `Constants.h`) and uses that for AR tag detection. The camera must have intrinsic parameters, and extrinsics are recommended; if none are present then readLandmarks assumes the camera is at the center of the rover, facing straight forward.

The current policy is that readLandmarks returns the first AR tag seen of any ID. If we wanted to use a different approach (such as averaging coordinates for all tags of the same ID) that could be done as well.


src/Networking/real_world_interface.cpp
```diff
@@ -1,81 +1,160 @@
...
+	// if it is not, return a list of 11 zero points.
+	if (ar_cam.isOpen()) {
+		// if we have a next frame, retrieve it and perform AR tag detection.
+		if(ar_cam.hasNext(last_frame_no)){
+			ar_cam.next(frame, last_frame_no);
+			std::vector<AR::Tag> tags = ar_detector.detectTags(frame);
+			log(LOG_DEBUG, "readLandmarks(): %d tags spotted\n", tags.size());
+
+			// build up map with first tag of each ID spotted.
+		    points_t output(NUM_LANDMARKS);
```
nhatch, line 107: Indentation

src/Networking/real_world_interface.cpp
```diff
@@ -1,81 +1,160 @@
...
+			std::map<int, cv::Vec3d> first_tag;
```
nhatch, line 108: I might call this something different; e.g. `ids_to_camera_coords`

src/Networking/real_world_interface.cpp
```diff
@@ -1,81 +1,160 @@
...
+
+			// for every possible landmark ID, go through and if the map contains a value for
+			// that ID, add it to the output array (doing appropriate coordinate space
+			// transforms). If not, add a zero point.
+			for(int i = 0; i < NUM_LANDMARKS; i++){
+				if(first_tag.find(i) != first_tag.end()){
+					cv::Vec3d coords = first_tag[i];
+					// if we have extrinsic parameters, multiply coordinates by them to do
+					// appropriate transformation. If not, just change the coordinate axes (no
+					// rotation/translation)
```
nhatch, line 124: Arguably changing the coordinate axes _is_ a rotation. Maybe just say "no translation"?

camera-config/MastCameraCalibration.yml
```diff
@@ -27,4 +27,11 @@ intrinsic_params:
...
           -1.1179480533495888e-03, -5.7450430930239835e-04, 0. ]
-# TODO: add extrinsic parameters
+extrinsic_params: !!opencv-matrix
+  rows: 4
+  cols: 4
+  dt: d
+  data: [ 0., 0., 1., -.50,
+         -1., 0., 0., -.25,
+          0., 0., 0., 1.,
+          0., 0., 0., 0.]
```
nhatch, line 37: Do we document anywhere how to interpret this?

src/Networking/real_world_interface.cpp
```diff
@@ -1,81 +1,160 @@
...
+					if (ar_cam.hasExtrinsicParams()) {
+						cv::Vec4d coords_homogeneous = {coords[0], coords[1], coords[2], 1};
+						cv::Mat transformed =
+							ar_cam.getExtrinsicParams() * cv::Mat(coords_homogeneous);
+						output[i] = {transformed.at<double>(0, 0),
+									 transformed.at<double>(1, 0),
+									 1.0};
+					} else {
+						// just account for coordinate axis change.
```
nhatch, line 133: It might be good to expand on this comment a bit. Just to say what the camera coordinate axes are vs. the rover frame coordinate axes.

nhatch (commented):
Looks great! Some small comments about variable names and comments, but otherwise LGTM. Feel free to merge when ready

camera-config/MastCameraCalibration.yml
evan3334: Yeah, it is documented somewhat under [Top-Level Format](https://huskyroboticsteam.github.io/PY2020/cameraconfig.html#toplevel) in the camera configuration docs, but I wrote some more documentation on the specific format of the matrix that I'll push soon.

PR #105, opened 2021-05-24
Build a lidar map

abhaybd:
This PR fixes some functionality in `GlobalMap` and `QuadTree`, and also hooks them up the the path planning/autonomous code so the robot can assemble a map over time.

I still have to implement some pruning mechanisms to prevent the size of the tree from spiraling, but I'll do that in a future PR.


src/planning/plan.cpp
```diff
@@ -96,80 +97,94 @@ class NodeCompare
 
 using pqueue_t = std::priority_queue<Node*, std::vector<Node*>, NodeCompare>;
 using set_t = std::set<Node*, NodeEqualityCompare>;
+using collides_predicate_t = std::function<bool(double x, double y , double radius)>;
```
nhatch, line 100: Technically whether a pose is in collision might also depend on orientation (since the rover's not shaped like a circle), but obviously we are very far from the point where we need to worry about that.

tests/worldmap/GlobalMapTest.cpp
```diff
@@ -37,18 +38,85 @@ double calculateMSE(const points_t &p1, const points_t &p2)
 }
 } // namespace
 
-TEST_CASE("Global Map")
+TEST_CASE("Global Map - ScanStride", "[GlobalMap]")
```
nhatch, line 41: It's awesome that you're writing tests for this as you go.

nhatch (commented):
Looks pretty clean. Do you have any kind of visualization of the map quality that's generated from this code? For instance, you might be able to publish all of the map points on a new topic and visualize that in PlanViz.

It's awesome that we almost have this working. On the other hand there is relatively little time to test this out before Virtual URC (plus there's the GPS-without-heading problem). So as an FYI, I think we might want to be able to quickly disable this code if it turns out to be taking too much computation time or leading to strange plans. I think you've done a good job of that, in that this PR has relatively few changes to Autonomous.cpp.

nhatch (approved):

evan3334 (approved):
Looks good to me, I just had a minor comment on one of the functions in `plan.h` which doesn't have parameter names.

nhatch:
The drive-through-gates PR caused some merge conflicts, so I fixed those.

I tested this in simulation and found the computational issues were causing the robot to crash into obstacles. Since I don't think we should merge stuff into master that causes a regression in simulation navigation performance, I commented out the map stuff for now. That done, I think this is fine to merge @abhaybd (once you've thought about Evan's suggestion about parameter naming).

PR #107, opened 2021-05-29
Do planning in separate thread

evan3334:
Changed the autonomous code to compute the plan in a separate thread using `std::async`; I opted to make it the caller's responsibility to do planning in a different thread just so I didn't have to change any of the planning code itself. The `Autonomous` class now holds a `std::future` object which represents a pending plan, and every time it would recompute a plan it checks if one is already pending before dispatching a thread to compute one. Tested using the simulator and it seems to work fine, but I don't have much experience with the planning code so let me know if anything is wrong.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -432,9 +432,8 @@ void Autonomous::autonomyIter()
 			pending_plan =
 				std::async(std::launch::async, &computePlan, collide_func, point_t_goal);
 		}
-		// if there is a plan pending, wait maximum 1ms for it to be ready.
-		// TODO possibly change this time if we want it to be shorter.
-		else if(pending_plan.wait_for(std::chrono::milliseconds(1)) == std::future_status::ready){
+		// if there is a plan pending, check if it is ready.
+		else if(pending_plan.wait_for(std::chrono::milliseconds(0)) == std::future_status::ready){
 			// plan is ready; retrieve it and check if the cost is satisfactory.
 			// upon calling get(), pending_plan.valid() will return false, so if we reject this
 			// plan another will be computed.
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -364,6 +364,10 @@ void Autonomous::updateSearchTarget()
   }
 }
 
+plan_t computePlan(std::function<bool(double, double, double)> collide_func, const point_t& goal){
+	return getPlan(collide_func, goal, PLANNING_GOAL_REGION_RADIUS);
+}
+
 void Autonomous::autonomyIter()
 {
 	transform_t gps = readGPS(); // <--- has some heading information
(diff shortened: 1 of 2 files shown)


src/Autonomous.cpp
```diff
@@ -412,30 +416,39 @@ void Autonomous::autonomyIter()
...
-      // we want a significant improvement to avoid thrash
-      log(LOG_DEBUG, "old cost %f, new cost %f\n", plan_cost, new_plan_cost);
-      if (new_plan_cost < plan_cost * 0.8) {
-        plan_idx = 0;
-        plan_base = pose;
-        plan_cost = new_plan_cost;
-        plan = new_plan;
-        time_since_plan = 0;
-      }
+    if ((plan_cost >= INFINITE_COST || ++time_since_plan % REPLAN_PERIOD == 0)) {
```
arnavthareja, line 419: Since planning runs in a separate thread now, I think we should increase the frequency at which we replan by either removing this condition or decreasing `REPLAN_PERIOD`. I ran some tests in the simulator with this condition removed, and the computed plans were significantly better, responding to changes in `plan_target` faster and resulting in more optimal paths.

src/Autonomous.cpp
```diff
@@ -412,30 +416,39 @@ void Autonomous::autonomyIter()
...
+				// transform the point to check into map space
+				point_t relPoint = {x, y, 1};
+				point_t p = invTransform * relPoint;
+				return map.hasPointWithin(p, radius);
+			};
+			pending_plan =
+				std::async(std::launch::async, &computePlan, collide_func, point_t_goal);
+		}
+		// if there is a plan pending, check if it is ready.
+		else if(pending_plan.wait_for(std::chrono::milliseconds(0)) == std::future_status::ready){
```
arnavthareja, line 436: I think this block should be moved outside the parent if statement so that we can check if a plan is ready on every iteration and choose better plans faster. The current behavior is slow to adopt new plans and start replanning.

src/Autonomous.h
```diff
@@ -69,6 +71,7 @@ class Autonomous : rclcpp::Node
 	ControlState control_state;
 	int time_since_plan;
 	plan_t plan;
+	std::future<plan_t> pending_plan;
```
arnavthareja, line 74: Should we explicitly initialize this in the constructor?

src/Autonomous.h
```diff
@@ -3,6 +3,8 @@
 #include <cmath>
 #include <memory>
 #include <vector>
+#include <future>
+#include <thread>
```
arnavthareja, line 7: This is a minor thing, but is this import used anywhere? `async`, `future`, and `future_status` all seem to be a part of the `future` header.

arnavthareja (changes requested):
I ran some simulator tests, and the threading and planning seem to be working, but computed plans are suboptimal and replanning is infrequent. I think this will be easily fixed if we decrease `REPLAN_PERIOD` or attempt to replan on every iteration. Since the threading is working, this shouldn't cause too much of a performance hit, although it will log a lot of information.

src/Autonomous.h
evan3334: I think it only has a default constructor and a move constructor, so it shouldn't be strictly necessary, but we can do it to be explicit.

src/Autonomous.cpp
```diff
@@ -412,30 +416,39 @@ void Autonomous::autonomyIter()
...
       mapDoesOverlap = lidar_scan.size() > mapOverlapSampleThreshold;
     }
-    if (plan_cost >= INFINITE_COST || ++time_since_plan % REPLAN_PERIOD == 0) {
-      // TODO planning should happen in a separate thread
-      // because it takes a long time
-      point_t point_t_goal;
-      point_t_goal.topRows(2) = plan_target.topRows(2);
-      point_t_goal(2) = 1.0;
-      point_t_goal = toTransform(pose) * point_t_goal;
-      //plan_t new_plan = getPlan(lidar_scan, point_t_goal, PLANNING_GOAL_REGION_RADIUS);
```
nhatch, line 422: I'd like to keep this commented-out line to make it easy to revert the map construction next week (hopefully this is not necessary, but just in case)

evan3334:
Made the requested changes. Seems like a good idea to do planning on every iteration if we don't have to worry about it being costly in terms of time.

src/Autonomous.cpp
nhatch: Yeah, I think now that planning is a separate thread we can just have that thread constantly be running. Obviously we should make sure not to launch a bunch of overlapping plan computations but as long as we only have one `pending_plan` member variable I think this should be fine.

PR #108, opened 2021-05-29
Move lidar reading into separate thread

evan3334:
Added a separate thread for reading the lidar in `read_hokuyo_lidar.cpp`. Decided to add the thread in the world interface implementation instead of the `URGLidar` since I thought it would change too much of the interface (a few clients expect the `createFrame` method and I wasn't sure how to communicate the success/failure information that method returns to clients). There's currently no way of stopping the thread but hopefully that isn't a problem.

**I have no idea if this code works, since I can't test on the lidar.** It would be great if someone with access to the lab could test this and let me know if it works.

Didn't think it was necessary to move the AR stuff into a separate thread - I did some timing with `ScopedTimer` and it reports under 1µs for `readLandmarks()`. Our old AR code used to take quite a while but the ARUco library seems to be pretty efficient.


src/lidar/URGLidar.h
```diff
@@ -1,71 +1,71 @@
...
-    }
-} Polar2D;
-
-class URGLidar
-{
-private:
-    urg_t urg;
-    int error = 0;
-    const char connect_device[13] = "/dev/ttyACM0";
-    const long connect_baudrate = 115200;
```
nhatch, line 31: Good fix

src/lidar/URGLidar.cpp
```diff
@@ -1,81 +1,85 @@
...
+	// axis Add PI / 2 to the angle
+	std::vector<Polar2D> frame;
+	for (int i = 0; i < length_data_size; ++i)
+	{
+		Polar2D pd{static_cast<double>(length_data[i]), urg_index2rad(&urg, i) + M_PI / 2};
+		frame.push_back(pd);
+	}
+	lastFrame = frame;
+	error = 0;
+	free(length_data);
```
nhatch, line 71: Oh my goodness. Good catch.

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -2,48 +2,74 @@
...
+		else
+		{
+			lidar_thread = std::thread(&readLidarLoop);
+			lidar_initialized = true;
+		}
+	}
+
+	points_t pts;
+	points_lock.lock();
+	pts = last_points;
```
nhatch, line 72: It might be good to do this like for the GPS, where we track whether or not this is fresh data. Like, we don't want to use an old lidar frame from a new location---that might mess up our mapping / localization code. And if there's no fresh data, we can just return an empty list.

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -2,48 +2,74 @@
...
-      // will think it's in collision with an obstacle.
-      pts[i](2) = 0.0;
-    }
-  }
-
-  return pts;
+	if (!lidar_initialized) {
+		if (!urg_lidar.open())
+		{
+			// TODO: should we be using log() here instead?
```
nhatch, line 59: The nice thing about `perror` is that it automatically appends a string that uses the `errno` global variable to explain what the cause of the error was. We could *also* add a log message

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -2,48 +2,74 @@
...
-    {
-      // We're inside lidar min range, so this garbage data.
-      // We don't want to include this hit because otherwise the robot
-      // will think it's in collision with an obstacle.
-      pts[i](2) = 0.0;
-    }
-  }
-
-  return pts;
+	if (!lidar_initialized) {
```
nhatch, line 56: Now that we have a separate thread, it feels a little weird to do this initialization code inside `readLidarScan`. Any chance we could move the initialization code into a new public method (e.g. `initializeLidar()`) and call that from the `world_interface_init` method?

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -2,48 +2,74 @@
...
+		if (!urg_lidar.open())
+		{
+			// TODO: should we be using log() here instead?
+			perror("failed to open lidar");
+			return {};
+		}
+		else
+		{
+			lidar_thread = std::thread(&readLidarLoop);
+			lidar_initialized = true;
```
nhatch, line 66: We set `lidar_initialized` to true at the same time we start the thread, so I don't think we need the `lidar_initialized` check within the thread.

nhatch (changes requested):
There's some really great cleanup in this PR! Thanks! The rest of it looks pretty good too. I think there's a couple small things that would make it even better---see comments.

src/lidar/read_hokuyo_lidar.cpp
evan3334: Yeah, we could probably do that. Would it be in the world_interface.h then?

evan3334:
Made the requested changes. I added a new public `read_hokuyo_lidar.h` file (which is now included from `real_world_interface.cpp`) and added some functions similar to the GPS - mostly an initialization function and a function to check whether data is fresh, as well as returning the data itself.

I'm going to add some code to do the AR stuff in a separate thread shortly.

src/gps/read_usb_gps.cpp
```diff
@@ -51,6 +51,7 @@ point_t gpsToMeters(double lon, double lat) {
 transform_t readGPS() {
   gps_mutex.lock();
   transform_t tf = most_recent_tf;
+  fresh_data = false;
```
nhatch, line 54: Oh hmmm. I think I fixed this in the branch Abhay has been using but it didn't make it back to this branch. Sorry.

src/lidar/read_hokuyo_lidar.cpp
```diff
@@ -2,48 +2,82 @@
...
+#include "read_hokuyo_lidar.h"
 
 // This website may be helpful
 // https://sourceforge.net/p/urgnetwork/wiki/top_en/
 
-bool lidar_initialized = false;
+constexpr double LIDAR_MIN_RANGE = 0.015;
+
+std::atomic<bool> lidar_initialized(false);
+std::atomic<bool> lidar_fresh(false);
```
nhatch, line 18: This is neat, I didn't know about atomic bools

src/lidar/read_hokuyo_lidar.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "../simulator/utils.h"
+
+namespace lidar
+{
+	bool initializeLidar();
+	bool isLidarInitialized();
```
nhatch, line 8: I don't see an implementation for this method anywhere. We can probably just delete this declaration

nhatch (approved):

PR #109, opened 2021-05-30
Networking improvements

nhatch:
The IP addresses of the base station and the rover are different on the antenna network. This PR adds those IP addresses to the DEFAULT_FASTRTPS_PROFILES.xml (used by ROS2 for peer discovery) and changes the rover initialization code to pull the server IP address from the environment variable `SSH_CLIENT` (which makes it so we don't need to recompile the program every time the server IP address changes).

There's also a small unrelated change: we fixed the PPJR bug on the shoulder board, so we now have a better estimate for the PPJR for that joint.


evan3334 (approved):
Looks good to me.

PR #110, opened 2021-05-31
Implement incremental PID interface for arm motors

nhatch:
The base station can now send JSON packets of the format
{'type':'motor','motor':'arm_base','incremental PID speed':1.0}
    
This is similar to PWM control except that it uses feedback from the
encoders to keep the arm moving at a consistent speed (regardless of
external factors such as carrying a weight, for example).
    
This code handles tracking whether each mode is using PID or PWM control
and switching the motor board modes as necessary.
    
It also implements a little bit of the inverse kinematics version of
this: we also accept JSON packets of the format
{'type':'motor','motor':'IK_X','incremental IK speed':1.0}
except that we'll return an error saying we haven't implemented this
yet.

### DIFF
--- a/src/Networking/motor_interface.cpp
+++ b/src/Networking/motor_interface.cpp
@@ -1,6 +1,10 @@
 
 #include <string>
+#include <time.h>
+#include <sys/time.h>
+
 #include "../log.h"
+#include "../Rover.h"
 #include "CANUtils.h"
 #include "ParseCAN.h"
 extern "C"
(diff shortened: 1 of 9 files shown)


src/Networking/motor_interface.cpp
```diff
@@ -49,6 +67,76 @@ int getIndex(const std::vector<std::string> &arr, std::string &value)
   return -1;
 }
 
+bool isValidOperationMode(int motor_serial, int ik_axis, const std::string &op_mode)
+{
+  if (op_mode != "PWM target" && (motor_serial > 3 || motor_serial < 1) && ik_axis < 0)
```
evan3334, line 72: Might be better to use named constants in place of 3 and 1 here just so it's clearer what this does.

evan3334 (approved):
Sorry for the delay on reviewing this. Looks pretty good to me. Had one comment about the constants used in one place but other than that I can't see any problems.

PR #111, opened 2021-06-01
Monday changes from rover testing

nhatch:
See individual commits for descriptions!

The main update is that I tried the AR code using the rover mast camera and it successfully saw the AR tag!

Screenshot from 2021-05-31 18-59-11

### DIFF
--- a/.motion/camera0.conf
+++ b/.motion/camera0.conf
@@ -1,0 +1,2 @@
+videodevice /dev/video20
+stream_port 8080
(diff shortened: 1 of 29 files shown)


nhatch:
There are some whitespace changes here which make the diff kind of hard to read. You can hide whitespace changes to make it a little easier: https://github.com/huskyroboticsteam/PY2020/pull/111/files?diff=split&w=1

evan3334 (approved):
Looked through all the changes, looks good to me.

PR #112, opened 2021-06-03
Forward kinematics

nhatch:
This PR implements a `forward_kinematics()` method that computes the (x,y,z) position of the hand relative to the arm base, given the current arm angles as reported by the motor encoders. The constants for the lengths of the arm segments and the sign flips for the motor encoders are unknown. I'll do a follow up PR to set those constants after trying things out on the actual rover.

### DIFF
--- a/src/Networking/IK.cpp
+++ b/src/Networking/IK.cpp
@@ -1,27 +1,58 @@
 
+#include "IK.h"
 #include "ParseBaseStation.h"
 #include "motor_interface.h"
 #include "../Constants.h"
+#include "../Globals.h"
 #include <iostream>
 
-// TODO check with electronics to see if this is accurate
-double RAD_PER_INT = (2*M_PI) / ((double) (1<<16));
+// Angular positions from the motor boards are given in millidegrees
+double RAD_PER_INT = (2*M_PI) / (360*1000.0);
 
-int32_t radToInt(double d, double offset, int32_t sign_flip)
+// See IK.h for explanation of sign / offset conventions
+//                                     arm_base, shoulder,                elbow
+const std::array<int, 3> sign_flips = {1,        1,                       1};
+const std::array<double, 3> offsets = {0.0,      Constants::SHOULDER_MAX, Constants::ELBOW_MAX};
+
+int32_t radToInt(double d, int motor_serial)
 {
+  double offset = offsets[motor_serial];
+  int sign_flip = sign_flips[motor_serial];
   int32_t i = sign_flip * (d - offset) / RAD_PER_INT;
   return i;
 }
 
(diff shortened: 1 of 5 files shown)


src/Networking/IK.cpp
```diff
@@ -1,27 +1,70 @@
...
+  double x = r * cos(arm_base);
+  double y = r * sin(arm_base);
+  return {x,y,z};
+}
+
 bool ParseIKPacket(json &message) {
-  double ELBOW_LENGTH = Constants::ELBOW_LENGTH;
-  double SHOULDER_LENGTH = Constants::SHOULDER_LENGTH;
+  const double ELBOW_LENGTH = Constants::ELBOW_LENGTH;
+  const double SHOULDER_LENGTH = Constants::SHOULDER_LENGTH;
```
evan3334, line 67: Is this to make the code less wordy? Also might want to make these constexpr
```suggestion
  constexpr double ELBOW_LENGTH = Constants::ELBOW_LENGTH;
  constexpr double SHOULDER_LENGTH = Constants::SHOULDER_LENGTH;
```

src/Networking/IK.cpp
```diff
@@ -1,27 +1,70 @@
...
-// TODO check with electronics to see if this is accurate
-double RAD_PER_INT = (2*M_PI) / ((double) (1<<16));
+// Angular positions from the motor boards are given in millidegrees
+double RAD_PER_INT = (2*M_PI) / (360*1000.0);
 
-int32_t radToInt(double d, double offset, int32_t sign_flip)
+// See IK.h for explanation of sign / offset conventions
+//                                     arm_base, shoulder,                elbow
+const std::array<int, 3> sign_flips = {1,        1,                       1};
+const std::array<double, 3> offsets = {0.0,      Constants::SHOULDER_MAX, Constants::ELBOW_MAX};
```
evan3334, line 16:
```suggestion
constexpr std::array<int, 3> sign_flips = {1,        1,                       1};
constexpr std::array<double, 3> offsets = {0.0,      Constants::SHOULDER_MAX, Constants::ELBOW_MAX};
```

evan3334 (approved):
Looks good to me, I tried to double check the math and it looks good (but I didn't check it that rigorously). Had some really minor nitpicks about `constexpr` just to make sure stuff gets evaluated at compile time - not strictly necessary changes.

One question - what frame of reference are the forward kinematics coordinates in? Same rover reference frame as for navigation (+x front, +y left, +z up)?

src/Networking/IK.cpp
nhatch: Yeah, the values are squared below so it seemed worth it to save some line length. Probably not the cleanest way though. Thanks for the suggestions!

PR #113, opened 2021-06-04
Merge no-gps stuff and other fixes from on-rover testing

nhatch:
I'm sorry that this PR is such a mess. Please check individual commits for more information.

It's true that no-heading-sensor navigation doesn't quite work yet, but this is a good start.

This commit also implements readOdom by integrating the command velocities.

### DIFF
--- a/src/Networking/real_world_interface.cpp
+++ b/src/Networking/real_world_interface.cpp
@@ -67,9 +67,14 @@ void setCmdVelToIntegrate(double dtheta, double dx) {
 	odom_dx_ = dx;
 }
 
-constexpr double WHEEL_BASE = 1.0;				 // Distance between left and right wheels. Eyeballed
+// Distance between left and right wheels.
+constexpr double WHEEL_BASE = 0.66;
+// Effective distance between wheels. Tweaked so that actual rover angular rate
+// roughly matches the commanded angular rate.
+constexpr double EFF_WHEEL_BASE = 1.40;
+
 constexpr double WHEEL_RADIUS = 0.15;			 // Eyeballed
-constexpr double PWM_FOR_1RAD_PER_SEC = 10000; // Eyeballed
+constexpr double PWM_FOR_1RAD_PER_SEC = 5000; // Eyeballed
 
 bool setCmdVel(double dtheta, double dx) {
 	if (Globals::E_STOP && (dtheta != 0 || dx != 0))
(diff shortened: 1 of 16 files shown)


src/Networking/real_world_interface.cpp
```diff
@@ -25,40 +25,89 @@ extern "C"
...
 
 	/* This is the inverse of the formula:
-	 *    dx = (right_ground_vel + left_ground_vel) / 2
-	 *    dtheta = (right_ground_vel - left_ground_vel) / WHEEL_BASE
+	 *		dx = (right_ground_vel + left_ground_vel) / 2
+	 *		dtheta = (right_ground_vel - left_ground_vel) / EFF_WHEEL_BASE
 	 */
-	double right_ground_vel = dx + WHEEL_BASE / 2 * dtheta;
-	double left_ground_vel = dx - WHEEL_BASE / 2 * dtheta;
+	double right_ground_vel = dx + EFF_WHEEL_BASE / 2 * dtheta;
```
abhaybd, line 89: This is pretty minor since the calculations are pretty short, but if you want you could reuse the existing `DiffDriveKinematics` class to do these kind of calculations

abhaybd:
Looks good to me!

src/Networking/tests.cpp
```diff
@@ -125,7 +125,7 @@ TEST_CASE("Incremental PID", "[BaseStation]")
 
     REQUIRE(Globals::status_data["arm_base"]["operation_mode"] == "incremental PID speed");
     REQUIRE(Globals::status_data["arm_base"]["target_angle"] == start_angle);
-    int expected_increment = M_PI * 1000 / 8 / 10;
+    int expected_increment = 360 * 1000 * ((M_PI/8)/(2*M_PI)) / 10;
```
evan3334, line 128: Just curious, what does this expression mean in context? Isn't it equivalent to 1/16?

evan3334 (approved):
Also looks good to me, although admittedly I don't have a great understanding of the autonomous navigation stuff.

src/Networking/real_world_interface.cpp
nhatch: Yeahhh I was trying to remember where that code was. I figured it would be good practice to re-derive it, but from a software engineering standpoint we probably ought to use just one implementation

src/Networking/tests.cpp
nhatch: Yeah. I give the incremental PID rate in terms of radians per second in `motor_interface.h` as `M_PI/8`.

PR #114, opened 2021-06-04
Implement autonomous leg sequencing

arnavthareja:
This PR implements sequencing multiple autonomous legs. When the rover completes a leg, it will start the next leg once the enter key has been pressed in the terminal. This works well in the simulator, but we may want to send this as a command from the base station when used on the real rover.

This also implements parsing autonomous legs from GPS coordinates in a well-formatted file. The filepath can be changed in `Rover.cpp`. If no file is given or the file doesn't exist or contains no valid legs, the default simulator legs are used.

There are also some small refactors included.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -108,6 +108,7 @@ void Autonomous::setNavState(NavState s)
 		if (urc_targets.size() > 0)
 		{
 			setNavState(NavState::GPS);
+			search_target = {urc_targets.front().approx_GPS(0) - PI, urc_targets.front().approx_GPS(1), -PI / 2};
 			setCmdVel(0, 0);
 			Globals::AUTONOMOUS = false;
 		}
@@ -261,7 +262,7 @@ pose_t Autonomous::choose_plan_target(const pose_t &pose)
   {
     return getGPSTargetPose();
   }
-  else if (nav_state == NavState::SEARCH_PATTERN)
+  else if (nav_state == NavState::SEARCH_PATTERN || nav_state == NavState::SEARCH_PATTERN_SECOND_POST)
   {
     return search_target;
   }
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -97,13 +97,18 @@ bool Autonomous::currentLegDone()
 	return nav_state == NavState::DONE;
 }
 
+bool Autonomous::hasNextLeg()
+{
+	return urc_targets.size() > 0;
+}
+
 void Autonomous::startNextLeg()
 {
 	if (!currentLegDone())
 	{
 		log(LOG_INFO, "Cannot start next leg: current leg not done\n");
 	}
-	else if (urc_targets.size() <= 0)
+	else if (!hasNextLeg())
 	{
 		log(LOG_INFO, "Cannot start next leg: no more legs to start\n");
 	}
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -98,31 +98,28 @@ double dist(const pose_t &p1, const pose_t &p2, double theta_weight)
 	return diff.norm();
 }
 
-bool Autonomous::currentLegDone()
-{
-	return nav_state == NavState::DONE;
-}
-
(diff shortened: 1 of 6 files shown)


src/Autonomous.cpp
```diff
@@ -95,6 +95,34 @@ double dist(const pose_t &p1, const pose_t &p2, double theta_weight)
...
+bool Autonomous::hasNextLeg()
+{
+	return urc_targets.size() > 0;
+}
+
+void Autonomous::startNextLeg()
+{
+	if (!currentLegDone())
+	{
+		log(LOG_INFO, "Cannot start next leg: current leg not done\n");
```
nhatch, line 112: Since this is an unexpected case, this should probably be `LOG_WARN` instead of just `LOG_INFO`.

src/Autonomous.cpp
```diff
@@ -103,6 +131,18 @@ void Autonomous::setNavState(NavState s)
...
     log(LOG_INFO, "Changing navigation state to %s\n", NAV_STATE_NAMES[nav_state]);
     plan_cost = INFINITE_COST;
+	if (currentLegDone())
+	{
+		// Remove previous leg
+		urc_targets.pop();
+
+		// clear the cached data points
+		leftPostFilter.reset();
+		rightPostFilter.reset();
```
nhatch, line 141: As discussed below, let's remove the `currentLegDone()` method and move these three statements to the `startNextLeg()` method. Then I think we can get rid of this block of code completely, which simplifies things.

src/Rover.cpp
```diff
@@ -153,16 +192,34 @@ int rover_loop(int argc, char **argv)
...
+			while (elapsedTime < 10 * 1000)
+			{
+				long startTime = getElapsedUsecs(tp_rover_start);
+				// Blocks until a newline is found in cin
+				std::cin.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+				elapsedTime = getElapsedUsecs(tp_rover_start) - startTime;
+			}
+			autonomous.startNextLeg();
+        }
+
```
nhatch, line 222: As discussed above, I think we can avoid all of this code (lines 205-222) using the new `{"type":"autonomous"}` message from the base station. It's better not to block in the main control loop, anyway.

nhatch (changes requested):
Nice work!

I have a few specific suggestions. I also realize belatedly that it's a bit weird having the `gpsToMeters` thing in a file that's only included on the real rover, so I'll push a quick fix for that to this PR.

evan3334 (approved):
Nathan seems to have covered most of the minor bugs and organization issues, so I'm just going to approve. Aside from those issues, it looks good to me! I noticed that the result of `_targets.front()` is used directly in a lot of places so I was going to ask about the possibility of it being empty, but it seems like the invariant that the queue is not empty when these accesses happen is pretty reasonably enforced.

src/CMakeLists.txt
```diff
@@ -53,7 +53,9 @@ add_library(rover_common SHARED
   worldmap/GlobalMap.cpp
   worldmap/TrICP.cpp
   worldmap/QuadTree.cpp
+  gps/read_usb_gps.cpp
   )
+target_link_libraries(rover_common gps)
```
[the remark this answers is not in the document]
nhatch: Fixed.

arnavthareja:
Made the requested changes. I'm running into a new problem while testing where the rover almost always thinks a collision is imminent (I suspect this is related to planning or the recent simulator realism updates). When I test the changes with the collision stopping commented out, everything works as expected.

src/example_gps_legs.txt
```diff
@@ -0,0 +1,12 @@
...
+
+Leg 1:
+0 -1 47.653116 -122.305619
+
+Leg 2:
+1 2 47.653125 -122.305486
+
+3 4 47.653325 -122.325486 (Leg 3)
+
+5 -1 47.653525 -123.305486
```
nhatch, line 12: Nice, I like the demos of all the different valid syntax

nhatch (approved):
Looks great!

PR #115, opened 2021-06-04
Friday changes on the rover

nhatch:
See individual commits!

### DIFF
--- a/camera-config/cameras.rules
+++ b/camera-config/cameras.rules
@@ -3,9 +3,14 @@
 #
 # To help write these rules, try running:
 #   udevadm info -a --name /dev/video0
+#
+# Install this in /etc/udev/rules.d
 
 # Mast camera
-SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9601", SYMLINK+="video20"
+SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9601", KERNELS=="1-2.3", SYMLINK+="video20"
 
 # Forearm camera
-SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9230", SYMLINK+="video30"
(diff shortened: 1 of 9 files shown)


src/Networking/ParseBaseStation.cpp
```diff
@@ -139,7 +139,7 @@ bool ParseDrivePacket(json &message) {
   {
     return sendError("Drive targets not within bounds +/- 1.0");
   }
-  double MAX_X_VEL = 1.0; // m/s
+  double MAX_X_VEL = 0.5; // m/s
   double MAX_TH_VEL = 1.0; // rad/s
```
evan3334, line 143: If these are constants, we should probably move them to the beginning of the file

camera-config/cameras.rules
```diff
@@ -3,9 +3,14 @@
...
+#
+# Install this in /etc/udev/rules.d
 
 # Mast camera
-SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9601", SYMLINK+="video20"
+SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9601", KERNELS=="1-2.3", SYMLINK+="video20"
 
 # Forearm camera
-SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9230", SYMLINK+="video30"
+SUBSYSTEM=="video4linux", ATTRS{idVendor}=="05a3", ATTRS{idProduct}=="9601", KERNELS=="1-2.1", SYMLINK+="video30"
```
evan3334, line 13: This means that it's now important that these cameras get plugged into the same USB ports every time, right?

evan3334 (approved):
Had a few comments but looks good to me.

camera-config/cameras.rules
nhatch: Right. As far as I could tell, there was no other way to distinguish between the forearm cam and the mast cam.

Actually, I still didn't manage to get all three cameras working simultaneously. I think the udev rules worked, but `motion` was still having trouble. Here's hoping I'll be able to find a good solution tomorrow morning at 8:00....

PR #116, opened 2021-06-05
Navigate to gates and traverse them without compass

abhaybd:
Also includes some realism updates to the simulator. I've tried to minimize changes to Autonomous.cpp, since not having a compass is temporary.

I haven't had a lot of time to clean it up and verify it, but some corners had to be cut to get it done quickly.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -406,6 +406,8 @@ plan_t computePlan(std::function<bool(double, double, double)> collide_func, con
 
 void Autonomous::autonomyIter()
 {
+	transform_t gps = readGPS();
+
 #ifdef GATE_TRAVERSAL
 	if (!Globals::AUTONOMOUS)
 	{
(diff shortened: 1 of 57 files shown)


nhatch (commented):
Looks good overall (for the appropriate sense of "good"). There are a couple of things we need to do still:
- Incorporate Arnav's leg sequencing
- Handle the possibility that `readLandmarks` has no fresh data
I'll try to fix this up and then merge this PR.

It is probably unreasonable to incorporate any kind of obstacle avoidance, unfortunately, so we'll just hope we don't crash into any of the posts on our way to other posts.

src/Autonomous.cpp
```diff
@@ -50,7 +51,7 @@ Autonomous::Autonomous(const URCLeg &_target, double controlHz)
 		urc_target(_target),
 		search_target({_target.approx_GPS(0) - PI, _target.approx_GPS(1), -PI / 2}),
 		gate_targets({NAN, NAN, NAN}, {NAN, NAN, NAN}),
-		poseEstimator({1.5, 1.5}, gpsStdDev, Constants::WHEEL_BASE, 1.0 / controlHz),
+		poseEstimator({0.2, 0.2}, gpsStdDev, Constants::WHEEL_BASE, 1.0 / controlHz),
```
evan3334, line 54: In the future we might want to have this `{0.2, 0.2}` value as a constant so it is more clear what it's for when reading the code.

src/Networking/real_world_interface.cpp
```diff
@@ -75,6 +75,8 @@ constexpr double EFF_WHEEL_BASE = 1.40;
 
 constexpr double WHEEL_RADIUS = 0.15;			 // Eyeballed
 constexpr double PWM_FOR_1RAD_PER_SEC = 5000; // Eyeballed
+// This is a bit on the conservative side, but we heard an ominous popping sound at 20000.
```
evan3334: Does it still make the ominous popping sound?

evan3334 (approved):
Admittedly I don't understand a lot of the navigation/planning stuff yet, but it seems like it works in the simulator and I couldn't see any obvious issues in the code. Looks good to me!

src/Networking/real_world_interface.cpp
nhatch: Oh right. No, it does not. I left the comment as a warning to be careful about trying yet higher PWMs, but yes I should clean that up at some point.

PR #117, opened 2021-06-06
Changes made before / during vURC competition on Saturday

nhatch:
Some of these changes are a bit hacky; I will clean this up upon request but also wanted to open it up for comments.

I guess the autonomy changes are not super important because I think we're planning to redo the system before CIRC.


src/ar/read_landmarks.cpp
```diff
@@ -76,14 +76,24 @@ void detectLandmarksLoop(){
 			}
 
 			landmark_lock.lock();
-			current_landmarks = output;
+			for (int i = 0; i < NUM_LANDMARKS; i++) {
+				if (output[i](2) != 0.0) current_landmarks[i] = output[i];
+			}
```
evan3334, line 81: Just to make sure I understand this correctly, does this make the landmark detection "sticky," i.e. store all landmarks seen during the time period between calls to `readLandmarks()`, rather than replacing them every frame?

src/commands/DriveToGateNoCompass.cpp
```diff
@@ -37,7 +37,12 @@ command_t DriveToGateNoCompass::getOutput()
 	{
 		calibrationPoints.clear();
 		double err = calculateHeadingErr();
-		return {.thetaVel = err * angleKP, .xVel = 0};
+		double maxthvel = 0.5;
+		double targthvel = err * angleKP;
+		if (abs(targthvel) > maxthvel) {
+			targthvel = (targthvel > 0) ? maxthvel : -maxthvel;
+		}
```
abhaybd, line 44: Since it looks like this logic is duplicated in a few places, should it be moved to a more centralized place? Maybe we could do the clamping in `setCmdVel()`.

src/commands/DriveToGateNoCompass.cpp
```diff
@@ -37,7 +37,7 @@ command_t DriveToGateNoCompass::getOutput()
 	{
 		calibrationPoints.clear();
 		double err = calculateHeadingErr();
-		double maxthvel = 1.0;
+		double maxthvel = 0.5;
```
evan3334, line 40: Just thought of this - it's not important now, but in the future this probably ought to be in `Constants.h` since it's used in so many places.

src/commands/nogps/DriveToGate.cpp
```diff
@@ -15,7 +16,13 @@ command_t DriveToGate::getOutput()
 	{
 		double angle = atan2(post.y(), post.x());
 		double vel = (post.topRows<2>().norm() <= 2 * radius) ? slowVel : fastVel;
-		return {angle * thetaKP, vel};
+		double targthvel = angle * thetaKP;
+		double maxthvel = 0.5;
+        std::cout << "Requesting vel " << vel << std::endl;
```
evan3334, line 21: Indentation is a little funky here
```suggestion
		std::cout << "Requesting vel " << vel << std::endl;
```

src/ar/read_landmarks.cpp
abhaybd, line 81: I'm not familiar with the camera code so I definitely could be wrong about this, but wouldn't this cause stale data to stick around?

If the camera sees a target, but then the rover drives away, the next time somebody calls `readLandmarks()` it would return bad data, right?

This would be mitigated if clients call `readLandmark()` repeatedly, so there's not much time for data to become stale, but is that behavior we can we rely on?

src/commands/DriveToGateNoCompass.cpp
evan3334:
```suggestion
		// TODO move to Constants.h
		double maxthvel = 0.5;
```

src/ar/read_landmarks.cpp
evan3334: That was my question as well. I'm assuming we call `readLandmarks()` frequently enough that like a camera frame or two wouldn't make too much of a difference, but `current_landmarks` is updated by a loop in a separate thread as fast as the camera provides new frames, so theoretically, at about 30Hz. 

This seems like it's intended to mitigate the problem of the AR tag detection being spotty though - Nathan mentioned to me that tags would randomly sometimes be seen in one frame and not the next

abhaybd: Ah, I just saw your comment on the same snippet. That makes sense, though, for why we have to do it. 

Instead of relying on the client to call `readLandmarks()` frequently, would it be cleaner to assign each data point a timestamp? That way, the landmark thread can detect expired data without relying on the client, and we would get the same benefits of working around the spotty detection.

nhatch: I agree that we shouldn't rely on the client to call `readLandmarks` to clear stale data. However, coming up with the right solution would involve a bit of discussion and is not a small change. How about for this PR, I'll add a comment explaining why we're doing things this way, and I'll make a Clickup task for coming up with a better solution.

I put it as a subtask of https://app.clickup.com/t/x7cd5v

PR #119, opened 2021-06-19
Enable and fix compiler warnings

evan3334:
Turned on a lot of compiler warnings (except for ones about unused variables) and fixed a lot of problems they revealed; most of them appear to be comparisons between signed and unsigned values, which is a minor problem but I think it's best to avoid anything unexpected happening. 

I chose not to enable `-Wconversion` for now since it spits out a lot of errors for Eigen itself; however once I figure out how to stop that we can probably turn that on in a future commit/PR to discover the implicit conversion problems relating to `abs`.

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -312,7 +312,7 @@ void Autonomous::updateLandmarkInformation(const transform_t &invTransform)
 	// get landmark data and filter out invalid data points
 	points_t landmarks = readLandmarks();
 	printLandmarks(landmarks);
-	if (urc_targets.front().left_post_id < 0 || urc_targets.front().left_post_id > landmarks.size())
+	if (urc_targets.front().left_post_id < 0 || static_cast<unsigned>(urc_targets.front().left_post_id) > landmarks.size())
 	{
 		log(LOG_ERROR, "Invalid left_post_id %d\n", urc_targets.front().left_post_id);
 		return;
@@ -564,7 +564,7 @@ void Autonomous::autonomyIter()
   if (USE_MAP) {
       if (mapLoopCounter++ > mapBlindPeriod) {
           map.addPoints(toTransform(pose), lidar_scan, mapDoesOverlap ? 0.4 : 0);
-          mapDoesOverlap = lidar_scan.size() > mapOverlapSampleThreshold;
+          mapDoesOverlap = lidar_scan.size() > static_cast<unsigned>(mapOverlapSampleThreshold);
       }
   }
 
@@ -714,7 +714,7 @@ pose_t Autonomous::getGPSTargetPose() const
 static void printLandmarks(points_t& landmarks, int log_level){
 	std::ostringstream stream;
 	stream << "Landmarks: [";
-	for(int i = 0; i < landmarks.size(); i++){
+	for(size_t i = 0; i < landmarks.size(); i++){
 		stream << i << ": {"
 			   << landmarks[i][0] << ", "
 			   << landmarks[i][1] << ", "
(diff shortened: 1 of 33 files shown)


src/Autonomous.cpp
```diff
@@ -558,7 +558,7 @@ void Autonomous::autonomyIter()
   if (USE_MAP) {
       if (mapLoopCounter++ > mapBlindPeriod) {
           map.addPoints(toTransform(pose), lidar_scan, mapDoesOverlap ? 0.4 : 0);
-          mapDoesOverlap = lidar_scan.size() > mapOverlapSampleThreshold;
```
nhatch, line 561: Perhaps a cleaner way to fix this is change `mapOverlapSampleThreshold` to an unsigned int in `Autonomous.h`

src/Networking/IK.cpp
```diff
@@ -101,7 +104,7 @@ bool ParseIKPacket(json &message) {
     double cosElbowAngle = (crossSection*crossSection
       - ELBOW_LENGTH*ELBOW_LENGTH - SHOULDER_LENGTH*SHOULDER_LENGTH)
       /(-2*SHOULDER_LENGTH*ELBOW_LENGTH);
-    if (abs(cosElbowAngle) >= 1.0)
```
nhatch, line 104: I have another PR out that switches all the floating-point absolute values to `fabs` https://github.com/huskyroboticsteam/PY2020/pull/117

Do you have a strong preference which way we solve this (`fabs` vs `std::abs`)? Using `fabs` has the slight advantage that every time we read it it's a reminder that we need to be careful about this, but I don't have a terribly strong opinion.

src/Networking/tests.cpp
```diff
@@ -142,16 +142,18 @@ TEST_CASE("Incremental PID", "[BaseStation]")
     p = popCANPacket();
     REQUIRE(GetDeviceSerialNumber(&p) == 1);
     REQUIRE(PacketIsOfID(&p, ID_MOTOR_UNIT_PID_POS_TGT_SET));
-    REQUIRE(GetPIDTargetFromPacket(&p) == start_angle + expected_increment);
+	REQUIRE(GetPIDTargetFromPacket(&p) ==
+			static_cast<unsigned>(start_angle + expected_increment));
```
nhatch, line 146: PID targets can be negative, so I don't think this is a good idea.

Actually wait, the HindsightCAN library says they are unsigned. Yikes, that seems like a bug in the library. I'll ask electronics about this.

nhatch (commented):
Excellent

src/Networking/tests.cpp
nhatch: Okay I made a PR to fix this bug in the HindsightCAN repo: https://github.com/huskyroboticsteam/HindsightCAN/pull/32

src/Autonomous.cpp
evan3334: Yeah, makes sense. Better to change the variable itself if it doesn't need to be signed

src/Networking/tests.cpp
evan3334: Sounds good. That kind of didn't make sense to me as an unsigned value so it's good that we caught that.

src/Networking/IK.cpp
evan3334: `std::abs` might be nice because it will do what we're expecting by writing `abs` and select the right type automatically; however you're right that the other way has the advantage of being more explicit. I don't really have a strong preference either, but if we choose one we probably ought to use it consistently throughout the codebase (and maybe add something to the style guide about it?)

nhatch: If it doesn't matter to you, then I lean towards the `fabs` solution, just because I already implemented it. I can add a style guide update to the other PR!

PR #122, opened 2021-07-16
Improve AR tag detection somewhat

evan3334:
Improved some parts of the AR detection system; in particular, the `Detector` class now corrects passed-in images for lens distortion which should hopefully improve the ability of the system to approximate the sides of the tags using straight lines. We also made some internal changes that allow the client to change the settings of the underlying ARUco detection code after a `Detector` is constructed. 

We modified the `ARTester` to just use a plain `VideoCapture` object from OpenCV so it could read frames from the input synchronously; the camera system works well for reading live camera feeds, but if asked to stop or used on a video file, it will consume all the available input in its background thread, which makes it bad for the `ARTester`. The change made the tester program pretty messy so we need to refactor that at some point. Maybe make some option for the `Camera` class to run synchronously without using a thread? Definitely a question for after CIRC.

### DIFF
--- a/src/ar/ARTester.cpp
+++ b/src/ar/ARTester.cpp
@@ -144,6 +144,9 @@ int main(int argc, char *argv[])
 		return EXIT_FAILURE;
 	}
 
+	cap.set(cv::CAP_PROP_FRAME_WIDTH, 640);
+	cap.set(cv::CAP_PROP_FRAME_HEIGHT, 480);
+
 	std::cout << "Opening image window, press Q to quit" << std::endl;
 
 	cv::namedWindow(WINDOW_NAME);
@@ -155,6 +158,7 @@ int main(int argc, char *argv[])
 
 	bool show_grid = false;
 	int grid_spacing = 20;
+	bool show_rejected = false;
 
 	bool loop = true;
 	cv::Size imageSize = PARAMS.getImageSize();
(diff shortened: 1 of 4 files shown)


src/ar/ARTester.cpp
```diff
@@ -91,28 +104,38 @@ int main(int argc, char *argv[])
...
 	cam_config[cam::KEY_INTRINSIC_PARAMS] >> PARAMS;
+	// read filename or camera ID, and open camera.
+	if (!cam_config[cam::KEY_FILENAME].empty() && !parser.has("fo"))
+	{
+		cam_file = (std::string)cam_config[cam::KEY_FILENAME];
+	}
+	else if (!cam_config[cam::KEY_CAMERA_ID].empty() && !parser.has("co"))
+	{
+		cam_id = (int)cam_config[cam::KEY_CAMERA_ID];
+	}
```
evan3334, line 115: We're basically just reimplementing all the camera config file parsing that's already implemented in the `Camera` class here, which isn't great, so we'll definitely need to refactor this later

src/ar/ARTester.cpp
```diff
@@ -14,25 +14,33 @@
...
 	"rather than the one defined in the configuration file.}"
+	"{cam_override co    | -1     | If present, should be the ID of a camera to open "
+	"rather than the one defined in the configuration file. NOTE: Will take precedence over "
+	"file_override if both are present.}"
 	"{marker_set m       | urc    | The set of markers to look for. Currently "
 	"only \"urc\" and \"circ\" are supported.}"
 	"{frame_by_frame f   |        | If present, program will go frame-by-frame "
 	"instead of capturing continuously.}";
 
-cam::Camera cap;
```
nhatch, line 30: Yeah, in particular it would be good to be able to switch between video input vs. webcam input

nhatch (approved):
LGTM!

PR #123, opened 2021-07-17
Use pose graphs for localization

nhatch:
Our EKF assumes that we have a heading sensor, which in general we might not have. This PR implements a slightly more general solution for localization (and mapping) that uses a pose graph. The pose graph code is in the `simulator` subrepo, so this PR just updates our main autonomy code to use that API.


evan3334 (changes requested):
Seems good to me, but I requested changes on https://github.com/huskyroboticsteam/PlanarSimulator/pull/8 so we should probably wait to merge this one till that gets resolved because the submodule will need updating as well.

evan3334 (approved):

PR #124, opened 2021-07-21
Use a vector for urc_targets instead of popping from a queue

nhatch:
With a queue, once a target is popped from the queue, it disappears from memory. Using a std::vector seems a bit more flexible to me, in case we want to go back to previous targets or skip ahead to some future target.


evan3334 (approved):
Looks good to me!

PR #125, opened 2021-07-23
Three small tweaks to autonomy system

nhatch:
1. Fixes a thread safety issue with `computePlan` (we were passing the lidar points by reference, which could be an issue if the outside function returns while `computePlan` is still running).
2. Decreases epsilon for weighted A* (more expensive, but better plans)
3. Update to new `FriendlyGraph` API (and increase the std for GPS data, to try to make the pose estimates less jittery).

### DIFF
--- a/src/planning/plan.cpp
+++ b/src/planning/plan.cpp
@@ -11,6 +11,8 @@
 #include "../simulator/constants.h"
 #include "../simulator/graphics.h"
 
+const double EPSILON = 1.2; // heuristic weight for weighted A*
+
 using namespace NavSim;
 
 // TODO implement goal orientations?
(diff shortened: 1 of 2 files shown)


src/planning/plan.cpp
```diff
@@ -11,6 +11,8 @@
 #include "../simulator/constants.h"
 #include "../simulator/graphics.h"
 
+const double EPSILON = 1.2; // heuristic weight for weighted A*
```
evan3334, line 14:
```suggestion
constexpr double EPSILON = 1.2; // heuristic weight for weighted A*
```

evan3334 (approved):
Looks good to me.

PR #126, opened 2021-07-24
Implement using pose graph optimization to estimate landmark locations

nhatch:
Previously we used a filter, which essentially took a moving average of
the location estimates over time. However, now that the pose graph
solves may suddenly shift the robot's estimated location, just taking
the average of the post location estimates is not a good idea.

Fortunately, the pose graph itself can solve for landmark locations if
we give it the data. This commit implements that.


evan3334 (approved):
Looks good to me!

PR #127, opened 2021-07-31
Add science station packets to FakeCANBoard

nhatch:
I tested these on the rover and was able to move the three main motors (drill elevation, drill rotation, and soil drawer open/close).

### DIFF
--- a/src/Networking/FakeCANBoard.cpp
+++ b/src/Networking/FakeCANBoard.cpp
@@ -91,7 +91,8 @@ int main() {
     if (test_type == TEST_SCIENCE_TELEMETRY) {
         // Serial 0 seems to work. Nothing else (up to 17, the largest I tried)
         serial = prompt("Enter serial");
-        AssembleScienceSensorPullPacket(&p, science_group, (uint8_t) 0x0, 42);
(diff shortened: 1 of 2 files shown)


src/Networking/FakeCANBoard.cpp
```diff
@@ -73,5 +87,33 @@ int main() {
...
+    if (test_type == TEST_SCIENCE_TELEMETRY) {
+        // Serial 0 seems to work. Nothing else (up to 17, the largest I tried)
+        serial = prompt("Enter serial");
+        uint8_t sensor_code = 42; // Nonsense for now, just testing CAN connection
+        AssembleScienceSensorPullPacket(&p, science_group, (uint8_t) serial, sensor_code);
+        sendCANPacket(p);
+    }
+
+    if (test_type == TEST_SCIENCE_MOTORS) {
+        // CAREFUL, there are no safety checks / limit switches
```
evan3334, line 100: What should we do about the fact that there are no limit switches? We might just have to leave it up to the operator to not go over, but not having some kind of failsafe doesn't sound great :grimacing:

evan3334 (approved):
Seems good to me! Now that we've figured out a few motor IDs, should we add them in `HindsightCAN` under `CANSerialNumbers.h`?

PR #128, opened 2021-08-07
Move ROS publishing to a separate .cpp file

nhatch:
I thought that including rclcpp.h was slowing down compilation for
Rover.cpp and Autonomous.cpp, so I decided to move that into a separate
file rospub.cpp so that we don't need to recompile the publishing logic
every time we e.g. add a print statement to Autonomous.cpp for
debugging.

In the end, I'm not sure this actually improved compile times by much,
but I think it's a nice refactor regardless. Anything that shortens the
Autonomous.cpp file is a good thing, IMO....

### DIFF
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -8,6 +8,7 @@
 #include "log.h"
 #include "simulator/constants.h"
 #include "world_interface/world_interface.h"
+#include "rospub.h"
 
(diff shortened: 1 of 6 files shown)


evan3334 (approved):
Sorry for the late review, not sure how I missed this. Looks good to me, and yeah, anything that helps to modularize Autonomous.cpp is good haha

PR #133, opened 2021-11-13
Fix GPS bug and increase flexibility of SLAM solver vs. sensor availability

nhatch:
For some reason the GPS setup wasn't working (maybe related to the update to Ubuntu 20.04?). Fixed.

In a completely unrelated change, I added a MOVING_LANDMARKS compile-time constant that tells the SLAM solver whether it should expect the landmarks to be stationary. This is useful for debugging computer vision (if someone is moving around an AR target).

In addition, the SLAM solver now degrades gracefully in the absence of GPS data. It will revert to using just odometry and vision, or in the case of moving landmarks, it will rely purely on odometry (i.e. no SLAM at all).

Depends on https://github.com/huskyroboticsteam/PlanarSimulator/pull/11

There are a number of other improvements that I would like to make to the autonomy code. I will document them here but save them for a future PR:

1. Plan in odom space rather than GPS space. Otherwise the plan jumps around too much, making the robot zig-zag.
2. Make it easier to configure settings such as the initial pose estimate, initial pose covariance, and landmark covariance. These should not be compile-time constants.
3. Actually handle plan failures (e.g. when the plan target is too close to an obstacle). Currently the robot stops moving and gets stuck in an infinite loop.

### DIFF
--- a/enable_CAN_and_GPS.sh
+++ b/enable_CAN_and_GPS.sh
@@ -21,8 +21,10 @@ if [[ $? != 0 ]]; then
   sudo ip link set dev vcan0 up
 fi
 
+sudo killall gpsd
 sudo service gpsd start
 sudo service gpsd stop
+sudo systemctl stop gpsd.socket
 if [[ -e "${GPS_PATH}" ]]; then
   sudo gpsd -n "${GPS_PATH}"
 else
--- a/enable_CAN_and_GPS.sh
+++ b/enable_CAN_and_GPS.sh
@@ -21,12 +21,17 @@ if [[ $? != 0 ]]; then
   sudo ip link set dev vcan0 up
 fi
 
+# For some reason, on the rover, the gpsd that starts on boot does not
+# seem to properly publish information from our GPS.
+# Hence, we first make sure that it is not running, and then we restart it.
 sudo killall gpsd
-sudo service gpsd start
-sudo service gpsd stop
 sudo systemctl stop gpsd.socket
 if [[ -e "${GPS_PATH}" ]]; then
   sudo gpsd -n "${GPS_PATH}"
 else
+  # This code also ensures we do *not* start gpsd if there is no GPS connected,
+  # which allows for some performance optimizations. (For some reason, in the
+  # C++ gps library, it is easy to check whether gpsd is running, but it is hard
+  # to check whether gpsd is actually connected to any GPS hardware.)
   echo "No device connected at ${GPS_PATH}; gpsd not started!"
 fi
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -376,13 +376,16 @@ transform_t Autonomous::optimizePoseGraph(transform_t current_gps, transform_t c
 	}
 
 	points_t landmarks = readLandmarks();
-	printLandmarks(landmarks);
+	printLandmarks(landmarks, LOG_INFO);
 	for (size_t lm_id = 0; lm_id < landmarks.size(); lm_id++) {
 		point_t lm = landmarks[lm_id];
 		if (lm(2) != 0.0)
 			pose_graph.addLandmarkMeasurement(pose_id, (int)lm_id, lm);
 	}
-	pose_graph.addGPSMeasurement(pose_id, current_gps);
+	if (current_gps.norm() != 0.0) {
+		// TODO Currently the solver goes crazy if there is no GPS data (should add prior on x,y)
+		pose_graph.addGPSMeasurement(pose_id, current_gps);
+	}
 	pose_graph.solve();
 
 	return current_odom;
@@ -392,7 +395,7 @@ void Autonomous::autonomyIter() {
 	transform_t gps = readGPS();
 	transform_t odom = readOdom();
 
-	if ((gps.norm() != 0.0) && !pending_solve.valid()) {
+	if (!pending_solve.valid()) {
 		// TODO even if we already have a pose graph optimization running, we should still
 		// add this GPS measurement to the pose graph. However, that situation should never
 		// occur as long as pose graph solves take significantly less time than one second.
--- a/src/Autonomous.cpp
+++ b/src/Autonomous.cpp
@@ -29,6 +29,7 @@ constexpr double INFINITE_COST = 1e10;
 constexpr int REPLAN_PERIOD = 20;
 
 constexpr bool USE_MAP = false;
+constexpr bool USE_SLAM = false;
 
 constexpr bool LEFT = true;
 constexpr bool RIGHT = false;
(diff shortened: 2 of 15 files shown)


evan3334 (approved):
Seems good to me.

src/Autonomous.cpp
```diff
@@ -478,68 +514,14 @@ void Autonomous::autonomyIter() {
...
-		if (!found_target) {
-			plan_idx = plan_size - 1;
-			driveTarget = plan_pose;
-		}
-
-		// This deals with drift
-		// TODO maybe we should just replan if the distance drifts too far
-		accumulated_cost += 2 * dist(plan_pose, plan_target, 1.0);
-		plan_cost = accumulated_cost;
+		driveTarget = getDriveTargetFromPlan(pose, plan_target, lidar_scan);
```
nhatch, line 523: Sorry this part looks so confusing---I pulled out the code into this `else` block into a separate function. The code right here is not actually new, just moved.

src/Autonomous.cpp
```diff
@@ -565,21 +547,102 @@ void Autonomous::autonomyIter() {
 	}
 }
```
nhatch, line 549: Here's the new function (below)

src/CMakeLists.txt
```diff
@@ -8,23 +8,21 @@ project(Rover LANGUAGES CXX C)
...
-
+find_package(Catch2 REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(geometry_msgs REQUIRED)
 include_directories(${rclcpp_INCLUDE_DIRS})
 include_directories(${geometry_msgs_INCLUDE_DIRS})
 
 add_definitions(-DCHIP_TYPE=CHIP_TYPE_TEMPLATE)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall \
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -Wall \
```
nhatch, line 18: I figure we might as well compile with debug symbols by default. Still runs plenty fast on my machine.

abhaybd:
Given the new rules for URC, do the landmarks give any useful pose information anymore? The first 3 legs don't have any landmarks for use with CV, and the remaining legs don't have accurate GPS coordinates; it starts at 5m inaccuracy and gets as high as 20m.

Does it make sense to use them as an input for the slam solver, then?

abhaybd (approved):

enable_CAN_and_GPS.sh
```diff
@@ -21,8 +21,10 @@ if [[ $? != 0 ]]; then
   sudo ip link set dev vcan0 up
 fi
 
+sudo killall gpsd
 sudo service gpsd start
 sudo service gpsd stop
+sudo systemctl stop gpsd.socket
```
evan3334, line 27: Less important, but what's the goal of this section here? I don't entirely understand why we kill gpsd, then start it, then stop it, then stop its socket, etc. In general we should probably comment the startup script so it's clear what it's doing. 

Also, if gpsd has a systemd service then maybe we should configure it to auto-start via systemd when the Jetson boots up, thoughts?

src/CMakeLists.txt
evan3334: I think there's actually a cmake way of doing this: https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

We should probably use that; I think we can just set it to "Debug" by default if it's not explicitly set from the command line. I'm reviewing this on the light rail lol so I'll provide a code snippet later.

src/gps/read_usb_gps.cpp
```diff
@@ -6,6 +6,7 @@
 #include <libgpsmm.h>
 #include <mutex>
 #include <thread>
+#include <unistd.h>
```
evan3334, line 9: Good catch, Mac seems to choke on `usleep` if this isn't explicitly included.

evan3334 (commented):
Seems overall good to me, although admittedly I don't know enough about autonomous navigation to comment on the specifics. I also think I saw Abhay had some comments so I won't explicitly approve just yet. I do think we may want to change that one part of the CMake configuration to do a debug build; I actually think there's a `RelWithDebugInfo` option that will do optimization while adding debug symbols. I'll suggest a concrete change later, as I'm currently traveling home.

src/CMakeLists.txt
nhatch: Ah, I see: it seems you can set `-DCMAKE_BUILD_TYPE=Debug` the first time you run cmake, and it'll keep that build type for all future calls to `make`. I hadn't quite understood how that worked. I'll revert this!

enable_CAN_and_GPS.sh
nhatch: Added comments to code!

I'm a little less than enthusiastic about trying to figure out systemd right now but it might be interesting to figure that out at some point

nhatch:
> Given the new rules for URC, do the landmarks give any useful pose information anymore? The first 3 legs don't have any landmarks for use with CV, and the remaining legs don't have accurate GPS coordinates; it starts at 5m inaccuracy and gets as high as 20m.
> 
> Does it make sense to use them as an input for the slam solver, then?

It's true, the different rules for URC this year make the AR tags less useful for SLAM. A few comments about this:

1. If we don't want to use landmarks for pose information, we can set "MOVING_LANDMARKS = true".
2. Even if we have no prior information about where the landmark is, it can still be useful for localization in the vicinity of the landmark. For example, if we know the landmark is not moving and we see it from two different angles, that allows us to correct for errors in odometry between those two positions.
3. We might not want to overfit our code to the current URC rules. Maybe in CIRC or future URC competitions, AR tags will be useful for localization.

abhaybd:
> It's true, the different rules for URC this year make the AR tags less useful for SLAM. A few comments about this:
> 
> 1. If we don't want to use landmarks for pose information, we can set "MOVING_LANDMARKS = true".
> 2. Even if we have no prior information about where the landmark is, it can still be useful for localization in the vicinity of the landmark. For example, if we know the landmark is not moving and we see it from two different angles, that allows us to correct for errors in odometry between those two positions.
> 3. We might not want to overfit our code to the current URC rules. Maybe in CIRC or future URC competitions, AR tags will be useful for localization.


Ah I didn't consider those points. That makes sense!

src/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ project(Rover LANGUAGES CXX C)
 find_package(OpenCV)
```
evan3334, line 8: How about something like this? This will set a reasonable default for `CMAKE_BUILD_TYPE` if the variable is not set explicitly. It also prints a message at the very beginning clearly stating what kind of build is being done.
```suggestion
# If a build type was not explicitly specified, default to "RelWithDebInfo" (Same optimization
# as the "Release" configuration but with debugging symbols).
# To change, you can pass -DCMAKE_BUILD_TYPE=... where ... is the configuration you would
# like to use. For example, if you want to do a debug build (little to no optimization and
# debugging symbols), add -DCMAKE_BUILD_TYPE=Debug to your `cmake` command.
if(NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING
      "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
endif()
message("========================================\n\
Building ${CMAKE_BUILD_TYPE} configuration...\n\
Add -DCMAKE_BUILD_TYPE=... to change\n\
========================================")

find_package(OpenCV)
```

evan3334: Otherwise everything looks good. I'm planning a CMake reorganization PR in the future so I could always include this with that instead.

nhatch: I think maybe save it for a future PR? This one's already kind of complicated.

Also, maybe it would be simpler just to put a note in the README at the cmake step letting people know about the `CMAKE_BUILD_TYPE` option? I don't mind much either way. But I think I read somewhere yesterday that the CMAKE_BUILD_TYPE setting is generally considered "the responsibility of the user"

evan3334: Ah okay, makes sense. I'm going to go ahead and merge this then

PR #134, opened 2021-11-14
changed world_interface.h to use new DataPoint class, added header guard

abhaybd:
Related to [this task](https://app.clickup.com/t/1mzuhpw) on ClickUp.

No world_interface methods have been added or removed yet, this simply refactors existing methods to use the newly added `DataPoint` class, which provides utility for checking data freshness, representing an invalid measurement, etc. Additionally, documentation has been added to the methods in `world_interface.h`.

To check if data is new, we just have to save the old timestamp and compare against the current one. Additionally, we can also see how old the data is by checking `dataclock::now() - dataPoint.time`.

Additionally, I added a header guard, as well as standardizing time measurement utilities. Now, to measure time, we can just use `dataclock::now()`, which returns a `datatime_t`.

### DIFF
--- a/src/world_interface/world_interface.h
+++ b/src/world_interface/world_interface.h
@@ -1,16 +1,94 @@
+#pragma once
 
 #include "../simulator/utils.h"
 
-void world_interface_init(); // Call this before trying to do anything else
+#include <chrono>
+
+// the clock used for time measurements for data
+using dataclock = std::chrono::steady_clock;
+// a point in time as measured by dataclock
+using datatime_t = std::chrono::time_point<dataclock>;
+
+/**
+ * @brief Represents data measured using a sensor at a given time.
+ *
+ * @tparam T The type of data measured from the sensor. Requires defined default constructor.
+ */
+template <typename T> class DataPoint {
+public:
+    /**
+     * @brief Construct an invalid DataPoint, holding no data
+     */
+    DataPoint() : valid(false), time(), data() {}
+
+	/**
+	 * @brief Construct a new DataPoint object, measured now.
+	 *
+	 * @param data The piece of data
+	 */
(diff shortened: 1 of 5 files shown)


src/world_interface/world_interface.h
```diff
@@ -1,16 +1,94 @@
...
+    bool isFresh(std::chrono::milliseconds duration) {
+        return dataclock::now() - duration <= time;
+    }
+
+    // true iff this measurement is valid, false otherwise.
+    bool valid; 
+	// the time at which the data was measured. Use only if valid == true.
+	datatime_t time;
+	// the measurement data. Use only if valid == true.
+	T data;
```
evan3334, line 60: I know this class will only be used internally by us, but it would probably be best if these were private fields with accessor methods just to make sure we can enforce class invariants (e.g. valid should not be `true` if there's no data).

evan3334 (changes requested):
Nice, this is a good idea! Code looks mostly good to me, I just had a comment on the fields of the DataPoint class.

src/world_interface/world_interface.h
```diff
@@ -1,16 +1,94 @@
...
+    /**
+     * @brief Check if this measurement was taken recently.
+     * 
+     * @param duration The data is fresh if it was taken at most this many milliseconds ago.
+     * @return true if the data was measured at most \p duration milliseconds ago, false otherwise.
+     */
+    bool isFresh(std::chrono::milliseconds duration) {
+        return dataclock::now() - duration <= time;
+    }
+
```
evan3334, line 54:
```suggestion
private:
```

src/world_interface/world_interface.h
abhaybd: Makes sense!

evan3334 (approved):
Looks good to me!

PR #136, opened 2021-11-23
Add methods to world_interface

abhaybd:
Moved motor control methods from `motor_interface` to `world_interface`, since I think that makes more sense. Additionally, that allows code other than the networking code to control those motors.

Additionally, I added a `getCmdVel()` method, as well as a `readVisualOdomVel()` in anticipation of visual odometry.

Let me know if there are other things that should be added to `world_interface`, or if there's something else that I missed.


evan3334 (approved):
Seems good to me. I think it's a good idea to have the motors as part of the world interface since they are indeed hardware that can be simulated or physically present. Would also help us keep our CAN code somewhat confined to that one part, so Simulator developers don't need to compile it.

PR #137, opened 2021-11-24
Change clang format to only allow empty functions on single line

abhaybd:
I didn't realize in my last clang-format PR that setting `AllowShortFunctionsOnASingleLine: true` would make _all_ short functions be on a single line, which isn't ideal.

Setting it to `Empty` instead only makes empty functions be on a single line, which I think is better.

Sorry about the duplicate PR, I'm not super familiar with clang format.


evan3334 (approved):
Good catch. Seems fine to me.

PR #138, opened 2021-12-04
Added camera methods to world_interface

abhaybd:
Changed world_interface to provide camera information and camera frames. I also changed the AR code to use this interface instead of creating and owning its own camera. Additionally, I added timestamps to the frames inside `cam::Camera`.

A slightly unrelated change is that I restructured the internals of `DataPoint` to be better.


src/world_interface/real_world_interface.cpp
```diff
@@ -28,17 +29,86 @@ struct timeval last_odom_reading_;
...
+std::map<CameraID, std::shared_ptr<cam::Camera>> cameraMap;
+
+void setupCameras() {
+	try {
+		auto arCam = std::make_shared<cam::Camera>();
+		arCam->openFromConfigFile(Constants::AR_CAMERA_CONFIG_PATH);
+		cameraMap[Constants::AR_CAMERA_ID] = arCam;
+	} catch (const std::exception &e) {
+		log(LOG_ERROR, "Error opening camera with id %s:\n%s\n", Constants::AR_CAMERA_ID, e.what());
+	}
```
evan3334, line 42: We may want to map camera IDs to their config file paths in the future. Can be saved for a later PR though.

src/world_interface/real_world_interface.cpp
```diff
@@ -28,17 +29,86 @@ struct timeval last_odom_reading_;
...
+		return DataPoint<CameraFrame>{};
+	}
+}
+
+std::optional<cam::CameraParams> getCameraIntrinsicParams(CameraID cameraID) {
+	auto itr = cameraMap.find(cameraID);
+	if (itr != cameraMap.end()) {
+		auto camera = itr->second;
+		return camera->hasIntrinsicParams() ? camera->getIntrinsicParams()
+											: std::optional<cam::CameraParams>{};
```
evan3334, line 93: Could this be shortened to `{}`?

src/world_interface/real_world_interface.cpp
```diff
@@ -28,17 +29,86 @@ struct timeval last_odom_reading_;
...
+		return {};
+	}
+}
+
+std::optional<cv::Mat> getCameraExtrinsicParams(CameraID cameraID) {
+	auto itr = cameraMap.find(cameraID);
+	if (itr != cameraMap.end()) {
+		auto camera = itr->second;
+		return camera->hasExtrinsicParams() ? camera->getExtrinsicParams()
+											: std::optional<cv::Mat>{};
```
evan3334, line 105: Same here, could this be shortened to `{}`?

evan3334 (approved):
Looks good to me! I had a couple questions about various areas but otherwise it looks good to go. Sorry for the delay on this.

src/world_interface/real_world_interface.cpp
abhaybd: Looks like [brace initialization can't be used inside a ternary statement](https://stackoverflow.com/questions/18843017/why-can-braced-init-list-not-be-used-in-ternary-operator), so I don't think that can be shortened to `{}`.

src/world_interface/dummy_cv_classes.h
```diff
@@ -0,0 +1,11 @@
...
+
+namespace cam {
+// dummy CameraParams class, so the type isn't incomplete anymore
+class CameraParams {};
+} // namespace cam
+
+namespace cv {
+// same as above
+class Mat {};
+} // namespace cv
```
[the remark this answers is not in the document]
abhaybd: I realized that this solution, while it works, isn't really necessary.

My original thinking was this:
If you included opencv headers in the same compilation unit then yeah they would cause naming conflicts. However, the point is that code should either include AR/OpenCV code _or_ this header file. This header file's only purpose is to allow the implementations of `world_interface.h` that don't use AR to compile. The `readCamera()` and co. functions aren't used by anyone in these build targets, so it doesn't matter that the `cv::Mat` and `cam::CameraParams` classes are just empty classes.

However, I just realized that since nobody calls these functions, there's no need to even define them. Therefore, we can skip this trick and just remove the function definitions from world interface implementations that don't use cameras.

src/ar/read_landmarks.cpp
```diff
@@ -24,20 +25,24 @@ static points_t make_zero_landmarks() {
...
+bool initialized = false;
 
 void detectLandmarksLoop() {
 	cv::Mat frame;
 	uint32_t last_frame_no = 0;
 	while (true) {
-		if (ar_cam.hasNext(last_frame_no)) {
-			ar_cam.next(frame, last_frame_no);
+		if (hasNewCameraFrame(Constants::AR_CAMERA_ID, last_frame_no)) {
+			auto camData = readCamera(Constants::AR_CAMERA_ID);
```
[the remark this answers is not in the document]
abhaybd: I've actually been wondering that myself lol, I think that makes sense for a quick PR in the near future.

src/world_interface/real_world_interface.cpp
abhaybd: That makes sense, it would probably simplify the `setupCameras()` function too, so the only change required to add/remove cameras is to just update the map.

src/world_interface/real_world_interface.cpp
evan3334: Oh okay, we should leave it as is then.

evan3334 (approved):

PR #139, opened 2022-01-10
Reorganize the CMake setup somewhat

evan3334:
This project isn't yet finished but it's progressed enough for us to rewrite the networking code, so I'm making this pull request at this point and will finish reorganizing the CMake config later.

Our CMake config has gained a lot of targets and libraries over the years and it's starting to get a little hard to untangle. Additionally, we want to make it possible in the near future to allow the project to work natively on systems besides Linux, so we need to be able to disable some components (primarily CAN) that don't work on non-Linux systems, or might be unnecessary for certain purposes (e.g. if you're just trying to get the rover driving for a demo, or you're working on the simulator, you don't really need to have all the physical GPS and LiDAR dependencies). 

I took the approach of having configuration options that can be set at CMake configuration time and affect different parts of the build. For example, there is a `WITH_CAN` option that will be set to `OFF` by default on non-Linux systems, and if set to `OFF` will cause the `Rover` executable to be compiled with the stub CAN interface rather than the real CAN interface.

Additionally, I also started moving towards having system-specific test programs be defined in subdirectory `CMakeLists.txt` files instead of the main one, just to try to clean up the configuration a bit. For example, all the networking and lidar test programs got moved to their respective subdirectories.


abhaybd (approved):
I'm not entirely familiar with cmake, but if it builds all the targets correctly then it should be fine, right?

In either case I'll take your word for it lol

PR #140, opened 2022-01-11
Use freshness for data from sensors

abhaybd:
Reworked how data is collected from sensors and used in autonomous code to account for data freshness and validity.

For GPS: Added timestamp to gps data, so that expired data can be thrown away and duplicate datapoints can be ignored.

For LIDAR: Added timestamp to point cloud, so that expired data can be thrown away and duplicate datapoints can be ignored.

Landmark data was more complicated, since we return a vector of landmarks. The entire vector is not a single datapoint. Instead, each element of the landmarks vector is a datapoint. Therefore, instead of returning a vector of points, it returns a vector of datapoints. Now, an unseen landmark is denoted using an empty datapoint instead of {0,0,0}. Additionally, each landmark has its own associated timestamp, which allows them to independently be thrown out when expired.

Additionally, before most data getters in the world interface would only return each data point once. That is, if there is new data available, it would return that data to the first person who calls the getter, and then no other module could get that data. This isn't very good, since we want to be able to support multiple consumers of that data. Therefore, the data will stick around until new data replaces it. This means that the same data can be returned multiple times, which means consumers need to handle duplicate data points. This was implemented in `Autonomous.cpp` by tracking the last timestamp of each data type and comparing against that when new data is acquired.


src/Autonomous.cpp
```diff
@@ -51,6 +55,20 @@ const transform_t VIZ_BASE_TF =
  */
 static bool printLandmarks(const points_t& landmarks, int log_level = LOG_DEBUG);
```
evan3334, line 56: Is this form of `printLandmarks` still used anywhere?

evan3334 (approved):
Looks great to me! In particular, the landmark detection code is so much cleaner now that we don't have to deal with "zero landmarks". Also, thanks for the documentation, that will help a lot.

src/Autonomous.cpp
abhaybd: Some planner code solves for landmark locations as a vector of points, so this definition of the method is used to print those landmarks that were solved for. It's not great since ideally we want to deprecate the use of zero landmarks, but I didn't want to either convert from a vector of points to a landmarks_t, or I didn't want to refactor the planner code, so I just left it as-is.

PR #141, opened 2022-01-21
Various improvements to GPS

abhaybd:
This consists of refactoring, quality of life, and feature additions which are required for GPS use.

Now, GPS providers (sim or IRL sensor) should implement the `readGPS_private()` method instead of `readGPS()`. This method returns lat/long coordinates instead of meters, and a single piece of common code is responsible for the conversion between gps coordinates and the 2d plane.

To facilitate that conversion, GPS utils have been added. We use the WGS84 datum (specified by URC), so those parameters have been added, and a converter class has been created that takes an arbitrary datum and uses that to map gps coords <-> 2d plane.

Additionally, since the `readGPS()` method doesn't provide heading information, it's return type has been changed to `DataPoint<point_t>` so confusion is avoided. There's no point in returning a transform if it's invalid due to no heading component.

Finally, unit tests have been added for the new GPS utility classes.


src/world_interface/simulator_world.cpp
```diff
@@ -7,6 +7,11 @@
 World world;
 std::pair<double, double> cmdVel(0, 0);
 
+// gps coords of UW
+static double ORIGIN_LAT = 47.6553;
+static double ORIGIN_LON = -122.3035;
```
evan3334, line 12: Small nitpick, should probably make these `constexpr`
```suggestion
constexpr static double ORIGIN_LAT = 47.6553;
constexpr static double ORIGIN_LON = -122.3035;
```

src/gps/gps_util.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include "../simulator/utils.h"
+
+// Represents a GPS coordinate in degrees. Positive is north/east.
+struct gpscoords_t {
+	// the latitude of the gps coordinate, in degrees
+	double lat;
+	// the longitude of the gps coordinate, in degrees
+	double lon;
+};
```
evan3334, line 11: Could you convert these to Doxygen comments?
```suggestion
/** Represents a GPS coordinate in degrees. Positive is north/east. */
struct gpscoords_t {
	/** the latitude of the gps coordinate, in degrees */
	double lat;
	/** the longitude of the gps coordinate, in degrees */
	double lon;
};
```

src/gps/gps_util.h
```diff
@@ -0,0 +1,80 @@
...
+
+	// convert a coordinate on the xy-plane to a gps coordinate
+	gpscoords_t metersToGPS(const point_t& pos) const;
+
+	// Get the number of meters per degree of latitude.
+	// Given the datum and the origin, this is a constant.
+	double getMetersPerDegLat() const;
+
+	// Get the number of meters per degree of longitude.
+	// Given the datum and the origin, this is a constant.
```
evan3334, line 72: Likewise, could you convert these to Doxygen comments?
```suggestion
	/** Get the semi-major axis of the reference ellipsoid */
	double getA() const;

	/** Get the semi-minor axis of the reference ellipsoid */
	double getB() const;

	/** Get the flattening parameter of the reference ellipsoid */
	double getFlattening() const;

	/** Get the square of the (first) eccentricity of the ellipsoid */
	double getSquareEccentricity() const;

private:
	const double a;
	const double b;
};

/**
 * @brief A class used to convert gps coordinates to coordinates on a flat xy-plane, and vice
 * versa. The xy-plane will always be measured in meters. Additionally, the xy-plane will be
 * aligned such that +x = +lat, +y = -lon.
 */
class GPSToMetersConverter {
public:
	/**
	 * @brief Construct a new GPSToMetersConverter object.
	 *
	 * @param datum The datum to use for the GPS calculations.
	 * @param origin The gps coordinates of the origin of the xy-plane.
	 */
	GPSToMetersConverter(const GPSDatum& datum, const gpscoords_t& origin);

	/** convert the given gps coordinates to a coordinate on the xy plane, in meters */
	point_t gpsToMeters(const gpscoords_t& coords) const;

	/** convert a coordinate on the xy-plane to a gps coordinate */
	gpscoords_t metersToGPS(const point_t& pos) const;

	/**
	 * Get the number of meters per degree of latitude.
	 * Given the datum and the origin, this is a constant.
	 */
	double getMetersPerDegLat() const;


	/**
	 * Get the number of meters per degree of longitude.
	 * Given the datum and the origin, this is a constant.
	 */
```

src/world_interface/world_interface.h
```diff
@@ -69,20 +70,28 @@ std::optional<cv::Mat> getCameraExtrinsicParams(CameraID camera);
...
 // Calculates the current transform in the global frame based purely on forward kinematics
 DataPoint<transform_t> readOdom();
 
 // Calculate the current pose velocity (in the robot's reference frame) using visual odometry.
 DataPoint<pose_t> readVisualOdomVel();
 
-// Given the current longitude and latitude, convert to a point_t position representation
-point_t gpsToMeters(double lon, double lat);
-
 // `index` must be in the range 0-6 (the URC competition will have 7 legs)
```
evan3334, line 95: Same here, could you make these Doxygen comments?

evan3334 (changes requested):
Looks pretty good to me; just had a couple nitpicks with regards to comment style, etc. 

Also wanted to ask, do you think we should put all the GPS-related functions into a `gps` namespace for organization? There isn't any specific benefit I can think of other than organization.

abhaybd:
> Looks pretty good to me; just had a couple nitpicks with regards to comment style, etc.
> 
> Also wanted to ask, do you think we should put all the GPS-related functions into a `gps` namespace for organization? There isn't any specific benefit I can think of other than organization.

@evan3334 

I think it makes sense to use namespaces in the code a lot more, I think you mentioned before that the world interface methods get confusing because it's not clear that they're defined in another file when they're used.

When you said put "all gps-related functions into a `gps` namespace", would that include world interface methods like `readGPS()` and `gpsHasFix()`? Or would those go under the world interface namespace (once we create that in the future)

I think it makes sense for the non-world interface stuff to go in a `gps` namespace though, like `readGPS_private()` and `startGPSThread()` and the like. Should I do that in this PR or lump it in with the refactoring task on clickup to do in the future?

evan3334:
I'd say maybe just do the non-world interface methods in this PR, and then we can create the world interface namespace later. Might as well refactor this small bit of it now while it's still manageable

abhaybd:
@evan3334 Makes sense. I renamed `readGPS_private()` to `readGPSCoords()` and put it in the `gps` namespace, and put the usb gps specific methods in the `gps::usb` namespace.

evan3334 (approved):
Looks good to me!

PR #142, opened 2022-01-22
Install nlohmann/json as a dependency

abhaybd:
Before, the json library used to be in our source code as `Networking/json.hpp`. However, it's better practice to have our dependencies be declared in our cmake instead of including them in our own code.

Therefore, this PR removes `Networking/json.hpp`, declares the dependency in cmake, updates the install instructions in the readme, and updates the CI with the new dependency.


evan3334 (approved):
Good call! Looks good to me.

PR #143, opened 2022-01-29
Create websocket server

abhaybd:
This PR creates a standalone WS server for use with the Rover networking to mission control, sim, and the dgps station.

Clients can add protocols to the server, which are served at different endpoints, differentiated by the path. For example, `/foo` is one, and `/bar` would be another.

This PR also adds a default server in the `Globals` namespace.


src/Networking/websocket/WebSocketProtocol.h
```diff
@@ -0,0 +1,90 @@
...
+using nlohmann::json;
+
+/**
+ * @brief Defines a protocol which will be served at an endpoint of a server.
+ * This protocol defines several message types, and processes messages of those types,
+ * dispatching events to message handlers wherever necessary.
+ */
+class WebSocketProtocol {
+public:
+	WebSocketProtocol(const std::string& protocolPath);
```
evan3334, line 22: This should be documented - in particular, what is the expected format of `protocolPath`? Should it have slashes (leading, trailing)?

evan3334 (changes requested):
Looks good to me! Just had one comment about the constructor in the `WebSocketProtocol` class but otherwise it is good.

evan3334 (approved):
Looks good to me!

PR #145, opened 2022-02-02
Bug fixes and improvements to WS server

abhaybd:
This PR introduces a feature that is required for the simulator interface, as well as a bug fix and some minor improvements.

1. Fixed a bug where the client field associated with a protocol wasn't reset when a client disconnects.
2. Minor improvements to logging format.
3. Minor improvements to documentation.
4. Added connection and disconnection event handlers for protocols, which are necessary in some cases.


evan3334 (approved):
Good catch! Looks good to me.

PR #148, opened 2022-02-10
Separate logic of reading camera config from opening camera HW

abhaybd:
Reader the camera config previously couldn't be done without actually opening a hardware camera. This isn't ideal, particularly for the simulator.

This PR adds the `CameraConfig` struct which contains the information stored in a camera config YAML file, as well as a `readConfigFromFile()` method which reads the config from the file.

### DIFF
--- a/src/camera/CameraConfig.h
+++ b/src/camera/CameraConfig.h
@@ -34,15 +34,48 @@ private:
 	std::string _msg;
 };
 
+/**
+ * @brief A struct that represents the information outlined in @ref cameraparams.
+ */
 struct CameraConfig {
+	/**
+	 * @brief The name of the camera.
+	 */
 	std::string name;
+	/**
+	 * @brief If specified, gives the intrinsic parameter matrix.
+	 */
 	std::optional<CameraParams> intrinsicParams;
+	/**
+	 * @brief If specified, gives the extrinsic parameter matrix.
+	 */
 	std::optional<cv::Mat> extrinsicParams;
+	/**
+	 * @brief If specified, gives the file to which the camera should be streamed.
+	 */
 	std::optional<std::string> filename;
+	/**
+	 * @brief If specified, gives the id of the camera.
+	 */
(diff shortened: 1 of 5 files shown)


src/camera/CameraConfig.h
```diff
@@ -0,0 +1,112 @@
...
+	/**
+	 * @brief If specified, gives the intrinsic parameter matrix.
+	 */
+	std::optional<CameraParams> intrinsicParams;
+	/**
+	 * @brief If specified, gives the extrinsic parameter matrix.
+	 */
+	std::optional<cv::Mat> extrinsicParams;
+	/**
+	 * @brief If specified, gives the file to which the camera should be streamed.
```
evan3334, line 85: Should probably be "from which the camera should be streamed"
```suggestion
	 * @brief If specified, gives the file from which the camera should be streamed.
```

src/camera/CameraConfig.h
```diff
@@ -0,0 +1,112 @@
...
+	 */
+	std::optional<std::string> filename;
+	/**
+	 * @brief If specified, gives the id of the camera.
+	 */
+	std::optional<int> cameraID;
```
evan3334, line 91: Also something to note is that filename and camera ID are mutually exclusive, so a union or variant would probably have been more appropriate here - poor design decision on my part (I didn't know about either when I first wrote the camera system). We don't have to fix it now, but just a note for later

evan3334 (approved):
Looks good to me - only had a note about one word in the docs. Otherwise, seems fine!

src/camera/CameraConfig.h
abhaybd: Huh, I figured a union wouldn't be correct because it's not typesafe but I didn't know std::variant was a thing. I'll add that to this PR, it shouldn't be too much work.

evan3334 (approved):
Looks good to me!

PR #151, opened 2022-02-15
Update ar tag

johnli01:
- Updated AR Tag for the 2022 season
- Removed unnecessary ALVAR functions and variables

### DIFF
--- a/src/ar/MarkerSet.cpp
+++ b/src/ar/MarkerSet.cpp
@@ -110,7 +110,7 @@ constexpr size_t ARUCO_BIT_SIZE = 4;
 /** The border size in pixels of one side of an ARUco marker. */
 constexpr size_t ARUCO_BORDER_SIZE = 1;
 /** The physical size (in mm) of an ARUco marker. */
-constexpr float ARUCO_PHYS_SIZE = 0.1; // we don't know this yet, change later
+constexpr float ARUCO_PHYS_SIZE = 0.15; // we don't know this yet, change later
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/ar/MarkerSet.cpp
```diff
@@ -105,185 +105,27 @@ bool MarkerSet::getMarkerByMappedID(int mapped_id, MarkerPattern& out) const {
...
-constexpr float ALVAR_PHYS_SIZE = 0.2;
-/** The bit array size for one ALVAR marker, equal to the bit size squared. */
-constexpr size_t BIT_ARR_SIZE = ALVAR_BIT_SIZE * ALVAR_BIT_SIZE;
-
 /** The size in pixels of the data region of an ARUco marker. */
 constexpr size_t ARUCO_BIT_SIZE = 4;
 /** The border size in pixels of one side of an ARUco marker. */
 constexpr size_t ARUCO_BORDER_SIZE = 1;
 /** The physical size (in mm) of an ARUco marker. */
 constexpr float ARUCO_PHYS_SIZE = 0.1; // we don't know this yet, change later
```
evan3334, line 113: This should probably be `0.15`, as we now know the size of the markers will be 15cm (20cm, minus 2.5cm of whitespace on both sides)
```suggestion
constexpr float ARUCO_PHYS_SIZE = 0.15; // we don't know this yet, change later
```

evan3334 (approved):
Looks good to me! I had one correction for the marker physical size, but otherwise it looks good.

PR #153, opened 2022-02-21
Remove old simulator

abhaybd:
This is a fairly large PR since it removes the `simulator` submodule, which was depended on by basically most files in this project.

The `simulator/utils.h` file defined useful typedefs as well as utility functions. The relevant utility functions have been moved to `Util.h` and the types have been added under the `navtypes` namespace in `navtypes.h`.

The pose graph code also lived in the `simulator` module, so those have been pulled into the `filters::pose_graph` namespace. To fit with this style, the other pose estimator classes were moved into the `filters` namespace.

Beyond that, a few constants have been added to `Constants.h`, which were previously in `simulator/constants.h`.

Additionally, the old planar simulator has been removed entirely, and its dependencies were removed from CMake and the instructions. Other removed files include plan viz, as well as a few old command files, where it was easier to just delete the file instead of updating it to use not use the `simulator` folder anymore.

Lastly, as a minor fix the time tracking in `Rover.cpp` was updated to use `std::chrono` instead of C-style idioms.


src/Util.h
```diff
@@ -62,7 +64,56 @@ class ScopedTimer {
...
+
+navtypes::transform_t toTransform(const navtypes::pose_t& pose);
+
+// One sample from a standard normal distribution.
+// The thread_id is used to choose which random number generator to use.
+// This is important when trying to rerun a particular random seed; each thread
+// needs its own dedicated sequence of random numbers.
+double stdn(int thread_id);
+// Returns the random seed used for the stdn() function
+long getNormalSeed();
```
evan3334, line 117: Can we change these to doxygen comments?
```suggestion
/**
   @brief Given a transform, gives the robot pose that would have that transform.
   
   That is, the x,y location and angle theta of the robot such that left-multiplying a
   world-frame landmark location by `trf` will give you the robot-frame landmark location.

   @warning THIS IS AN INVERSE OF toTransform(), NOT toTransformRotateFirst(). 
   The parameters of toTransformRotateFirst() are not a robot pose
   (rotation and translation operations do not commute).
*/
navtypes::pose_t toPose(const navtypes::transform_t& trf, double prev_theta);
double nearestHeadingBranch(double theta, double prev_theta);

navtypes::transform_t toTransform(const navtypes::pose_t& pose);

/**
  @brief Gets one sample from a standard normal distribution.
 
  @param thread_id Used to choose which random number generator to use; 
  this is important when trying to rerun a particular random seed; each thread
  needs its own dedicated sequence of random numbers.
*/
double stdn(int thread_id);
/**
 Returns the random seed used for the stdn() function
*/
long getNormalSeed();
```

src/rospub.h
```diff
@@ -24,7 +24,7 @@ NLOHMANN_JSON_SERIALIZE_ENUM(ArrayPub, {{AP_INVALID, nullptr},
 
 void init();
 void shutdown();
-void publish(const pose_t& pose, PointPub topic); // Also can be used for point_t
-void publish_array(const points_t& points, ArrayPub topic);
+void publish(const navtypes::pose_t& pose, PointPub topic); // Also can be used for point_t
+void publish_array(const navtypes::points_t& points, ArrayPub topic);
```
evan3334, line 28: Do we actually still need `rospub.cpp`/`rospub.h` anymore? They currently just implement a custom websocket protocol I made to replace ROS and bridge the gap while we were waiting for the rest of the protocol stuff to get done. As a result they have a bit of a weird interface that made more sense for ROS. Now that we're implementing PlanViz in Mission Control we should probably replace this with the appropriate Mission Control protocol methods for publishing that. We can leave it for now but at some point we should remove this

evan3334 (changes requested):
Looks good to me - just had a couple notes in some places.

src/rospub.h
abhaybd: Yeah I figured the rospub stuff is obsolete now, but I wasn't sure if it should be removed now or after the mission control interface was merged.

evan3334: Yeah, let's just leave it for now - I can remove it when I implement the PlanViz part on the rover side.

src/Util.h
abhaybd: Good catch, I totally forgot to add docs and clean up the utility functions. I've added documentation to everything, and I've removed a few vestigial methods that we can't use anymore.

evan3334 (approved):

PR #154, opened 2022-02-23
Minor bug fix to simulator setCmdVel()

abhaybd:
The PWM is normalized, so it should be compared to 1, not to tangential speed, which is wrong anyway.


evan3334 (approved):
Good catch!

PR #155, opened 2022-02-26
Don't block rover while waiting for gps fix

abhaybd:
Before, the rover would block during initialization while waiting for a fix. This is fine during the competition, but makes testing unwieldy.

Instead, the GPS coordinates of the map legs are lazily mapped to world space once the gps fix becomes available. Since the pose graph initialization required this too, the pose graph is lazily initialized too.

The robot will be automatically initialized by `autonomyIter()` when the gps fix becomes available.


evan3334 (approved):
Seems good to me

PR #156, opened 2022-04-20
Implemented RP Lidar to world interface

johnli01:
Finished implementing RP Lidar to world interface and it now works with MainVisualizer.
Made some quick changes to ease in running the program at a certain baudrate.

Note: I wasn't able to test the quality of life changes with the baudrate because my RPLidar sdk doesn't have the baudrate constructor


src/lidar/read_rp_lidar.h
```diff
@@ -0,0 +1,24 @@
...
+
+typedef struct Polar2D {
+	double r, theta;
+	double asCartesianX() {
+		return r * cos(theta);
+	}
+	double asCartesianY() {
+		return r * sin(theta);
+	}
+} Polar2D;
```
evan3334, line 22: TODO: We might want to clean this up in the (near) future - this class is duplicated between RPLidar and Hokuyo Lidar. Maybe we could include casting operators in the class to ease conversion between Polar2D and PointXY (although for the latter we should be using `navtypes::point_t`).

evan3334 (changes requested):
Looks pretty good but I think we may want to do a little more cleanup (plus use the new baudrate constructor for RPLidar). Most of it is I think we should move some literals to `Constants.h` so they can all be defined in one place; longer term, we may want to clean up some of the stuff in `PointCloudProcessing.h` (e.g. `PointXY`, `Polar2D` having a private implementation that has to be repeated across lidar interfaces, etc).

src/lidar/LidarVis.h
```diff
@@ -11,7 +12,7 @@
 #include <opencv2/imgproc.hpp>
 
 namespace lidar {
-constexpr int lidar_max_range = 5000;
+constexpr int lidar_max_range = 10000;
```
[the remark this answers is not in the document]
johnli01: Should probably also make this aspect configurable so that it works with lidars that could exceed this range

src/lidar/MainVis.cpp
```diff
@@ -45,3 +52,93 @@ int main(int argc, char** argv) {
...
+		}
+	}
+	return 0;
+}
+
+static void help() {
+	printf("Lidar Visualizer. \n"
+		"Usage: \n"
+		"     -l=<lidar_num>         # 1=hokyuo, 2=rplidar"
+		"     -b=<baudrate>          # set baudrate, defaults to 115200"
```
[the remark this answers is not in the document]
johnli01: Yeah this is parsed by LidarVis, didn't know we even had a CommandLineOptions but could look into that in the future.

src/lidar/read_rp_lidar.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <atomic>
+#include <mutex>
+#include <thread>
+#include <iostream>
+
+using namespace navtypes;
+
+std::atomic<bool> lidar_initialized(false);
+RPLidar rp_lidar("/dev/ttyUSB0");
```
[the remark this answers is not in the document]
johnli01: Sounds good, I didn't really know how to set up that udev rules but I agree that having a constant with the same path would be WAY better than this.

src/lidar/read_rp_lidar.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace lidar {
+
+bool initializeLidar() {
+    if (!lidar_initialized) {
+        if (!rp_lidar.checkHealth()) {
+            perror("failed to connect to rp lidar");
+        }
+
+        rp_lidar.setBaudrate(115200);
```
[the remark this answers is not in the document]
johnli01: Oops, I forgot to use it in the read_rp_lidar.cpp but only changed the one in MainVis.

src/lidar/read_rp_lidar.h
johnli01: Sounds good, there is some redundancy that should be cleaned up.

abhaybd (changes requested):
Whoops I got a little carried away with the comments lol

Overall it looks pretty good though, good work getting this going! I've just noted some fixes that I thought would help.

Have the installation instructions changed at all, or is the rplidar library bundled with the rest of the stuff?

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
...
 
+/**
+ * @brief Startsup RPLidar with default settings
+ */
+bool initializeLidar() {
+    if (!lidar_initialized) {
+        if (!rp_lidar.checkHealth()) {
+            perror("failed to connect to rp lidar");
+        } else {
+            rp_lidar.setMaxDistance(16.0);
```
abhaybd, line 69: Is 16 a hardware limit or is it an arbitrary limiter? If the former this should be a constant at the top, and if the latter then we could either make it a constant or a parameter to `initializeLidar`. Thoughts?

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
...
-            // Converts to regular points
-            std::vector<point_t> pts(currFrames.size());
-            point_t origin({0, 0, 1});
-            for (unsigned long i = 0; i < currFrames.size(); i++) {
-                pts[i] = lidar::polarToCartesian2(currFrames[i]);
-            }
             points_lock.lock();
-            last_points = pts;
+            datatime_t time = lidar_time;
+            last_points = {time, pts};
```
abhaybd, line 48: This is almost correct. `lidar_time` isn't being updated in this loop, so this will cause the data's timestamp to never change.

Instead, the `DataPoint` class has a constructor that uses the current time as the timestamp, so you can just replace these lines with `last_points = {pts}` and that will use the current time as the new timestamp.

Then, you can remove the `lidar_time` variable since it's not used anymore.

abhaybd, line 48: This is almost correct. `lidar_time` isn't being updated in this loop, so this will cause the data's timestamp to never change.

Instead, the `DataPoint` class has a constructor that uses the current time as the timestamp, so you can just replace these lines with `last_points = {pts}` and that will use the current time as the new timestamp.

Then, you can remove the `lidar_time` variable since it's not used anymore.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
...
             points_lock.unlock();
-            lidar_initialized = true;
         } else {
             perror("failed to get frame");
+            points_lock.lock();
+            datatime_t time = lidar_time;
+            last_points = {};
+            points_lock.unlock();
```
abhaybd, line 55: If the lidar fails to get a frame it might make sense to leave `last_points` as its current value. Clients will see that the timestamp hasn't changed so they'll know that the data is stale.

If we set an empty data point then the previous data is no longer available, even if it's still relatively new.

abhaybd, line 55: If the lidar fails to get a frame it might make sense to leave `last_points` as its current value. Clients will see that the timestamp hasn't changed so they'll know that the data is stale.

If we set an empty data point then the previous data is no longer available, even if it's still relatively new.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
...
+            lidar_initialized = true;
+            points_lock.unlock();
+        }
+    }
+    return lidar_initialized;
+}
+
+/**
+ * @brief Provides RP Lidar data at current timeframe
+ */
```
abhaybd, line 81: Same as above, can you move this to the header?

abhaybd, line 81: Same as above, can you move this to the header?

src/lidar/read_rp_lidar.h
```diff
@@ -9,16 +9,16 @@ namespace lidar {
...
+struct Polar2D {
 	double r, theta;
 	double asCartesianX() {
 		return r * cos(theta);
 	}
 	double asCartesianY() {
 		return r * sin(theta);
 	}
-} Polar2D;
+};
```
abhaybd, line 22: Can you add comments for this struct too? Just minor stuff like degrees vs rads or something would be good.

abhaybd, line 22: Can you add comments for this struct too? Just minor stuff like degrees vs rads or something would be good.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
...
+DataPoint<points_t> readLidar() {
+    if (lidar_initialized) {
+        DataPoint<points_t> pts;
+        points_lock.lock();
+        pts = last_points;
+        points_lock.unlock();
+        return pts;
+    }
+    return {};
+}
```
abhaybd, line 91: If you make the change suggested above wrt `initializeLidar()` then `points_lock` should be locked before the if statement to avoid race conditions.

If it helps you could use the scoped [`lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard) instead of manually calling `lock()` and `unlock()`, but this way is fine too.

abhaybd, line 91: If you make the change suggested above wrt `initializeLidar()` then `points_lock` should be locked before the if statement to avoid race conditions.

If it helps you could use the scoped [`lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard) instead of manually calling `lock()` and `unlock()`, but this way is fine too.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,60 +12,84 @@
 
 using namespace navtypes;
 
-std::atomic<bool> lidar_initialized(false);
 constexpr uint32_t RPLIDAR_BAUDRATE = 115200;
+const double MM_TO_M = 1000;
+
+std::atomic<bool> lidar_initialized(false);
+std::atomic<datatime_t> lidar_time(datatime_t{});
```
abhaybd, line 19: As mentioned below, `lidar_initialized` doesn't need to be atomic, and `lidar_time` isn't needed. Read the comments below for reasoning and shit

abhaybd, line 19: As mentioned below, `lidar_initialized` doesn't need to be atomic, and `lidar_time` isn't needed. Read the comments below for reasoning and shit

abhaybd (changes requested):
Pretty good! Just some things that I probably missed last time and some minor notes about your latest changes.

After these changes I think it should be looking good.

Let me know if my comments are wrong or if you have any questions!

src/lidar/MainVis.cpp
```diff
@@ -109,8 +102,8 @@ int runRPLidar(unsigned long baudrate) {
 static void help() {
 	printf("Lidar Visualizer. \n"
 		"Usage: \n"
-		"     -l=<lidar_num>         # 1=hokyuo, 2=rplidar"
-		"     -b=<baudrate>          # set baudrate, defaults to 115200"
+		"     -l=<lidar_num>         # 1=hokyuo, 2=rplidar \n"
+		"     -b=<baudrate>          # set baudrate, defaults to 115200 \n"
```
evan3334, line 106: This help message implies that you can use the flags `-l` and `-b` with the program, but is it actually parsing these flags? The code in `main()` seems to imply that it is just getting the second and third positional arguments, which is fine, but we should probably change the help message to specify this. 
`Usage: lidar_vis <lidar_num> <baud_rate>` would probably work fine.

Also, rather than using a `lidar_num`, it might be more intuitive to use pass a `lidar_type` as a string, either `"hokuyo"` or `"rplidar"`; you can construct a string with `std::string lidar_type(argv[1])` and then just check `lidar_type == "hokuyo"`, etc.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,61 +12,84 @@
 
 using namespace navtypes;
 
+constexpr uint32_t RPLIDAR_BAUDRATE = 115200;
+const double MM_TO_M = 1000;
+
 std::atomic<bool> lidar_initialized(false);
-RPLidar rp_lidar("/dev/ttyUSB0");
+std::atomic<datatime_t> lidar_time(datatime_t{});
+RPLidar rp_lidar("/dev/ttyUSB0", RPLIDAR_BAUDRATE);
```
evan3334, line 20: We should probably store `"/dev/ttyUSB0"` in a constant in `Constants.h` as well.

evan3334 (changes requested):

src/lidar/MainVis.cpp
```diff
@@ -66,24 +66,17 @@ int runRPLidar(unsigned long baudrate) {
...
 		vis.clear();
         if (auto scan = rp_lidar.poll()) {
-			// Converts Scanned data into Polar Coord
-			std::vector<Polar2D> currFrames;
+			std::vector<PointXY> pts;
 			double dtheta = (scan.value().angle_max-scan.value().angle_min)/(scan.value().ranges.size()-1);
 			for (long unsigned i = 0; i < scan.value().ranges.size(); i++) {
 				double rad = dtheta*i;
-				double dist = scan.value().ranges[i] * 1000;
+				double MM_TO_M = 1000;
```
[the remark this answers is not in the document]
johnli01: Oh the measurements isn't entirely specified but dist is in meters, and scan.value().ranges[i] is in millimeters. I needed to multiply scan.value().ranges[i] by 1000 in order to convert it to meters. I'll make the variable more specific so it's less confusing to read.

src/lidar/read_rp_lidar.cpp
johnli01: When I was testing with the A1 and A2, the limits were around 12m and 16m. I'm all for using the latter since the range for different RP Lidar's ranges, and if were testing with other RP Lidars, we should be able to increase this limit. I'll go ahead and implement an optional parameter for the client.

src/lidar/MainVis.cpp
```diff
@@ -70,8 +76,8 @@ int runRPLidar(unsigned long baudrate) {
 			double dtheta = (scan.value().angle_max-scan.value().angle_min)/(scan.value().ranges.size()-1);
 			for (long unsigned i = 0; i < scan.value().ranges.size(); i++) {
 				double rad = dtheta*i;
-				double MM_TO_M = 1000;
-				double dist = scan.value().ranges[i] * MM_TO_M;
+				double MM_PER_M = 1000;
```
evan3334, line 79: Can you make this `constexpr`?
```suggestion
				constexpr double MM_PER_M = 1000;
```

evan3334: I think you've defined this `MM_PER_M` constant a few times in different `.cpp` files, so it is probably best to move it to `Constants.h` (and make it `constexpr`).

src/lidar/read_rp_lidar.cpp
```diff
@@ -61,16 +53,15 @@ void readLidarLoop() {
...
  */
-bool initializeLidar() {
+bool initializeLidar(double MAX_DIST = 16) {
     if (!lidar_initialized) {
         if (!rp_lidar.checkHealth()) {
             perror("failed to connect to rp lidar");
         } else {
-            rp_lidar.setMaxDistance(16.0);
-            points_lock.lock();
+            rp_lidar.setMaxDistance(MAX_DIST);
```
evan3334, line 61: Probably shouldn't capitalize `MAX_DIST` if it is not a constant.

evan3334 (changes requested):
Couple of super nitpicky things I noticed, but otherwise I think this looks great!

src/Constants.h
```diff
@@ -45,4 +45,12 @@ const double GPS_WAYPOINT_RADIUS = Constants::ROBOT_LENGTH * 1.5;
 const double LANDMARK_WAYPOINT_RADIUS = Constants::ROBOT_LENGTH * 1.3;
 const double EPS = 2.0; // heuristic weight for weighted A*
 }
+
+// Lidar
+namespace Lidar {
+const std::string RP_PATH = "/dev/ttyUSB0";
+constexpr double MM_PER_M = 1000;
+constexpr uint32_t RPLIDAR_A1_BAUDRATE = 115200;
+constexpr uint32_t RPLIDAR_S1_BAUDRATE = 115200;
```
evan3334, line 54: We are using 256000 baud for the S1 (since I think the serial adapter is broken)
```suggestion
constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
```

src/lidar/read_rp_lidar.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include <atomic>
+#include <mutex>
+#include <thread>
+#include <iostream>
+
+using namespace navtypes;
+
+bool lidar_initialized = false;
+RPLidar rp_lidar(Constants::Lidar::RP_PATH, Constants::Lidar::RPLIDAR_A1_BAUDRATE);
```
evan3334, line 17: We are probably going to use the S1s over the A1s on the real rover - we got a mount made for them already and everything.
```suggestion
RPLidar rp_lidar(Constants::Lidar::RP_PATH, Constants::Lidar::RPLIDAR_S1_BAUDRATE);
```

evan3334 (approved):
Looks good to me - I had a couple more changes but they are small enough that I'll just apply them myself. Good work on this, I think it's ready to merge if @abhaybd approves

src/lidar/read_rp_lidar.h
```diff
@@ -7,10 +7,19 @@
 
 namespace lidar {
 
+/**
+ * @brief Startsup RPLidar with default settings
+ */
 bool initializeLidar();
```
abhaybd, line 13: This function prototype no longer matches the signature for the implementation of `initializeLidar` in `read_rp_lidar.cpp`. I'd recommend you change this line to `bool initializeLidar(double max_dist=16);` and change the implementation to be `bool initializeLidar(double max_dist) {...}`

src/lidar/read_rp_lidar.h
```diff
@@ -7,10 +7,19 @@
...
 
+/**
+ * @brief Provides RP Lidar data at current timeframe
+ */
 DataPoint<navtypes::points_t> readLidar();
 
+/**
+ * @brief Polar struct that holds radians, and theta
+ */
```
abhaybd, line 22: I think you mean radius instead of radians

src/lidar/read_rp_lidar.cpp
```diff
@@ -61,16 +50,15 @@ void readLidarLoop() {
 /**
  * @brief Startsup RPLidar with default settings
  */
-bool initializeLidar() {
+bool initializeLidar(double max_dist = 16) {
```
abhaybd, line 53: See my note on the function prototype in `read_rp_lidar.h`

src/lidar/read_rp_lidar.cpp
```diff
@@ -82,9 +70,8 @@ bool initializeLidar() {
 DataPoint<points_t> readLidar() {
     if (lidar_initialized) {
         DataPoint<points_t> pts;
-        points_lock.lock();
+        std::lock_guard<std::mutex> lk(points_lock);
         pts = last_points;
-        points_lock.unlock();
         return pts;
     }
     return {};
```
abhaybd, line 77: `lidar_initialized` is protected by `points_lock` so the `lock_guard` should be created *before* the if statement.

src/lidar/read_rp_lidar.cpp
```diff
@@ -61,16 +50,15 @@ void readLidarLoop() {
...
-            rp_lidar.setMaxDistance(16.0);
-            points_lock.lock();
+            rp_lidar.setMaxDistance(max_dist);
+            std::lock_guard<std::mutex> lk(points_lock);
             lidar_thread = std::thread(&readLidarLoop);
             lidar_initialized = true;
-            points_lock.unlock();
         }
     }
     return lidar_initialized;
```
abhaybd, line 64: `lidar_initialized` should be protected by the mutex, so create the `lock_guard` just before the `if(!lidar_initialized)` line. This is necessary to ensure that the initialization state is synchronized correctly.

src/lidar/read_rp_lidar.cpp
```diff
@@ -12,21 +13,14 @@
...
 DataPoint<points_t> last_points = {};
 std::mutex points_lock;
 
 namespace lidar {
 
-/**
- * @brief RP Lidar continuously scans environment, updates current timeframe with lidar data
- */
 void readLidarLoop() {
     using namespace std::chrono;
```
abhaybd, line 25: Is this still needed?

src/lidar/read_rp_lidar.h
abhaybd: Can you mention that the units of theta are radians though?

src/lidar/read_rp_lidar.cpp
```diff
@@ -37,22 +31,17 @@ void readLidarLoop() {
             double dtheta = (scan.value().angle_max-scan.value().angle_min)/(scan.value().ranges.size()-1);
             for (unsigned long i = 0; i < scan.value().ranges.size(); i++) {
                 double rad = dtheta*i;
-                double dist = scan.value().ranges[i] * MM_TO_M;
+                double dist_m = scan.value().ranges[i] * Constants::Lidar::MM_PER_M;
```
abhaybd, line 34: It looks like scan.value().ranges[i] is in meters, so no conversions necessary?

src/lidar/MainVis.cpp
```diff
@@ -70,8 +77,7 @@ int runRPLidar(unsigned long baudrate) {
 			double dtheta = (scan.value().angle_max-scan.value().angle_min)/(scan.value().ranges.size()-1);
 			for (long unsigned i = 0; i < scan.value().ranges.size(); i++) {
 				double rad = dtheta*i;
-				double MM_TO_M = 1000;
-				double dist = scan.value().ranges[i] * MM_TO_M;
+				double dist = scan.value().ranges[i] * Constants::Lidar::MM_PER_M;
```
abhaybd, line 80: It looks like `scan.value().ranges[i]` is in meters, so no conversions necessary?

abhaybd (changes requested):
Looking much better now! Just a few minor comments, mostly about thread safety.

src/lidar/read_rp_lidar.cpp
abhaybd: For some reason the line isn't highlighted so by "this" I mean the `using namespace std::chrono` line

abhaybd:
I just merged my PR which needs some changes in this PR. I've resolved the merge conflicts but there are some more changes that are necessary, such as using the new `robot` and `robot::types` namespaces, which you'll have to do where necessary in the code. The most prominent one I can think of is using `robot::types::DataPoint` instead of `DataPoint`, but there may be others.

abhaybd (approved):
Looks great! Great job getting this done

PR #159, opened 2022-05-27
Add class for computing differential wrist kinematics

evan3334:
This class should compute the kinematics of the differential joint on the arm - it has functions to compute gear positions from the joint position, and vice versa. Additionally, it can compute necessary gear power for a desired joint power, making sure everything stays normalized. 

Also added some small cleanup changes while I was at it.


src/Constants.h
```diff
@@ -10,6 +10,30 @@ constexpr size_t PACKET_PAYLOAD_SIZE = 8;
...
+};
+
+// So far only the base, shoulder, elbow have been tuned
+//
+// base, shoulder, elbow, forearm, diff_left, diff_right
+constexpr std::array<int32_t, 6> arm_Ps = {1000, 100, 500, 0, 0, 0};
+constexpr std::array<int32_t, 6> arm_Is = {50, 0, 50, 0, 0, 0};
+constexpr std::array<int32_t, 6> arm_Ds = {10000, 1000, 10000, 0, 0, 0};
+constexpr std::array<uint8_t, 6> arm_encoder_signs = {0, 0, 1, 0, 0, 0};
+
```
abhaybd, line 36: These look like physical constants, should these be in `real_world_constants.h` instead? The simulator probably wouldn't have the same values/constants.

src/world_interface/noop_world.cpp
```diff
@@ -8,7 +8,16 @@ using namespace robot::types;
...
+extern const WorldInterface WORLD_INTERFACE = WorldInterface::noop;
+
+const DiffDriveKinematics driveKinematics() {
+	//doesn't really matter what we set this to
+	return DiffDriveKinematics(1); 
+}
+
+const DiffWristKinematics wristKinematics() {
+	return DiffWristKinematics();
+}
```
abhaybd, line 20: Maybe these should return references to a statically allocated kinematics object? That would reduce a lot of unnecessary copying.

src/world_interface/world_interface.h
```diff
@@ -28,6 +30,9 @@ enum class WorldInterface { real, sim2d, sim3d, noop };
 /** @brief The current world interface being used. */
 extern const WorldInterface WORLD_INTERFACE;
 
+const DiffDriveKinematics driveKinematics();
+const DiffWristKinematics wristKinematics();
```
abhaybd, line 34: Is there any point in returning a const object? The object's still being copied everywhere, isn't it?

abhaybd (changes requested):
Overall pretty good! I just had some minor notes about style or efficiency.

src/Constants.h
evan3334: Yeah, not sure why I put those there - I think since I cherry-picked these commits from the protocol branch there's another commit that when composed with this one will move them to `real_world_constants.h`. I'll remove this.

abhaybd (approved):
Looks good to me!

PR #161, opened 2022-07-27
Implement basic mission control protocol

evan3334:
This PR (which is way overdue, sorry :disappointed:) introduces the base infrastructure required for the WebSocket-based communication protocol between the rover and Mission Control. Several things are still not implemented or supported, but the basic communication works and Mission Control is able to drive the chassis. Notable things that are not yet supported include:
- Position control of the differential wrist joints (we will need to write our own Rover-side PID control code for that since the position feedback is given for the joints, not the motors)
- Detection or reporting of the mounted peripheral (science station/arm/lidar); this is planned to be set via a command-line flag until we can get a robust solution for actually detecting what is mounted
- Reporting of the current autonomous plan and localization confidence via the mission control protocol (although this is currently reported through the old PlanViz system, which is deprecated and needs to be removed)
- Control of the science station


camera-config/MastCameraCalibration.yml
```diff
@@ -3,7 +3,7 @@
 name: "mast"
 description: "Camera located on the mast of the rover."
 # Make sure that this camera_id matches the one defined in the udev rules.
-camera_id: 20
+camera_id: 0
```
evan3334, line 6: Clarifying this, we set the camera ID to 20 since we created udev rules that would ensure this specific camera (identified by its USB vendor/device IDs) was always mounted at `/dev/video20` as well as whatever camera ID it was automatically assigned by the kernel. This means even with multiple cameras that might have different auto-assigned IDs each time, we could guarantee they all had stable IDs that wouldn't conflict with any others.
If the udev rules aren't modifed, we should probably revert this change. Thoughts?

src/Constants.h
```diff
@@ -1,25 +1,80 @@
...
 #include "world_interface/data.h"
 
 #include <cmath>
 #include <string>
 
 namespace Constants {
+/**
+   @deprecated No need for this constant once we fully switch to the new network protocol
+ */
 constexpr size_t PACKET_PAYLOAD_SIZE = 8;
```
evan3334, line 13: Whoops, forgot to remove this.
```suggestion
```

camera-config/MastCameraCalibration.yml
evan3334: Udev rules are at [rover-config/50-rover-cameras.rules](https://github.com/huskyroboticsteam/Resurgence/blob/master/rover-config/50-rover-cameras.rules), and honestly should probably be updated to reflect the hardware we are using this year.

abhaybd: Oh I wasn't aware of the udev rules, that's a good solution. The camera wasn't getting picked up on 20 so I changed the value to 0, but if we can update the udev rules then I agree this change should be reverted.

For now, until we can change those rules and test it, let's leave as is and add a to-do?

abhaybd (approved):
Wasn't really able to review this thing in its entirety, but from a short-ish skim it looks good enough. I know this was a really big task, so thanks for getting that done!

PR #162, opened 2022-08-02
Tuned arm scales, fixed differential bug

abhaybd:
Small PR that tunes all but one of the arm PWM scales, and fixes a bug with the differential. Now, the arm works with teleop as expected, except for the arm base which is untested due to not having a motor right now.

### DIFF
--- a/src/world_interface/kinematic_common_interface.cpp
+++ b/src/world_interface/kinematic_common_interface.cpp
@@ -180,14 +180,14 @@ void setJointMotorPower(robot::types::jointid_t joint, double power) {
 		float pitchPwr = power;
 		float rollPwr = getJointPowerValue(jointid_t::differentialRoll);
 		gearpos_t gearPwr = wristKinematics().jointPowerToGearPower(jointpos_t{pitchPwr, rollPwr});
-		setMotorPos(motorid_t::differentialLeft, gearPwr.left);
-		setMotorPos(motorid_t::differentialRight, gearPwr.right);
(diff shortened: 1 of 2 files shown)


src/world_interface/kinematic_common_interface.cpp
```diff
@@ -180,14 +180,14 @@ void setJointMotorPower(robot::types::jointid_t joint, double power) {
...
+		setMotorPower(motorid_t::differentialLeft, gearPwr.left);
+		setMotorPower(motorid_t::differentialRight, gearPwr.right);
 	} else if (joint == jointid_t::differentialRoll) {
 		float pitchPwr = getJointPowerValue(jointid_t::differentialPitch);
 		float rollPwr = power;
 		gearpos_t gearPwr = wristKinematics().jointPowerToGearPower(jointpos_t{pitchPwr, rollPwr});
-		setMotorPos(motorid_t::differentialLeft, gearPwr.left);
-		setMotorPos(motorid_t::differentialRight, gearPwr.right);
+		setMotorPower(motorid_t::differentialLeft, gearPwr.left);
+		setMotorPower(motorid_t::differentialRight, gearPwr.right);
```
evan3334, line 190: Good catch - totally missed this!

evan3334 (approved):
Looks good to me!

src/world_interface/kinematic_common_interface.cpp
abhaybd: Caused a little confusion when the differential wasn't moving but thankfully wasn't that bad to track down lol

PR #163, opened 2022-08-04
Add a thread to repeat setJointPower to keep motors alive

evan3334:
This PR adds a method that immediately calls `robot::setJointPower` but also stores requested joint power values in a map which are periodically read by a thread and sent to `robot::setJointPower` again, to prevent the motor boards' watchdog timers from killing the motors when the requested power does not change. The period is 1/3 of a second right now, but we can change that.

This is a draft PR for now until I can test the code (not sure if there is a watchdog timer on the simulator, so I will probably just put a debug print in setJointPower to make sure it's being called periodically), but I wanted to get the review process started.

### DIFF
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -208,13 +208,15 @@ void MissionControlProtocol::handleConnection() {
 	this->_server.sendJSON(Constants::MC_PROTOCOL_NAME, j);
 
 	// start power repeat thread (if not already running)
-	{
-		std::unique_lock<std::mutex> power_repeat_lock(_joint_repeat_mutex);
-		if (!this->_joint_repeat_running) {
-			this->_joint_repeat_running = true;
-			this->_joint_repeat_thread =
-				std::thread(&MissionControlProtocol::jointPowerRepeatTask, this);
-		}
+	this->startPowerRepeat();
+}
+
+void MissionControlProtocol::startPowerRepeat() {
+	std::unique_lock<std::mutex> power_repeat_lock(_joint_repeat_mutex);
+	if (!this->_joint_repeat_running) {
+		this->_joint_repeat_running = true;
+		this->_joint_repeat_thread =
+			std::thread(&MissionControlProtocol::jointPowerRepeatTask, this);
 	}
 }
 
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -86,6 +86,7 @@ static bool validateEmergencyStopRequest(const json& j) {
 static void handleEmergencyStopRequest(const json& j) {
 	bool stop = j["stop"];
 	if (stop) {
+		// TODO: shutdown joint power repeat
 		robot::setCmdVel(0, 0);
 		stopAllJoints();
 		// TODO: do e-stop properly. Add emergencyStop() method to world interface,
@@ -103,6 +104,7 @@ static bool validateOperationModeRequest(const json& j) {
 
 static void handleOperationModeRequest(const json& j) {
 	std::string mode = j["mode"];
+	// TODO: shutdown joint power repeat if we go to autonomous mode
 	Globals::AUTONOMOUS = (mode == "autonomous");
 }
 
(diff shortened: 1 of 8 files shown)


src/network/MissionControlProtocol.h
```diff
@@ -29,15 +30,22 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
 
 private:
 	void videoStreamTask();
+	void jointPowerRepeatTask();
 	SingleClientWSServer& _server;
 	std::shared_mutex _stream_mutex;
+	std::shared_mutex _joint_power_mutex;
```
abhaybd, line 36: If there's only one reader and possibly many writers, is there any point in making this a shared mutex? There aren't any concurrent readers to benefit from anyway, so imo a regular mutex would do the same job just fine, with less overhead.

src/network/MissionControlProtocol.cpp
```diff
@@ -216,17 +220,44 @@ MissionControlProtocol::MissionControlProtocol(SingleClientWSServer& server)
...
-	// TODO: add new thread to handle resending motor power values
+	this->_joint_repeat_running = true;
+	this->_joint_repeat_thread =
+		std::thread(&MissionControlProtocol::jointPowerRepeatTask, this);
 	// TODO: add new thread to periodically send motor status reports
 }
 
 MissionControlProtocol::~MissionControlProtocol() {
+	this->_joint_repeat_running = false;
+	this->_joint_repeat_thread.detach();
```
abhaybd, line 231: In the destructor we should probably do more cleanup. For example, waiting for the joint thread to stop and then sending 0 power to all motors. We probably don't want to leave the motors running and rely on the watchdog to turn them off.

On that note, we should also definitely have a disconnection handler that shuts off all the motors. It would be pretty bad for MC to disconnect and leave all the motors running. Maybe the disconnection handler can be called from the destructor, that would fix both things.

abhaybd (changes requested):
Overall looks good, just one minor nitpick and also some features that I think are missing.

abhaybd:
Also, there seems to be a pretty bad bug that I think we totally missed. Basically, for some reason I didn't consider the use-case of deriving from `WebSocketProtocol`. So, when the `MissionControlProtocol` object is registered with the server, `WebSocketProtocol::operator=` is being called, which obviously misses all the fields in `MissionControlProtocol`.

This means that all the fields of `MissionControlProtocol` are not copied over, and when the original instance goes out of scope they will all be destroyed, which breaks everything. By pure chance since `mcProto` is instantiated in `main()` it's been working so far since it never goes out of scope, but that's definitely not something we should rely on.

Basically, the fix here is to a) make `WebSocketProtocol` OOP-friendly (virtual destructor, etc) and b) make `WebSocketServer` accept a `unique_ptr<WebSocketProtocol>` instead of `const WebSocketProcotol &`, which avoids all issues with polymorphic assignment. This way, ownership of resources gets explicitly handled and we won't run into any weird issuse.

I'll get that done in a PR ASAP, and then you should be able to rebase on top of those changes.

src/network/MissionControlProtocol.h
evan3334: Yeah, you're probably right, I think I was just copying the stuff I already wrote for the stream mutex without much thought lol

src/network/MissionControlProtocol.cpp
evan3334: Yeah, that's a good idea. I'll get that implemented

evan3334:
Updated this with a couple other features:
* power repeater now includes drive velocity
* power repeater thread waits on a condition variable instead of just a timer; allows waking up immediately in case we need to stop it
* power repeater thread is stopped when we enter autonomous or e-stop, and is restarted when we enter tele-op or disengage e-stop.
* power repeater thread does not start until mission control connects, and stops whenever mission control disconnects (also shutting down all motors). This behavior only occurs in tele-op mode, to prevent unnecessarily stopping the rover while it is under autonomous operation.

Notable things that still need to be done:
* we should ignore all motor-related requests when in autonomous mode, since we are not permitted to tele-operate.

abhaybd (approved):
Looks good! I tested it on the sim and it seems to all work well. Since we don't have a lot of time, I think we should merge this PR now since it most likely will work fine on the real rover, and we can always revert if necessary. Thoughts? If that makes sense feel free to convert to a real PR and merge.

abhaybd (changes requested):
I just found a bug in the disconnection handler. When MC disconnects, it hangs for some reason when testing in the sim, which doesn't allow MC to reconnect again.

I suspect some deadlock is happening somewhere? We should iron this out before merging.

evan3334:
Good catch, I totally forgot that having the locks in a different order is a really easy way to cause deadlock 😬

abhaybd (approved):
Looks good to me! Hardware testing should show if any issues remain, but I think we're probably fine.

PR #164, opened 2022-08-05
Bugfixes for websockets

abhaybd:
Changed WebSocketProtocol to be OOP friendly, use unique_ptr in WebSocketServer to avoid assignment.

See [this comment](https://github.com/huskyroboticsteam/Resurgence/pull/163#issuecomment-1205682474) for an explanation of the bug that is fixed in this PR.


evan3334 (approved):
Looks good to me, although admittedly I don't know a whole lot about `std::move` and `unique_ptr`s. I read through the linked comment though, so it seems to make sense.

Good catch!

PR #165, opened 2022-08-07
Fix steering inversion bug

abhaybd:
This PR fixes a bug where the steering would reverse direction if the steer value became too large. Looking at the code, it seems that there was a misplaced negative sign, probably a typo.


evan3334 (approved):
Whoops, no idea why I did that, haha. Good catch!

PR #170, opened 2022-10-20
Follow style guide in StyleGuide

huttongrabiel:


abhaybd (approved):

abhaybd:
Good catch! Ironic that the style guide doesn't follow itself lol

PR #173, opened 2022-10-25
Add peripheral console arg and updated README

DrDab:


abhaybd:
Overall looks good! Before I accept, any reason it shows whitespace-only changes to `easy-install.sh`? Don't see anything different, did line endings get changed or something?

If that wasn't on purpose, can you revert your changes to that file so this pr doesn't touch it?

PR #175, opened 2022-10-28
Made easy_install.sh executable and update README accordingly

DrDab:


abhaybd (approved):
LGTM! Nice catch on the file permissions

PR #177, opened 2022-10-29
Fixed undefined Eigen casting behavior

DrDab:


abhaybd (approved):
Good job tracking this down!

PR #178, opened 2022-10-29
Log level console arg

DrDab:
Made log level specifiable using command line argument.

### DIFF
--- a/src/Globals.cpp
+++ b/src/Globals.cpp
@@ -1,6 +1,6 @@
 #include "Globals.h"
-
 #include "Constants.h"
+#include "log.h"
 
 #include <vector>
 #include <atomic>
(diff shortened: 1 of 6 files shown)


src/Globals.cpp
```diff
@@ -1,6 +1,6 @@
 #include "Globals.h"
-
 #include "Constants.h"
+#include "log.h"
```
abhaybd, line 3: Is this include necessary?

src/log.cpp
```diff
@@ -1,17 +1,21 @@
-
+#include "Globals.h"
```
abhaybd, line 1: Do we need this include either?

src/log.h
```diff
@@ -3,12 +3,11 @@
...
-extern int LOG_LEVEL;
-
 constexpr int LOG_TRACE = 0;
 constexpr int LOG_DEBUG = 1;
 constexpr int LOG_INFO = 2;
 constexpr int LOG_WARN = 3;
 constexpr int LOG_ERROR = 4;
 
 void log(int level, const char* fmt, ...);
+void setLogLevel(int level);
```
abhaybd, line 13: I know it wasn't there before, but could we add doxygen documentation to these methods?

abhaybd (changes requested):
Overall well done, just some minor nitpicks

DrDab:
Added documentation for log function and removed the unnecessary includes.

abhaybd (approved):
Nice clean code, LGTM!

PR #179, opened 2022-10-29
Create general-purpose controllers

abhaybd:
This PR adds several general-purpose controllers for use in controlling different mechanisms with position control.
- `JacobianController` follows a motion profile using the kinematics jacobian to follow the path as well as possible
- `JacobianVelController` tries to achieve a mechanism velocity using the kinematics jacobian
- `TrapezoidalVelocityProfile` creates a velocity profile that has differentiable position and continuous velocity
- `PIDController` creates, well, a pid controller

### DIFF
--- a/src/control/JacobianController.h
+++ b/src/control/JacobianController.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "../filters/StateSpaceUtil.h"
 #include "../world_interface/data.h"
 #include "TrapezoidalVelocityProfile.h"
 
@@ -24,10 +25,13 @@ public:
 		const std::array<double, outputDim>& maxVels,
 		const std::array<double, outputDim>& maxAccels)
 		: kinematicsFunc(kinematicsFunc),
-		  jacobianFunc(jacobianFunc ? jacobianFunc : jacobianFunc),
+		  jacobianFunc(jacobianFunc
+						   ? jacobianFunc
+						   : std::bind(filters::statespace::numericalJacobian, kinematicsFunc,
+									   std::placeholders::_1, outputDim)),
 		  velocityProfile(maxVels, maxAccels) {
-		// TODO: numerically approximate with statespace::numericalJacobian
-		assert(kinematicsFunc && jacobianFunc);
+		assert(this->kinematicsFunc);
+		assert(this->jacobianFunc);
 	}
 
 	Vectord<inputDim> getCommand(double elapsedTime, const Vectord<inputDim>& currPos) const {
--- a/src/control/JacobianController.h
+++ b/src/control/JacobianController.h
@@ -12,12 +12,31 @@
 
 namespace control {
 
+/**
+ * @brief This class controls the position of a multidimensional mechanism
+ * with a trapezoidal velocity profile using the jacobian of the kinematics.
+ *
+ * @tparam outputDim The dimension of the output vector of the kinematic function.
+ * @tparam inputDim The dimension of the input vector of the kinematic function.
+ * @see https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant
+ */
 template <int outputDim, int inputDim> class JacobianController {
 private:
 	template <int dim> using Vectord = filters::statespace::Vectord<dim>;
 	template <int rows, int cols> using Matrixd = filters::statespace::Matrixd<rows, cols>;
 
 public:
+	/**
+	 * @brief Construct a new controller.
+	 *
+	 * @param kinematicsFunc The kinematics of the mechanism being controlled. This is
+	 * a function that takes in some input vector (ex: joint angles of an arm) and outputs
+	 * another vector (ex: the 3d pose of the hand)
+	 * @param jacobianFunc The jacobian of the kinematic function.
+	 * When executing a command, getCommand() should be called at this rate.
+	 * @param maxVels The maximum velocity in the output space of the kinematic function.
+	 * @param maxAccels The maximum acceleration in the output space of the kinematic function.
+	 */
 	JacobianController(
 		const std::function<Vectord<outputDim>(const Vectord<inputDim>&)>& kinematicsFunc,
 		const std::function<Matrixd<outputDim, inputDim>(const Vectord<inputDim>&)>&
@@ -34,11 +53,29 @@ public:
 		assert(this->jacobianFunc);
 	}
 
+	/**
+	 * @brief Get the current command.
+	 *
+	 * @param elapsedTime The time since the target was commanded.
+	 * @param currPos The current position, in the input space.
+	 * @return Vectord<inputDim> The target position, in the input space of the kinematic function.
+	 * Returns the current position if no target is set.
+	 */
 	Vectord<inputDim> getCommand(double elapsedTime, const Vectord<inputDim>& currPos) const {
 		double unused;
 		return getCommand(elapsedTime, currPos, unused);
 	}
 
+	/**
+	 * @brief Get the current command.
+	 *
+	 * @param elapsedTime The time since the target was commanded.
+	 * @param currPos The current position, in the input space.
+	 * @param[out] cosineSim Output parameter, gives the cosine similarity of the delta to the returned target and the delta to the commanded target.
+	 * This similarity may be low if the kinematics of the mechanism do not allow it to move in the commanded direction.
+	 * @return Vectord<inputDim> The target position, in the input space of the kinematic function.
+	 * Returns the current position if no target is set.
+	 */
 	Vectord<inputDim> getCommand(double elapsedTime, const Vectord<inputDim>& currPos,
 								 double& cosineSim) const {
 		if (!velocityProfile.hasTarget()) {
@@ -58,15 +95,29 @@ public:
 		return currPos + inputPosDiff;
 	}
 
+	/**
+	 * @brief Set the target position of the controller.
+	 *
+	 * @param currPos The current position in the input space.
+	 * @param target The target position in the output space.
+	 */
 	void setTarget(const Vectord<inputDim>& currPos, const Vectord<outputDim>& target) {
 		Vectord<outputDim> currPosOutput = kinematicsFunc(currPos);
 		velocityProfile.setTarget(currPosOutput, target);
 	}
 
+	/**
+	 * @brief Check if a target has been set.
+	 *
+	 * @return bool True iff a target has been set.
+	 */
 	bool hasTarget() {
 		return velocityProfile.hasTarget();
 	}
 
+	/**
+	 * @brief Reset the controller, unsetting the target.
+	 */
 	void reset() {
 		velocityProfile.reset();
 	}
--- a/src/control/JacobianController.h
+++ b/src/control/JacobianController.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "../filters/StateSpaceUtil.h"
+#include "../navtypes.h"
 #include "../world_interface/data.h"
 #include "TrapezoidalVelocityProfile.h"
 
@@ -21,10 +22,6 @@ namespace control {
  * @see https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant
  */
 template <int outputDim, int inputDim> class JacobianController {
-private:
-	template <int dim> using Vectord = filters::statespace::Vectord<dim>;
-	template <int rows, int cols> using Matrixd = filters::statespace::Matrixd<rows, cols>;
-
 public:
 	/**
 	 * @brief Construct a new controller.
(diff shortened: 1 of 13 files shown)


src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @brief This class controls the position of a multidimensional mechanism
+ * with a trapezoidal velocity profile using the jacobian of the kinematics.
+ *
+ * @tparam outputDim The dimension of the output vector of the kinematic function.
+ * @tparam inputDim The dimension of the input vector of the kinematic function.
+ * @see https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant
+ */
+template <int outputDim, int inputDim> class JacobianController {
```
simranmalhi, line 24: Not sure how in depth you want to go with the math, but I think it would be helpful to explicitly state that the dimensions of the jacobian function is: kinematicOutputDim x kinematicInputDim in the description.

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+public:
+	/**
+	 * @brief Construct a new controller.
+	 *
+	 * @param kinematicsFunc The kinematics of the mechanism being controlled. This is
+	 * a function that takes in some input vector (ex: joint angles of an arm) and outputs
+	 * another vector (ex: the 3d pose of the hand)
+	 * @param jacobianFunc The jacobian of the kinematic function.
+	 * When executing a command, getCommand() should be called at this rate.
```
simranmalhi, line 33: I don't think line 33 should be included. I'm also not sure of what it means to be called "at this rate".

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+	 * @param maxVels The maximum velocity in the output space of the kinematic function.
+	 * @param maxAccels The maximum acceleration in the output space of the kinematic function.
+	 */
+	JacobianController(
+		const std::function<navtypes::Vectord<outputDim>(const navtypes::Vectord<inputDim>&)>&
+			kinematicsFunc,
+		const std::function<navtypes::Matrixd<outputDim, inputDim>(
+			const navtypes::Vectord<inputDim>&)>& jacobianFunc,
+		double maxVel, double maxAccel)
```
simranmalhi, line 42: This is completely a style preference but I think the indentation and length of the parameters could be more consistent.

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+		: kinematicsFunc(kinematicsFunc),
+		  jacobianFunc(jacobianFunc
+						   ? jacobianFunc
```
simranmalhi, line 45: Indentation is a little large here.

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+						   : std::bind(filters::statespace::numericalJacobian, kinematicsFunc,
+									   std::placeholders::_1, outputDim)),
+		  velocityProfile(maxVel, maxAccel) {
+		assert(this->kinematicsFunc);
+		assert(this->jacobianFunc);
+	}
+
+	/**
+	 * @brief Get the current command.
```
simranmalhi, line 54: Current command here seems a little vague and repetitive of the method name. Maybe saying "Get the target position" would be more helpful?

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+	 *
+	 * @param currTime The current timestamp.
+	 * @param currPos The current position, in the input space.
+	 * @return navtypes::Vectord<inputDim> The target position, in the input space of the
+	 * kinematic function. Returns the current position if no target is set.
+	 */
+	navtypes::Vectord<inputDim> getCommand(robot::types::datatime_t currTime,
+										   const navtypes::Vectord<inputDim>& currPos) const {
```
simranmalhi, line 62: Again, indentation seems to be a little too large here (sorry I'm a stickler for this haha)

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+
+	/**
+	 * @brief Get the current command.
+	 *
+	 * @param currTime The current timestamp.
+	 * @param currPos The current position, in the input space.
+	 * @param[out] cosineSim Output parameter, gives the cosine similarity of the delta to the
+	 * returned target and the delta to the commanded target. This similarity may be low if the
+	 * kinematics of the mechanism do not allow it to move in the commanded direction. May be
+	 * NaN if stuck in singularity point, in which case the command is the current position.
```
simranmalhi, line 75: I think saying that the command is the current position in line 75 should just be in the @return below for consistency.

src/control/JacobianController.h
```diff
@@ -0,0 +1,139 @@
...
+	 * kinematic function. Returns the current position if no target is set.
+	 */
+	navtypes::Vectord<inputDim> getCommand(robot::types::datatime_t currTime,
+										   const navtypes::Vectord<inputDim>& currPos,
+										   double& cosineSim) const {
+		if (!velocityProfile.hasTarget()) {
+			return currPos;
+		}
+
+		navtypes::Vectord<outputDim> currTarget = velocityProfile.getCommand(currTime);
```
simranmalhi, line 86: I think it would be good to add comments explaining what is happening at a high level.

src/control/JacobianController.h
simranmalhi: Or specify that the command is for the target position.

src/control/JacobianController.h
simranmalhi, line 24: Since you called the other class JacobianVelController maybe call this one JacobianPosController to be more consistent

src/control/JacobianVelController.h
```diff
@@ -0,0 +1,138 @@
...
+public:
+	/**
+	 * @brief Construct a new controller.
+	 *
+	 * @param kinematicsFunc The kinematics of the mechanism being controlled. This is
+	 * a function that takes in some input vector (ex: joint angles of an arm) and outputs
+	 * another vector (ex: the 3d pose of the hand)
+	 * @param jacobianFunc The jacobian of the kinematic function.
+	 * @param dt The time delta of this controller.
+	 * When executing a command, getCommand() should be called at this rate.
```
simranmalhi, line 30: Don't think line 30 needs to be included.

src/control/JacobianVelController.h
```diff
@@ -0,0 +1,138 @@
...
+	 */
+	JacobianVelController(
+		const std::function<navtypes::Vectord<outputDim>(const navtypes::Vectord<inputDim>&)>&
+			kinematicsFunc,
+		const std::function<navtypes::Matrixd<outputDim, inputDim>(
+			const navtypes::Vectord<inputDim>&)>& jacobianFunc)
+		: kinematicsFunc(kinematicsFunc),
+		  jacobianFunc(jacobianFunc
+						   ? jacobianFunc
```
simranmalhi, line 39: I think the indentation should be more consistent.

src/control/JacobianController.h
simranmalhi: You say "@brief Get the current target position." in the method descriptions of the getCommand() in the JacobianVelController so you can stick with that here.

src/control/JacobianController.h
simranmalhi, line 24: Could call this class "JacobianPosController" to be consistent with "JacobianVelController"

src/control/JacobianVelController.h
```diff
@@ -0,0 +1,138 @@
...
+	 * @param[out] cosineSim Output parameter, gives the cosine similarity of the delta to the
+	 * returned target and the delta to the commanded target. This similarity may be low if the
+	 * kinematics of the mechanism do not allow it to move in the commanded direction.
+	 * @return navtypes::Vectord<inputDim> The target position of the mechanism, in the
+	 * kinematic input space. Returns the current position if no target is set.
+	 */
+	navtypes::Vectord<inputDim> getCommand(robot::types::datatime_t currTime,
+										   const navtypes::Vectord<inputDim>& currPos,
+										   double& cosineSim) {
+		if (!controllerState) {
```
simranmalhi, line 74: Could also add some comments here explaining what is happening at a high level.

src/control/JacobianVelController.h
```diff
@@ -0,0 +1,138 @@
...
+	 * @param targetVel The target velocity, in the output space of the kinematic function.
+	 */
+	void setTarget(robot::types::datatime_t currTime,
+				   const navtypes::Vectord<outputDim>& targetVel) {
+		state_t state = {.timestamp = currTime, .targetVel = targetVel, .lastTarget = {}};
+		this->controllerState = state;
+	}
+
+	/**
+	 * @brief Check if a target has been set.
```
simranmalhi, line 111: You could specify here whether the target is for position or velocity.

src/control/PIDControl.h
```diff
@@ -0,0 +1,131 @@
...
+	 * @brief Set the target of the controller.
+	 *
+	 * @param setPoint An array of setpoints, one for each dimension.
+	 */
+	void setTarget(const Arrayd<dim>& setPoint) {
+		this->setPoint = setPoint;
+	}
+
+	/**
+	 * @brief Get the output from the controller.
```
simranmalhi, line 78: Instead of saying output from the controller, you can say "Get the dimension targets from the controller" so it's less repetitive with the method name.

src/control/PIDControl.h
```diff
@@ -0,0 +1,131 @@
...
+	 * @param currTime The current timestamp.
+	 * @param currValue The current values of the controlled signals.
+	 * @return Arrayd<dim> Targets for each dimension if target is set, else zeros.
+	 */
+	Arrayd<dim> getOutput(robot::types::datatime_t currTime, const Arrayd<dim>& currValue) {
+		if (!setPoint.has_value()) {
+			return Arrayd<dim>::Zero();
+		}
+
+		Arrayd<dim> err = setPoint.value() - currValue;
```
simranmalhi, line 89: Adding high level comments here would be helpful if someone just wants to casually read the code.

src/control/TrapezoidalVelocityProfile.cpp
```diff
@@ -0,0 +1,35 @@
...
+	Vectord<1> startVec{startPos};
+	Vectord<1> endVec{endPos};
+	profile.setTarget(currTime, startVec, endVec);
+}
+
+bool SingleDimTrapezoidalVelocityProfile::hasTarget() const {
+	return profile.hasTarget();
+}
+
+double
```
simranmalhi, line 26: Why is double in a line by itself? Or is this another git issue from my end?

src/control/TrapezoidalVelocityProfile.h
```diff
@@ -0,0 +1,205 @@
...
+
+#include <Eigen/Core>
+
+namespace control {
+
+/**
+ * @brief Trapezoidal velocity profile in multiple dimensions.
+ * Motion is coordinated so that all dimensions arrive at the goal at the same time.
+ */
+template <int dim> class TrapezoidalVelocityProfile {
```
simranmalhi, line 20: Need to add a @tparam for the dimensions.

src/control/TrapezoidalVelocityProfile.h
```diff
@@ -0,0 +1,205 @@
...
+
+	/**
+	 * @brief Calculate a profile, assuming the mechanism begins and ends at rest.
+	 *
+	 * @param currTime The current timestamp.
+	 * @param startPos The position the profile starts at.
+	 * @param endPos The position the profile ends at.
+	 */
+	void setTarget(robot::types::datatime_t currTime, const navtypes::Vectord<dim>& startPos,
+				   const navtypes::Vectord<dim>& endPos) {
```
simranmalhi, line 72: high level comments here would be helpful

src/navtypes.h
```diff
@@ -19,6 +19,9 @@ struct gpscoords_t {
 using obstacle_t = Eigen::ArrayX2d;
 using obstacles_t = std::vector<obstacle_t>;
 
+template <int d> using Vectord = Eigen::Matrix<double, d, 1>;
+template <int r, int c> using Matrixd = Eigen::Matrix<double, r, c>;
```
simranmalhi, line 23: adding a comment explicitly stating what r and c is would be good

src/control/TrapezoidalVelocityProfile.cpp
abhaybd: The rest of the function signature is so long, and since there's only 1 parameter it can't split those, so the clang-format decided that this was the best way to limit the line length. I'd say it's pretty clear and since it follows the clang-format rules, it's probably good enough?

simranmalhi (approved):

PR #182, opened 2022-11-08
Add logging colors

DrDab:


src/log.cpp
```diff
@@ -4,17 +4,57 @@
...
+			std::cout << prefixStarter;
+		}
+
+		va_list arg;
 		va_start(arg, fmt);
 		vprintf(fmt, arg);
 		va_end(arg);
+
+		if (useColors) {
+			std::cout << "\033[0m";
```
abhaybd, line 47: Can you just add a brief comment saying what this ANSI code is?

abhaybd (changes requested):
Overall pretty good! Just a minor nitpick about magic constants

src/log.cpp
DrDab: added!

abhaybd (approved):
LGTM, nice job!

PR #183, opened 2022-11-11
Change for loop to while loop + clang-format

huttongrabiel:
Fixes for loop that looped infinitely and did not use its iteration variable.
Also, run clang-format over the file.


abhaybd (approved):
LGTM! It's always good to go back and clean up code for style conformance. Make sure to update this to be up-to-date with master before merging though.

PR #186, opened 2022-12-03
Added swerve drive IK/FK, tests and docs

DrDab:


tests/kinematics/SwerveDriveKinematicsTest.cpp
```diff
@@ -0,0 +1,66 @@
...
+	double wv = sqrt(2) * PI / 4;
+	double wr = PI / 4;
+	assertApprox(
+		{0, 0, PI / 2},
+		kinematics.getPoseUpdate(wheelVel(wv, wr * 3, wv, wr, wv, wr * 5, wv, wr * 7), 0, 1));
+
+	// drive forward 4 m, left 5 m, rotate 0.42 * PI radians CCW
+	swervewheelvel_t targetWheelVels = kinematics.robotVelToWheelVel(4, 5, 0.42 * PI);
+	assertApprox({4, 5, 0.42 * PI}, kinematics.getPoseUpdate(targetWheelVels, 0, 1));
+}
```
abhaybd, line 66: I like these test cases, can we also have some `getPoseUpdate` test cases where the heading is nonzero and the dt is not 1?

src/kinematics/SwerveDriveKinematics.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include "../navtypes.h"
+
+using namespace navtypes;
+
+struct swervewheelvel_t {
+	/**
+	 * Represents robot wheel velocities in robot reference frame,
+	 * in polar form (wheel speed and angle)
+	 */
```
abhaybd, line 11: Can you put this comment before the struct (not inside it) so it renders in doxygen correctly?

src/kinematics/SwerveDriveKinematics.cpp
```diff
@@ -0,0 +1,88 @@
...
+	// rby}
+
+	return {hypot(wheelVels(1), wheelVels(0)), 
+            atan2(wheelVels(1), wheelVels(0)),
+			hypot(wheelVels(3), wheelVels(2)), 
+            atan2(wheelVels(3), wheelVels(2)),
+			hypot(wheelVels(5), wheelVels(4)), 
+            atan2(wheelVels(5), wheelVels(4)),
+			hypot(wheelVels(7), wheelVels(6)), 
+            atan2(wheelVels(7), wheelVels(6))};
```
abhaybd, line 71: For cleanliness, can we use `std::hypot` and `std::atan2` instead of the global versions?

src/kinematics/SwerveDriveKinematics.h
```diff
@@ -0,0 +1,113 @@
...
+	double lfRot;
+	double rfVel;
+	double rfRot;
+	double lbVel;
+	double lbRot;
+	double rbVel;
+	double rbRot;
+};
+
+class SwerveDriveKinematics {
```
abhaybd, line 23: We should have a class-level comment too, can you just put a brief description of the class?

abhaybd (changes requested):
In general, great job! This is pretty high quality code, and I like the documentation you've added. One main nitpick: Wherever possible we should use compile-time matrix dimensions. Since the dimensions of swerve bases are known in advance, can you replace every `Eigen::MatrixXd` usage with the compile-time version of `Eigen::Matrix<double, ..., ...>` instead? Aside from that, there were just a few minor changes we should make.

PR #190, opened 2023-01-14
DriveToWaypointCommand tests

DrDab:
Added tests and documentation for DriveToWaypointCommand


src/commands/DriveToWaypointCommand.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace commands {
+
+DriveToWaypointCommand::DriveToWaypointCommand(const point_t& target, double thetaKP,
+											   double driveVel, double slowDriveVel,
+											   double doneThresh)
+	: target(target), pose(pose_t::Zero()), thetaKP(thetaKP), driveVel(driveVel),
+	  slowDriveVel(slowDriveVel), doneThresh(doneThresh), setStateCalledBeforeOutput(false) {}
+
+void DriveToWaypointCommand::setState(const navtypes::pose_t& pose) {
```
abhaybd, line 18: To make testing and usage easier, we probably shouldn't call the clock for the current time, let's take it as an argument here and store it as a field.

src/commands/DriveToWaypointCommand.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return {.thetaVel = thetaVel, .xVel = xVel};
+}
+
+bool DriveToWaypointCommand::isDone() {
+	if ((pose.topRows<2>() - target.topRows<2>()).norm() <= doneThresh) {
+		if (!closeToTargetStartTime.has_value()) {
+			datatime_t time = dataclock::now();
+			closeToTargetStartTime = time;
+		}
+		return dataclock::now() - closeToTargetStartTime.value() >= CLOSE_TO_TARGET_DUR;
```
abhaybd, line 49: Same as above, we should probably use the time passed in setState() instead of polling the clock to make testing easier.

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+  cmd.setState(aparallelY);
+  assertApprox(THETA_KP * M_PI, NORMAL_DRIVE_SPEED, cmd.getOutput());
+
+  // test robot uses slow drive speed if within done threshold
+  // (in this case, done threshold is 25 cm)
+  navtypes::pose_t close_enough(3.0 - (3.0 / (5.0 * 16.0)),
+                                4.0 - (3.0 / (5.0 * 12.0)), 0.0);
+  cmd.setState(close_enough);
+  assertApprox(THETA_KP * ATAN_4_3, SLOW_DRIVE_SPEED, cmd.getOutput());
+}
```
abhaybd, line 75: Can we add a test case testing isDone() functionality? Changing setState() to accept the current time should help with this.

abhaybd (changes requested):
Minor changes needed, I think I wasn't quite right when I mentioned what the signature for setState() should be

src/commands/DriveToWaypointCommand.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <chrono>
+#include <optional>
+
+namespace commands {
+
+/**
+ * Provides robot drive velocity commands to navigate the robot to a given waypoint given
+ * the velocity parameters to drive the robot at, the proportional factor to steer the robot with
+ * and the 2D Cartesian position of the target in the world reference frame.
```
abhaybd, line 13: Good documentation!

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include "../../src/commands/CommandBase.h"
+#include "../../src/commands/DriveToWaypointCommand.h"
+
+using namespace robot::types;
+using namespace std::chrono_literals;
+
+void assertApprox(double expectedThetaVel, double expectedXVel,
+                  const commands::command_t& actual, double angle = 1e-5,
+                  double vel = 1e-5) {
```
abhaybd, line 13: Maybe clarify that `angle` and `vel` are error tolerances, not the actual angle and vel.

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  if (velEqual && angleEqual) {
+    SUCCEED();
+  } else {
+    printf(
+        "Expected (thetaVel,xVel)=(%.5f,%.5f), "
+        "Actual (thetaVel,xVel)=(%.5f,%.5f)\n",
+        expectedThetaVel, expectedXVel, actual.thetaVel, actual.xVel);
+    FAIL();
+  }
```
abhaybd, line 25: This could all be replaced with:
```
REQUIRE_THAT(actual.thetaVel, Catch::WithinAbs(expectedThetaVel, angle));
REQUIRE_THAT(actual.xVel, Catch::WithinAbs(expectedXVel, vel));
```
See: https://github.com/catchorg/Catch2/blob/devel/docs/comparing-floating-point-numbers.md

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+const double ATAN_4_3 = 0.927295218002; // value of atan(4/3) in rad.
+const double THETA_KP = 3.5;
+const double NORMAL_DRIVE_SPEED = 4.0;
+const double SLOW_DRIVE_SPEED = 2.0;
+const double DONE_THRESHOLD = 0.25;
+const double CLOSE_TO_TARGET_DUR_VAL = 1.0;
```
abhaybd, line 33: Can you put all of these (and the additional helper method above) in an anonymous namespace? They should be private to this file.

See: https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+TEST_CASE("DriveToWaypointCommand Test") {
```
abhaybd, line 35: This test case is great! Could you also add a short case to check for a different `CLOSE_TO_TARGET_DUR` value?

abhaybd (changes requested):
This is almost there! Just a few comments about using the duration types to their fullest, as well as some minor nitpicks here and there.

Overall code quality is great!

src/commands/DriveToWaypointCommand.h
```diff
@@ -0,0 +1,67 @@
+#include "../navtypes.h"
```
abhaybd, line 1: Oh btw don't forget to add a `#pragma once` at the top here

tests/commands/DriveToWaypointCommandTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+using namespace robot::types;
+using namespace std::chrono_literals;
+
+void assertApprox(double expectedThetaVel, double expectedXVel,
+                  const commands::command_t& actual, double angleThresh = 1e-5,
+                  double velThresh = 1e-5) {
+  REQUIRE_THAT(actual.thetaVel,
+               Catch::WithinAbs(expectedThetaVel, angleThresh));
+  REQUIRE_THAT(actual.xVel, Catch::WithinAbs(expectedXVel, velThresh));
+}
```
abhaybd, line 17: Can you put this method in the anonymous namespace below?

abhaybd (changes requested):
One last minor fix!

abhaybd (approved):
Looks good, this is pretty good quality!

PR #191, opened 2023-01-16
Refactor util from single file to modular setup

abhaybd:
No functionality was added, this just splits up the single file into multiple smaller files with better organization.

This PR is required before I can add some more util classes.


DrDab (approved):
Looks good to me!

PR #193, opened 2023-01-17
Added PeriodicScheduler class

abhaybd:
This is a general purpose class for scheduling functions to be invoked periodically. This will be useful for the motor velocity control tasks.


abhaybd:
Hold up, still working on some improvements. Will let y'all know when I'm done

abhaybd:
Okay, this should be ready for review! I added the ability to use a custom clock which makes testing easier, and also added a method to remove scheduled events.

src/utils/scheduler.h
```diff
@@ -0,0 +1,176 @@
...
+			id = nextID++;
+			auto now = Clock::now();
+			schedule.push({id, now + period, period, fn});
+			newEventAdded = true;
+		}
+		scheduleCV.notify_one();
+		return id;
+	}
+
+private:
```
simranmalhi, line 103: Adding comments for the first two methods would be helpful

src/utils/scheduler.h
```diff
@@ -0,0 +1,176 @@
...
+					// wait until the event should be executed
+					scheduleCV.wait_until(lock, event.nextSendTime,
+										  [&] { return newEventAdded; });
+					newEventAdded = false;
+				}
+			}
+		}
+	}
+
+	struct schedule_t {
```
simranmalhi, line 134: Add comment description for struct

src/utils/scheduler.h
```diff
@@ -0,0 +1,176 @@
...
+		  thread(std::bind(&PeriodicScheduler::threadFn, this)) {}
+
+	/**
+	 * @brief Join the thread and destruct.
+	 */
+	~PeriodicScheduler() {
+		{
+			std::unique_lock lock(scheduleMutex);
+			quitting = true;
+			newEventAdded = true;
```
simranmalhi, line 48: Why is newEventAdded set to true in the destructor?

abhaybd: The condition variable is being notified, so we need to set the associated flag

src/utils/threading.h
```diff
@@ -0,0 +1,74 @@
...
+	bool wait_for(const std::chrono::duration<Rep, Period>& dur) const {
+		std::unique_lock lock(mutex);
+		return cv.wait_for(lock, dur, [&]() { return count <= 0; });
+	}
+
+	/**
+	 * @brief Wait until the latch is unlocked until a specific timepoint.
+	 *
+	 * @param tp The timepoint to wait until.
+	 * @return bool True iff the latch was unlocked before the given time.
```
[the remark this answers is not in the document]
abhaybd: Yeah, this is the same as `wait_for()` except:
* It's often convenient to wait until some time instead of for a duration
* It waits using the clock associated with the timepoint instead of the default

simranmalhi (approved):

PR #197, opened 2023-02-03
Created H264Encoder class

Geeoon:
This is a wrapper class for the Encoder object.  The class is able to encode a cv::Mat into a vector of NALs.

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -118,5 +118,10 @@ constexpr uint32_t RPLIDAR_A1_BAUDRATE = 115200;
 constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 } // namespace Lidar
 
+// Video encoding
+namespace video {
+   const int H264_RF_CONSTANT = 40;
+}
+
 constexpr double CONTROL_HZ = 10.0;
 } // namespace Constants
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -120,7 +120,7 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 
 // Video encoding
 namespace video {
-const int H264_RF_CONSTANT = 40;
+constexpr int H264_RF_CONSTANT = 40;
 }
 
 constexpr double CONTROL_HZ = 10.0;
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -120,7 +120,7 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 
 // Video encoding
 namespace video {
-   const int H264_RF_CONSTANT = 40;
+const int H264_RF_CONSTANT = 40;
 }
 
 constexpr double CONTROL_HZ = 10.0;
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -199,7 +199,8 @@ void MissionControlProtocol::handleCameraStreamOpenRequest(const json& j) {
 	if (supported_cams.find(cam) != supported_cams.end()) {
 		std::unique_lock<std::shared_mutex> stream_lock(this->_stream_mutex);
 		this->_open_streams[cam] = 0;
-		this->_camera_encoders[cam] = std::make_shared<video::H264Encoder>(j["fps"], Constants::video::H264_RF_CONSTANT);
+		this->_camera_encoders[cam] =
+			std::make_shared<video::H264Encoder>(j["fps"], Constants::video::H264_RF_CONSTANT);
 	}
 }
 
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -223,8 +223,8 @@ void MissionControlProtocol::sendJointPositionReport(const std::string& jointNam
 }
 
 void MissionControlProtocol::sendCameraStreamReport(
-	const CameraID& cam, const std::basic_string<uint8_t>& nal_data) {
-	json msg = {{"type", CAMERA_STREAM_REP_TYPE}, {"camera", cam}, {"data", nal_data}};
+	const CameraID& cam, const std::basic_string<uint8_t>& videoData) {
+	json msg = {{"type", CAMERA_STREAM_REP_TYPE}, {"camera", cam}, {"data", videoData}};
 	this->_server.sendJSON(Constants::MC_PROTOCOL_NAME, msg);
 }
 
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -28,7 +28,7 @@ using websocket::connhandler_t;
 using websocket::msghandler_t;
 using websocket::validator_t;
 
-const std::chrono::milliseconds JOINT_POS_REPORT_PERIOD = 100ms;
+const std::chrono::milliseconds TELEM_REPORT_PERIOD = 100ms;
 
 // TODO: possibly use frozen::string for this so we don't have to use raw char ptrs
 // request keys
@@ -48,6 +48,7 @@ constexpr const char* CAMERA_STREAM_REP_TYPE = "cameraStreamReport";
 constexpr const char* LIDAR_REP_TYPE = "lidarReport";
 constexpr const char* MOUNTED_PERIPHERAL_REP_TYPE = "mountedPeripheralReport";
 constexpr const char* JOINT_POSITION_REP_TYPE = "jointPositionReport";
+constexpr const char* ROVER_POS_REP_TYPE = "roverPositionReport";
 // TODO: add support for missing report types
 // autonomousPlannedPathReport, poseConfidenceReport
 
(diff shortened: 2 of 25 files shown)


src/network/MissionControlProtocol.cpp
```diff
@@ -190,6 +190,9 @@ void MissionControlProtocol::handleCameraStreamOpenRequest(const json& j) {
 	if (supported_cams.find(cam) != supported_cams.end()) {
 		std::unique_lock<std::shared_mutex> stream_lock(this->_stream_mutex);
 		this->_open_streams[cam] = 0;
+		if (this->_camera_encoders.find(cam) == this->_camera_encoders.end()) {
+			this->_camera_encoders[cam] = std::make_shared<video::H264Encoder>(j["fps"]);
+		}
```
abhaybd, line 195: This misses the case where the camera stream has been opened and closed in the past, and is being reopened with a different fps.

Since we're using pointers here my earlier advice doesn't apply, so instead of the `if` you can just replace this block with:
```c++
this->_camera_encoders[cam] = std::make_shared<video::H264Encoder>(j["fps"]);
```
This will handle the edge case described above too.

src/network/MissionControlProtocol.h
```diff
@@ -56,7 +59,7 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
 	void handleCameraStreamCloseRequest(const json& j);
 	void handleJointPowerRequest(const json& j);
 	void handleDriveRequest(const json& j);
-	void sendCameraStreamReport(const CameraID& cam, const std::string& b64_data);
+	void sendCameraStreamReport(const CameraID& cam, const std::basic_string<uint8_t>& nal_data);
```
abhaybd, line 62: Can we use a slightly more descriptive name than `nal_data`? How about `frameData` or something? Anyone who hasn't worked with video encoders won't know what `nal` means.

src/video/H264Encoder.h
```diff
@@ -0,0 +1,26 @@
...
+private:
+    std::unique_ptr<Encoder> encoder;
+    bool flag;
+    int fps;
+
+public:
+    H264Encoder(int fps);
+
+    /**
+     * Frame must be the same resolution as passed in the constructor.
```
abhaybd, line 22: What passed in the constructor?

src/video/H264Encoder.h
```diff
@@ -0,0 +1,26 @@
...
+
+class Encoder {  // delete once Encoder is included
+
+};
+
+namespace video {
+class H264Encoder {
+private:
+    std::unique_ptr<Encoder> encoder;
+    bool flag;
```
abhaybd, line 15: What does `flag` mean? Can this have a more descriptive name? What does it do?

src/video/H264Encoder.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+std::vector<std::basic_string<uint8_t>> H264Encoder::encode_frame(const cv::Mat& frame) {
+    if (!encoder) {
+        auto size = frame.size();
+        encoder = std::make_unique<Encoder>(size.width, size.height, size.width, size.height, fps);
+    }
+    std::vector<std::basic_string<uint8_t>> nals;
+    auto frame_size = encoder->encode(frame.data, &flag);
+    for (auto i = 0; i  < encoder->num_nals; i++) {
+        std::basic_string<uint8_t> data = std::basic_string<uint8_t>(encoder->nals[i].p_payload, encoder->nals[i].i_payload);
```
abhaybd, line 16: This can be simplified to:
```c++
std::basic_string<uint8_t> data(encoder->nals[i].p_payload, encoder->nals[i].i_payload);
```
which is a bit cleaner.

abhaybd (changes requested):
Overall looks pretty good! Just some minor changes needed, mostly style or naming conventions.

Also, don't forget to add the implementation file (`H264Encoder.cpp`) to the CMakeLists! I can show you how to do that if you're confused.

Geeoon:
This PR will add the new Encoder to the repository in order to allow camera frames to be encoded in H264.  The past few changes have addressed changes to the encoder library

src/network/MissionControlProtocol.cpp
```diff
@@ -219,9 +221,9 @@ void MissionControlProtocol::sendJointPositionReport(const std::string& jointNam
 	this->_server.sendJSON(Constants::MC_PROTOCOL_NAME, msg);
 }
 
-void MissionControlProtocol::sendCameraStreamReport(const CameraID& cam,
-													const std::string& b64_data) {
-	json msg = {{"type", CAMERA_STREAM_REP_TYPE}, {"camera", cam}, {"data", b64_data}};
+void MissionControlProtocol::sendCameraStreamReport(
+	const CameraID& cam, const std::basic_string<uint8_t>& nal_data) {
+	json msg = {{"type", CAMERA_STREAM_REP_TYPE}, {"camera", cam}, {"data", nal_data}};
```
abhaybd, line 226: Same comment as before, can we rename `nal_data` to something more descriptive? `frameData`, perhaps?

src/video/H264Encoder.h
```diff
@@ -0,0 +1,22 @@
...
+#include <opencv2/core/core.hpp>
+
+namespace video {
+class H264Encoder {
+private:
+	std::unique_ptr<h264encoder::Encoder> encoder;
+	int fps;
+
+public:
+	H264Encoder(int fps);
```
abhaybd, line 15: Can we get doxy comments for the class and the constructor? And also we prefer to put the public section above the private, so can you switch the order?

src/video/H264Encoder.h
```diff
@@ -0,0 +1,22 @@
...
+
+	/**
+	 * first frame will set the resolution, subsequent frames must match the resolution
```
abhaybd, line 18: Very minor, but capitalization pls

src/video/H264Encoder.cpp
```diff
@@ -0,0 +1,20 @@
...
+		encoder = std::make_unique<h264encoder::Encoder>(size.width, size.height, size.width,
+														 size.height, fps);
+	}
+	std::vector<std::basic_string<uint8_t>> nals;
+	auto frame_size = encoder->encode(frame.data);
+	for (auto nal : encoder->nals) {
+		std::basic_string<uint8_t> data(nal.p_payload, nal.i_payload);
+		nals.push_back(data);
+	}
+	return nals;
```
abhaybd, line 18: Can we avoid the usage of the world `nal` everywhere? Nobody really knows what that means

abhaybd (changes requested):
Left some comments, but overall looks good! In general, lets avoid using the world "nal" in places, since it's a very h264 specific thing that we don't need to be too specific about.

Also, since the h264 dependency has been updated, can you update the library version to 2.0.0 in the cmakelists? When the CI reruns it should pass.

Remember to update your version of the library with:
```bash
sudo apt update ; sudo apt install h264encoder
```

src/video/encoder.cpp
```diff
@@ -0,0 +1,120 @@
...
+		x264_param_apply_profile(&prms, "baseline");
+		prms.i_width = out_xres;
+		prms.i_height = out_yres;
+		prms.i_fps_num = fps;
+		prms.i_fps_den = 1;
+		prms.rc.i_qp_constant = 51; // maybe try 0 if it is lagging, max 51
+
+		prms.rc.i_rc_method = X264_RC_CRF;
+		prms.rc.f_rf_constant = 25; // maybe try 51 if it is lagging?
+		prms.rc.f_rf_constant_max = 25;
```
abhaybd, line 52: Can we make `qp_constant` and `rf_constant` parameters to the class? Not great to have hardcoded magic values that would affect the stream resolution

abhaybd (changes requested):

src/Constants.h
```diff
@@ -118,5 +118,10 @@ constexpr uint32_t RPLIDAR_A1_BAUDRATE = 115200;
 constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 } // namespace Lidar
 
+// Video encoding
+namespace video {
+const int H264_RF_CONSTANT = 40;
```
abhaybd, line 123: Minor nitpick: this should be `constexpr`, which is stronger than `const`

src/video/H264Encoder.h
```diff
@@ -0,0 +1,33 @@
...
+namespace video {
+/**
+ * This class is a wrapper for the h264encoder library
+ */
+class H264Encoder {
+public:
+	/**
+	 * Uses the h264encoder library to encode webcam frames into encoded video data
+	 */
+	H264Encoder(int fps, int rf);
```
abhaybd, line 18: Missing `@param` doxygen comments that explain what `fps` and `rf` mean. The constructor documentation should be sufficient for understanding it, users shouldn't have to go look at the comments on private fields.

src/video/H264Encoder.h
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * First frame will set the resolution and subsequent frames must match the resolution
+	 * @returns A vector of video data units representing the encoded frame
+	 */
+	std::vector<std::basic_string<uint8_t>> encode_frame(const cv::Mat& frame);
```
abhaybd, line 23: The `@returns` is nice, just missing a `@param` for `frame`

src/video/H264Encoder.h
```diff
@@ -0,0 +1,33 @@
...
+#include "encoder.hpp"
+
+#include <string>
+#include <vector>
+
+#include <opencv2/core/core.hpp>
+
+namespace video {
+/**
+ * This class is a wrapper for the h264encoder library
```
abhaybd, line 11: This isn't quite correct anymore. Plus, we don't care what it's a wrapper for. Explain what the class does, how it does it is abstracted away from the client.

src/video/encoder.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef ENCODER_HPP_
```
abhaybd, line 1: Lets leave a comment at the top linking to where we got this from.

src/video/encoder.hpp
```diff
@@ -0,0 +1,36 @@
...
+#define ENCODER_HPP_
```
abhaybd, line 2: Also, since we're making this a file in our own repo, let's get rid of the `ifndef` guard and use `#pragma once` instead, since that's consistent with what we do everywhere else.

So to clarify, remove the first two lines (the `ifndef` and `define`) and also the `endif` on the last line, and just put `#pragma once` at the top of the file.

src/video/H264Encoder.h
abhaybd: Also, this is where the min and max of `rf` should be explained. What do higher and lower numbers mean?

abhaybd (changes requested):
Great job getting this working! Some changes are needed regarding documentation, to make sure that users of this code can use it properly.

abhaybd (approved):
Great job! Just make sure this is up to date with master and we can merge

PR #200, opened 2023-02-07
Added motor classes (includes velocity control)

simranmalhi:
This pull request moves all of the motor code into motor classes (base motor, can motor, sim motor). The ability to set the velocity of a motor is also added.

Done:
- Created motor classes (base, can, and sim)
- Implemented method for setting the motor velocity
- Created mapping of motors to shared pointers in world interface
- Created getMotor method in world interface
- Initialized periodic scheduler in base motor constructor
- Updated motor methods in the implementation files for world interface to call methods from motor class
- Added cpp files to CMakeLists
- Created testing method in FakeCanBoard


src/CAN/can_motor.cpp
```diff
@@ -0,0 +1,44 @@
+#include "can_motor.h"
+#include "../world_interface/real_world_constants.h"
+#include "CAN.h"
+#include "CANMotor.h"
+#include "CANUtils.h"
+
+namespace robot {
+class can_motor: public base_motor {
+	public:
```
abhaybd, line 9: Check the style guide, you should be able to automate this with your editor. `public`/`private` directive aren't indented, and the body of a class is indented by one.

src/CAN/can_motor.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "CANMotor.h"
+#include "CANUtils.h"
+
+namespace robot {
+class can_motor: public base_motor {
+	public: 
+		can_motor(robot::types::motorid_t motor, bool hasPosSensor) {
+			motor_id = motor;
+			posSensor = hasPosSensor;
+		}
```
abhaybd, line 13: This type of initialization should be done with a constructor initialization list, not with a series of assignments. So this can be changed to:
```c++
can_motor(robot::types::motorid_t motor, bool hasPosSensor) : motor_id(motor), posSensor(hasPosSensor) {}
```

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,144 @@
...
+
+        // schedule position event
+        velEventID = pSched.scheduleEvent(100ms, [&]() -> void {
+                        types::datatime_t currTime = types::dataclock::now();
+                        const navtypes::Vectord<1> currPos {getMotorPos().getData()};
+                        navtypes::Vectord<1> posCommand = velController->getCommand(currTime, currPos);
+                        // TODO: is there only one position value in the matrix?
+                        setMotorPos(posCommand.coeff(0, 0));
+                    });
+    }
```
abhaybd, line 90: Can we put this implementation in a `base_motor.cpp`? Doesn't need to be in the header file, it seems.

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,144 @@
...
+            navtypes::Vectord<outputDim> res {inputVec(0)};
+            return res; }
+        ;
+
+        // create jacobian function (value will be 1 since it's the derivative of the kinematics function)
+        const std::function<navtypes::Matrixd<outputDim, inputDim>(const navtypes::Vectord<inputDim>&)>& jacobianFunct = 
+            [](const navtypes::Vectord<inputDim>& inputVec)->navtypes::Matrixd<outputDim, inputDim> { 
+                navtypes::Matrixd<outputDim, inputDim> res = navtypes::Matrixd<outputDim, inputDim>::Ones();
+            }
+        ;
```
abhaybd, line 127: This formatting is kinda fucked can you fix that

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,144 @@
...
+
+        velController = {kinematicsFunct, jacobianFunct};
```
abhaybd, line 129: This is copying unnecessarily, we can simplify this with `velController.emplace(kinematicsFunct, jacobianFunct);`

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * @brief Unschedules the velocity event if it exists
+     *
+     */
+    void resetEventID() {
+        if (velEventID) {
+            pSched.removeEvent(velEventID.value());
+            velEventID.reset();
+        }
+        assert(("velEventId has been removed", !velEventID));
```
abhaybd, line 141: This probably helps with debugging, but can you remember to remove some of these assertions when we're ready to merge this?

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,144 @@
...
+    bool posSensor;
+    static util::PeriodicScheduler<std::chrono::steady_clock> pSched;
+    std::optional<util::PeriodicScheduler<std::chrono::steady_clock>::eventid_t> velEventID;
+    std::optional<JacobianVelController<1, 1>> velController;
+
+    /**
+     * @brief Constructs a Jacobian Vel Controller
+     *
+     */
+    void constructVelController() {
```
abhaybd, line 103: Same as above, can we move the definition to an implementation file?

src/CAN/can_motor.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+		bool hasPositionSensor() const {
```
abhaybd, line 15: This functionality is probably common across all motor implementations, so can you move the definition to `base_motor`?

abhaybd (changes requested):
Overall pretty impressive! There are a few things that need fixing, but that's to be expected since this is a draft.

Particularly, make sure you minimize the amount of implementation in header files.

Also, don't forget to add the new `.cpp` files to the CMakeLists! I can show you how if you're unfamiliar with cmake.

src/CAN/can_motor.cpp
simranmalhi: Done

src/CAN/can_motor.cpp
simranmalhi: Done

src/world_interface/base_motor.h
simranmalhi: Done

src/world_interface/base_motor.h
simranmalhi: Done. Not sure if this is enough.

src/world_interface/base_motor.h
simranmalhi: Done

src/world_interface/base_motor.h
simranmalhi: Done

src/world_interface/base_motor.h
simranmalhi: Done

src/CAN/can_motor.h
```diff
@@ -0,0 +1,31 @@
...
+		bool hasPositionSensor() const;
+
+		void setMotorPower(double power);
+
+		void setMotorPos(int32_t targetPos);
+
+        types::DataPoint<int32_t> getMotorPos() const;
+
+    private:
+        void constructVelController();
```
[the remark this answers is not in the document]
simranmalhi: Done. Oh yes it is good catch.

src/CAN/can_motor.cpp
simranmalhi: Done

src/CAN/can_motor.cpp
```diff
@@ -0,0 +1,38 @@
+#include "can_motor.h"
```
abhaybd, line 1: Thinking about it a bit more, since the CAN folder is supposed to be isolated from most of the rest of the codebase, it might make more sense to put `can_motor.cpp` and `can_motor.h` in `src/world_interface` with the rest of the motor files.

It's not ideal to have files in `src/CAN` including files from `src/world_interface`, since the latter is supposed to strictly be a client of the former.

src/CMakeLists.txt
```diff
@@ -189,6 +189,7 @@ target_sources(can_interface PRIVATE
   CAN/CANMotor_PSOC.cpp
   CAN/CANMotor.cpp
   CAN/CANUtils.cpp
+  CAN/can_motor.cpp
```
abhaybd, line 192: After you move the can_motor files, can you put add `can_motor.cpp` to the `real_world_interface` target instead?

src/CMakeLists.txt
```diff
@@ -215,6 +216,8 @@ add_library(world_interface_core STATIC
   ar/MarkerPattern.cpp
   ar/Tag.cpp
   ar/read_landmarks.cpp
+  world_interface/base_motor.cpp
+  world_interface/sim_motor.cpp
```
abhaybd, line 220: Can you put `sim_motor.cpp` in the `simulator_interface` target?

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,94 @@
...
+	 * @param hasPosSensor Boolean to indicate if the motor has a position sensor.
+	 */
+	base_motor(robot::types::motorid_t motor, bool hasPosSensor)
+		: motor_id(motor), posSensor(hasPosSensor) {
+		// create scheduler if needed
+		mtx.lock();
+		if (!pSched) {
+			pSched.emplace();
+		}
+		mtx.unlock();
```
abhaybd, line 32: We should prefer using a [lock guard](https://en.cppreference.com/w/cpp/thread/lock_guard) over manually locking and unlocking

src/world_interface/real_world_interface.cpp
```diff
@@ -75,9 +83,18 @@ void initMotors() {
...
+		// initialize motor objects and add them to map
+		addMotorMapping(motor, true);
 	}
 
 	can::motor::initMotor(motorSerialIDMap.at(motorid_t::hand));
+	addMotorMapping(motorid_t::hand, true);
+}
+
+void addMotorMapping(motorid_t motor, bool hasPosSensor) {
+	std::shared_ptr<robot::base_motor> ptr(new can_motor(motor, hasPosSensor));
```
abhaybd, line 96: We prefer to use `make_shared` instead of the `new` keyword, since that allows us to treat all usages of `new` as a code smell. So, this should be:
```c++
std::shared_ptr<robot::base_motor> ptr = std::make_shared<can_motor>(motor, hasPosSensor);
```

src/world_interface/base_motor.cpp
```diff
@@ -0,0 +1,69 @@
...
+		// create vector for motor position
+		navtypes::Vectord<inputDim> posVector{motorPos.getData()};
+
+		// create kinematics function (input and output will both be the current motor
+		// position)
+		const std::function<navtypes::Vectord<outputDim>(const navtypes::Vectord<inputDim>&)>& kinematicsFunct = 
+            [](const navtypes::Vectord<inputDim>& inputVec) { 
+                // returns a copy of the input vector
+                navtypes::Vectord<outputDim> res {inputVec(0)};
+                return res;
```
abhaybd, line 49: C++ will perform the copy itself, so this can be just shortened to:
```c++
return inputVec;
```

src/world_interface/base_motor.cpp
```diff
@@ -0,0 +1,69 @@
...
+                // returns a copy of the input vector
+                navtypes::Vectord<outputDim> res {inputVec(0)};
+                return res; 
+            };
+
+        // create jacobian function (value will be 1 since it's the derivative of the 
+        // kinematics function)
+        const std::function<navtypes::Matrixd<outputDim, inputDim>(const navtypes::Vectord<inputDim>&)>& jacobianFunct = 
+            [](const navtypes::Vectord<inputDim>& inputVec) { 
+                navtypes::Matrixd<outputDim, inputDim> res = navtypes::Matrixd<outputDim, inputDim>::Ones();
```
abhaybd, line 56: This is correct in our specific case since we're using 1D vectors, but in general this should be an identity matrix, not a ones matrix.

Can you change it from `::Ones()` to `::Identity()`?

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,94 @@
+#pragma once
+#include "data.h"
+#include "../control/JacobianVelController.h"
+#include "../utils/scheduler.h"
+
+using namespace std::chrono_literals;
```
abhaybd, line 6: We should avoid `using` directives in header files. Seems like this isn't used anywhere in this file anyway, so can you just remove it?

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+class base_motor {
+public:
+	/**
+	 * @brief Constructor for base motor.
+	 *
+	 * @param motor The motor id to manipulate.
+	 * @param hasPosSensor Boolean to indicate if the motor has a position sensor.
+	 */
+	base_motor(robot::types::motorid_t motor, bool hasPosSensor)
```
abhaybd, line 25: Can we move this implementation of this method to `base_motor.cpp`? We should minimize implementation in header files.

src/world_interface/base_motor.h
```diff
@@ -0,0 +1,94 @@
...
+	 */
+	void setMotorVel(int32_t targetVel) {}
+
+protected:
+	robot::types::motorid_t motor_id;
+	bool posSensor;
+	static std::optional<util::PeriodicScheduler<std::chrono::steady_clock>> pSched;
+	std::optional<util::PeriodicScheduler<std::chrono::steady_clock>::eventid_t> velEventID;
+	std::optional<JacobianVelController<1, 1>> velController;
+	static std::mutex mtx;
```
abhaybd, line 80: Both `pSched` and `mtx` should be declared as `inline static`. It's a weird quirk in C++, but `static` class members are only *declared* here, not defined. To make this declaration also a definition, we need `inline`. You can read more [here](https://en.cppreference.com/w/cpp/language/static).

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -1,14 +1,17 @@
+#include "../CAN/CANUtils.h"
 #include "../Constants.h"
 #include "../Globals.h"
 #include "../Util.h"
+#include "../control/JacobianVelController.h"
 #include "../kinematics/DiffDriveKinematics.h"
 #include "../log.h"
 #include "../navtypes.h"
+#include "real_world_constants.h"
```
abhaybd, line 9: Are any of these 3 new includes required? The first and last break the hardware agnostic promise made by this file, and the middle seems unnecessary. They seem unused, can you remove them?

src/world_interface/real_world_interface.cpp
```diff
@@ -121,6 +138,18 @@ void world_interface_init() {
 	initMotors();
 }
 
+std::shared_ptr<robot::base_motor> getMotor(robot::types::motorid_t motor) {
+	auto itr = motor_ptrs.find(motor);
+
+	if (itr == motor_ptrs.end()) {
+		// motor id not in map
+		return nullptr;
```
abhaybd, line 146: Can we log an error here? We should know if some code is accessing motors that don't exist. Something like:
```c++
log(LOG_ERROR, "getMotor(): Unknown motor %d\n", static_cast<int>(motor));
```

src/world_interface/sim_motor.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+		json msg = {{"type", "simMotorPositionRequest"},
+					{"motor", motor_name},
+					{"position", targetPos}};
+		sendJSON(msg);
+	}
+
+	types::DataPoint<int32_t> getMotorPos() const {
+		// this calles the method implementation in world interface
+		robot::getMotorPos(motor_id);
```
abhaybd, line 33: I think you forgot `return`

src/world_interface/sim_motor.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "../network/websocket/WebSocketProtocol.h"
+#include "../world_interface/base_motor.h"
+
+#include <nlohmann/json.hpp>
+
+using nlohmann::json;
```
abhaybd, line 8: We should avoid `using` statements in header files.

src/world_interface/simulator_interface.cpp
```diff
@@ -107,12 +107,20 @@ static void openCamera(CameraID cam, std::optional<std::vector<double>> list1d=s
...
+	// openCamera(Constants::AR_CAMERA_ID);
 	openCamera("front", cfg.intrinsicParams->getIntrinsicList());
 	openCamera("rear", cfg.intrinsicParams->getIntrinsicList());
 	openCamera("upperArm", cfg.intrinsicParams->getIntrinsicList());
 }
 
+void initMotors() {
+	// initializes map of motor ids and shared ptrs of their objects
+	for (auto const& x : motorNameMap) {
+		std::shared_ptr<robot::base_motor> ptr(new robot::sim_motor(x.first, true, x.second, PROTOCOL_PATH));
```
abhaybd, line 119: Same as before, we should use `std::make_shared` here

src/world_interface/world_interface.h
```diff
@@ -35,6 +36,9 @@ extern const WorldInterface WORLD_INTERFACE;
 const DiffDriveKinematics& driveKinematics();
 const DiffWristKinematics& wristKinematics();
 
+// A mapping of (motor_id, shared pointer to object of the motor)
+std::unordered_map<robot::types::motorid_t, std::shared_ptr<robot::base_motor>> motor_ptrs;
```
abhaybd, line 40: This shouldn't be defined in a header file, it should be private to each world interface implementation. Just remove this line and copy it to the top of `real_world_interface.cpp` and `simulator_interface.cpp`

abhaybd (changes requested):
This is looking good, almost there! There are a few thing that need to be addressed, most of which I've noted inline.

First, can you move all the motor implementations to `src/world_interface/motor`? We're adding 6 new files, so putting them all in world_interface will create clutter. Remember to update the includes and the cmake.

Also, in a few places there are empty method definitions when there should be just method declarations. I've noted the first one but not the rest. Basically, anywhere in header files where there is an empty method defined, just replace the body with a semicolon.

Let me know if you have any questions!

src/world_interface/kinematic_common_interface.cpp
simranmalhi: Ah yeah I changed the kinematic common interface on accident instead of real world interface and forgot to remove the include statements

src/world_interface/sim_motor.cpp
simranmalhi: Great catch

src/world_interface/noop_world.cpp
```diff
@@ -57,6 +62,8 @@ types::DataPoint<int32_t> getMotorPos(robot::types::motorid_t motor) {
 	return {};
 }
 
+void setMotorVel(int32_t targetVel) {}
```
abhaybd, line 65: Don't think this is the right signature, it should also have a `motorid_t` parameter.

src/world_interface/motor/can_motor.cpp
```diff
@@ -0,0 +1,36 @@
...
+	// unschedule velocity event if exists
+	unscheduleVelocityEvent();
+
+	can::deviceserial_t serial = motorSerialIDMap.at(motor_id);
+	auto& scaleMap = power < 0 ? negative_pwm_scales : positive_pwm_scales;
+	auto entry = scaleMap.find(motor_id);
+	if (entry != scaleMap.end()) {
+		power *= entry->second;
+	}
+	can::motor::setMotorPower(serial, power);
```
abhaybd, line 22: Can we also move the logic of `ensureMotorMode` from `real_world_interface` to here? The motor mode should just be a field of this class, and we should do the same logic of setting the motor mode if required.

After that, we can get rid of the motor mode map in `real_world_interface`.

And remember to ensure motor modes for the setPos and setVel methods too. For the latter, override setVel, ensure the motor mode, and then invoke the base class implementation.

src/world_interface/simulator_interface.cpp
```diff
@@ -106,12 +108,21 @@ static void openCamera(CameraID cam, std::optional<std::vector<double>> list1d=s
...
-	//openCamera(Constants::AR_CAMERA_ID);
+	// openCamera(Constants::AR_CAMERA_ID);
 	openCamera("front", cfg.intrinsicParams->getIntrinsicList());
 	openCamera("rear", cfg.intrinsicParams->getIntrinsicList());
 	openCamera("upperArm", cfg.intrinsicParams->getIntrinsicList());
 }
 
+void initMotors() {
+	// initializes map of motor ids and shared ptrs of their objects
+	for (auto const& x : motorNameMap) {
```
abhaybd, line 119: We put `const` before type, so can you change this to `const auto& x`? We should add a rule to the clang-format for that later.

src/world_interface/simulator_interface.cpp
```diff
@@ -262,11 +273,24 @@ extern const WorldInterface WORLD_INTERFACE = WorldInterface::sim3d;
...
 	AR::initializeLandmarkDetection();
 }
 
-std::unordered_set<CameraID> getCameras(){
+std::shared_ptr<robot::base_motor> getMotor(robot::types::motorid_t motor) {
+	auto itr = motor_ptrs.find(motor);
+
+	if (itr == motor_ptrs.end()) {
+		// motor id not in map
+		return nullptr;
```
abhaybd, line 286: Same as in the real world interface, can we log an error here?

src/world_interface/world_interface.h
```diff
@@ -198,6 +207,9 @@ navtypes::URCLeg getLeg(int index);
  */
 void setIndicator(types::indication_t signal);
 
+// TODO: update implementation in the files that inherit this (now have motor class
+// implementation)
```
abhaybd, line 211: Seems vestigial. Remove?

src/world_interface/motor/base_motor.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+	// check to see if the event exists. if yes, unschedule it
+	base_motor::unscheduleVelocityEvent();
+
+	// schedule position event
+	velEventID = pSched->scheduleEvent(100ms, [&]() -> void {
+		types::datatime_t currTime = types::dataclock::now();
+		const navtypes::Vectord<1> currPos{getMotorPos().getData()};
+		navtypes::Vectord<1> posCommand = velController->getCommand(currTime, currPos);
+		// TODO: is there only one position value in the matrix?
```
abhaybd, line 38: Yeah, since all the dimensions are one, there's only one. So this todo can be removed

abhaybd (changes requested):
This is turning into a massive PR, kudos for being able to handle all this. It's sooo close now, just some more changes are needed, in the comments below. Lmk if you have questions!

src/control/JacobianVelController.h
```diff
@@ -4,6 +4,7 @@
 #include "../filters/StateSpaceUtil.h"
 #include "../navtypes.h"
 #include "../utils/math.h"
+#include "../world_interface/data.h"
```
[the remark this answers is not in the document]
simranmalhi: When I was including the JacobianVelController.h in my base motor class, it required that I include data.h before JacobianVelController.h in order for my code to build properly. However, the formatting check was failing because it wanted data.h to be at the end of my include statements, which would cause the build check to fail, so this was my workaround to get all of the checks to pass.

src/world_interface/motor/base_motor.cpp
```diff
@@ -0,0 +1,81 @@
...
+void base_motor::constructVelController() {
+	types::DataPoint<int32_t> motorPos = getMotorPos();
+	if (!motorPos.isValid()) {
+		return;
+	}
+	constexpr int32_t inputDim = 1;
+	constexpr int32_t outputDim = 1;
+
+	// create vector for motor position
+	navtypes::Vectord<inputDim> posVector{motorPos.getData()};
```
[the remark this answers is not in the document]
simranmalhi: Oh it's supposed to get passed into the kinematics function below. Looks like I removed that accidently

src/world_interface/motor/can_motor.h
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @param motor The motor id to manipulate.
+	 * @param hasPosSensor Boolean to indicate if the motor has a position sensor.
+	 * @param serial The serial number of the motor board.
+	 * @param pos_pwm_scales The positive pwm scales for the motor power.
+	 * @param neg_pwm_scales The positive pwm scales for the motor power.
+	 */
+	can_motor(robot::types::motorid_t motor, bool hasPosSensor, can::deviceserial_t serial,
+			  frozen::unordered_map<robot::types::motorid_t, double, 11UL> pos_pwm_scales,
+			  frozen::unordered_map<robot::types::motorid_t, double, 11UL> neg_pwm_scales);
```
abhaybd, line 26: Only take the scales for this specific motor, no need to pass the entire map.

src/world_interface/real_world_interface.cpp
```diff
@@ -91,13 +90,23 @@ void initMotors() {
...
+	addMotorMapping(motorid_t::rearLeftWheel, false);
+	addMotorMapping(motorid_t::rearRightWheel, false);
+
 	for (motorid_t motor : pidMotors) {
 		can::deviceserial_t serial = motorSerialIDMap.at(motor);
 		pidcoef_t pid = motorPIDMap.at(motor);
 		can::motor::setMotorPIDConstants(serial, pid.kP, pid.kI, pid.kD);
+
+		// initialize motor objects and add them to map
+		addMotorMapping(motor, true);
```
abhaybd, line 105: I'd say move this to the enc motor and pot motor for loops, since I'm not 100% sure that every positional motor is also a pid motor. So to be clear, the two loops would each have an `addMotorMapping()` call.

abhaybd (changes requested):
So close! I think this should *finally* be the last one.

abhaybd (commented):
This looks great, well done! I'll hold off on accepting it, since we need to do some hardware testing to ensure it all works. Let's try to do that (or at least get started) on saturday.

src/CAN/FakeCANBoard.cpp
```diff
@@ -101,6 +109,48 @@ int main() {
...
+				motor.setMotorVel(targetVel);
+				startTime = robot::types::dataclock::now();
+			}
+
+			// get x data: current time
+			robot::types::datatime_t currTime = robot::types::dataclock::now();
+
+			// get y data: set point (target vel * time since set velocity call)
+			double setPoint = targetVel * fabs((currTime.time_since_epoch().count() -
+												startTime.time_since_epoch().count()));
```
abhaybd, line 145: I believe this will multiply by the elapsed time in millis, but we need seconds. We can be pretty explicit about it though by first doing `util::durationToSec(currTime - startTime)`, which will give us elapsed time in seconds.

src/CAN/FakeCANBoard.cpp
```diff
@@ -101,6 +109,48 @@ int main() {
...
+
+			// get y data: motor position
+			robot::types::DataPoint<int32_t> motorPos = motor.getMotorPos();
+
+			// print data
+			std::cout << "current time: " << currTime.time_since_epoch().count()
```
abhaybd, line 151: Print elapsed time (`util::durationToSec(currTime-startTime)`) instead of time since epoch, that'll be a lot more human readable.

src/CAN/FakeCANBoard.cpp
abhaybd: Also add the initial position to the setpoint, it's basically the +C since we're integrating velocity.

src/CAN/FakeCANBoard.cpp
```diff
@@ -33,6 +35,7 @@ enum class TestMode {
 std::unordered_set<int> modes = {
 	static_cast<int>(TestMode::ModeSet),	   static_cast<int>(TestMode::PWM),
 	static_cast<int>(TestMode::PID),		   static_cast<int>(TestMode::Encoder),
+	static_cast<int>(TestMode::PIDVel),		   static_cast<int>(TestMode::Encoder),
```
abhaybd, line 38: looks like we're repeating `TestMode::Encoder` twice, can you remove the second occurrence?

abhaybd (changes requested):
Overall good, just some fixes needed to the fakecanboard testing

src/world_interface/motor/base_motor.h
```diff
@@ -0,0 +1,83 @@
...
+class base_motor {
+public:
+	/**
+	 * @brief Constructor for base motor.
+	 *
+	 * @param motor The motor id to manipulate.
+	 * @param hasPosSensor Boolean to indicate if the motor has a position sensor.
+	 */
+	base_motor(robot::types::motorid_t motor, bool hasPosSensor);
+
```
abhaybd, line 24: Copying motors should be banned, so we should delete the copy constructor and operator:

```
base_motor(const base_motor&) = delete;

base_motor& operator=(const base_motor&) = delete;
```

abhaybd (changes requested):
Some notes on the fakecanboard, and also a note about copy constructors

abhaybd (dismissed):
Looks good! Great job getting this done, this was a pretty big task. Resolve merge conflicts and you should be good to merge

abhaybd (approved):

PR #204, opened 2023-02-10
Add format checker

abhaybd:
This adds a CI pipeline to check if newly added code adheres to our style guide.

### DIFF
--- a/.github/workflows/format-checker.yml
+++ b/.github/workflows/format-checker.yml
@@ -1,13 +1,28 @@
 name: Check Formatting
-on: [push, pull_request]
+
+on: [pull_request]
+
 jobs:
-  formatting-check:
-    name: Formatting Check
+  build:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v3
-    - name: Run clang-format style check for C/C++/Protobuf programs.
-      uses: jidicula/clang-format-action@v4.10.2
+    - name: Checkout
+      uses: actions/checkout@v1
       with:
-        clang-format-version: '13'
-        check-path: 'src'
+        submodules: true
+    - name: Install dependencies from APT
+      run: |
+        sudo apt install git clang-format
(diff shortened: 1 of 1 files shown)


.github/workflows/format-checker.yml
```diff
@@ -0,0 +1,28 @@
...
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Install dependencies from APT
+      run: |
+        sudo apt install git clang-format
```
huttongrabiel, line 15: Should probably specify -y to auto confirm any prompt from apt.

abhaybd: Oh good catch I'll add that

huttongrabiel (approved):

huttongrabiel:
lgtm

huttongrabiel:
Non-important to merging this, code looks fine, but you might consider squashing the three commits into one. A lot of the first commit is changed in the second, and the third commit is a two character fix to the second. Not sure how strict you want commit history here, but it could be nice. However, not at all necessary. :+1:

abhaybd:
> Non-important to merging this, code looks good, but you might consider squashing the three commits into one. A lot of the first commit is changed in the second, and the third commit is a two character fix to the second. Not sure how strict you want commit history here, but it could be nice. However, not at all necessary. :+1: 

All PRs are automatically squashed while merging to master, so that will be addressed. It's good you're keeping that in mind though!

huttongrabiel:
> > Non-important to merging this, code looks good, but you might consider squashing the three commits into one. A lot of the first commit is changed in the second, and the third commit is a two character fix to the second. Not sure how strict you want commit history here, but it could be nice. However, not at all necessary. +1
> 
> All PRs are automatically squashed while merging to master, so that will be addressed. 

Good to know!

PR #215, opened 2023-02-18
Add code to set limit switch limits

abhaybd:
This adds code in the CAN layer to allow configuring the limit switch limits, so that the motor boards can reset encoder positions when they are triggered.

### DIFF
--- a/src/world_interface/real_world_constants.h
+++ b/src/world_interface/real_world_constants.h
@@ -7,7 +7,6 @@
 #include <cstdint>
 #include <unordered_map>
 #include <unordered_set>
-#include <utility>
 
 #include <frozen/unordered_map.h>
 #include <frozen/unordered_set.h>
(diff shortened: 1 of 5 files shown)


src/world_interface/real_world_constants.h
```diff
@@ -52,11 +53,24 @@ struct encparams_t {
...
+	/** Limit switch high, in millidegrees */
+	int limitSwitchHigh;
 };
 
-constexpr auto encMotors = frozen::make_unordered_map<motorid_t, encparams_t>(
-	{{motorid_t::shoulder, {.isInverted = false, .ppjr = 1620 * 1024}},
-	 {motorid_t::elbow, {.isInverted = true, .ppjr = 1620 * 1024}}});
+// TODO: verify limit switch limits
+constexpr auto encMotors =
+	frozen::make_unordered_map<motorid_t, encparams_t>({{motorid_t::shoulder,
```
huttongrabiel, line 64: Update formatting as we talked about

src/world_interface/real_world_constants.h
```diff
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <unordered_map>
 #include <unordered_set>
+#include <utility>
```
huttongrabiel, line 10: Unnecessary include

huttongrabiel (changes requested):
Few minor fixes. Formatting and an extra include. All else looks good.

huttongrabiel (approved):

PR #222, opened 2023-02-28
Send telem data

sarahfredrickson:
Added method to send telemetry data from rover to MC.


src/network/MissionControlProtocol.cpp
```diff
@@ -176,6 +177,31 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
...
+	auto gps = robot::readGPS();
+	if (gps.isValid() && heading.isValid()) {
+		double orientX = 0.0;
+		double orientY = 0.0;
+		double orientZ = sin(heading.getData() / 2);
+		double orientW = cos(heading.getData() / 2);
+		double posX = gps.getData()[0];
+		double posY = gps.getData()[1];
+		double posZ = 1.0;
+		double recency = util::durationToSec(dataclock::now() - gps.getTime());
```
abhaybd, line 191: Maybe take the older of the gps and imu timestamps? The recency should show the age of the oldest data in the packet.

abhaybd (changes requested):
Overall pretty good! A minor thing is it's not exactly clear what telemetry means here. Let's say that both joint positions and rover positions are telemetry data. So the individual packet types should jointPos and roverPos respectively, and the task that reports them is the telemetry report task.

There are some other minor fixes needed that I left in comments. Lmk if you have questions!

src/network/MissionControlProtocol.cpp
```diff
@@ -176,6 +177,34 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
...
+		double orientY = 0.0;
+		double orientZ = sin(heading.getData() / 2);
+		double orientW = cos(heading.getData() / 2);
+		double posX = gps.getData()[0];
+		double posY = gps.getData()[1];
+		double posZ = 0.0;
+		double recency = util::durationToSec(dataclock::now() - gps.getTime()) >
+								 util::durationToSec(dataclock::now() - heading.getTime())
+							 ? util::durationToSec(dataclock::now() - gps.getTime())
+							 : util::durationToSec(dataclock::now() - heading.getTime());
```
abhaybd, line 194: This statement is way too long. Can you put all the `util::durationToSec` calls into variables, and then just simplify this by calling `std::max`?

I.e.:
```c++
double gpsRecency = util::durationToSec(...);
double headingRecency = util::durationToSec(...);
double recency = std::max(gpsRecency, headingRecency);
```
Which is both easier to read and shorter.

src/network/MissionControlProtocol.cpp
```diff
@@ -370,7 +396,7 @@ void MissionControlProtocol::jointPowerRepeatTask() {
 	}
 }
 
-void MissionControlProtocol::jointPosReportTask() {
+void MissionControlProtocol::roverPosReportTask() {
```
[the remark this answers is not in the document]
abhaybd: This still hasn't been addressed

src/network/MissionControlProtocol.h
```diff
@@ -61,6 +61,8 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
 	void handleDriveRequest(const json& j);
 	void sendJointPositionReport(const std::string& jointName, int32_t position);
 	void sendCameraStreamReport(const CameraID& cam, const std::string& b64_data);
+	void sendRoverPos();
+	void roverPosReportTask();
```
abhaybd, line 65: We wanted to rename `roverPosReportTask` to `telemReportTask`, so this method declaration shouldn't be here. Just deleting this line should be fine.

abhaybd (changes requested):
Almost there! Some comments from last time weren't fully addressed, and then some minor cleanup required.

src/network/MissionControlProtocol.cpp
```diff
@@ -176,6 +177,33 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
...
 }
 
+void MissionControlProtocol::sendRoverPos() {
+	auto heading = robot::readIMUHeading();
+	auto gps = robot::readGPS();
+	if (gps.isValid() && heading.isValid()) {
+		double orientX = 0.0;
+		double orientY = 0.0;
+		double orientZ = sin(heading.getData() / 2);
+		double orientW = cos(heading.getData() / 2);
```
abhaybd, line 187: I'd say change `sin` and `cos` for `std::sin` and `std::cos`, respectively. `std` methods tend to play better with types and such, so we prefer to use them.

abhaybd (changes requested):
Looks great! Just one minor change.

abhaybd (approved):
Looks good!

PR #223, opened 2023-02-28
Create AutonomousTask to handle autonomous navigation

huttongrabiel:
- Use DriveToWaypointCommand functionality to execute an autonomous task.

- Add necessary constants to Constants.h

- Add handleWaypointNavRequest to handle waypointNavRequest packet from MissionControl, starts autonomous navigation

- Start autonomous task from MissionControlProtocol.cpp

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -254,8 +254,12 @@ add_library(control SHARED
 )
 target_link_libraries(control utils)
 
-add_library(commands STATIC
+add_library(commands SHARED
   commands/DriveToWaypointCommand.cpp)
+
+add_library(autonomous SHARED
+  autonomous/AutonomousTask.cpp)
+target_link_libraries(autonomous commands)
 
 list(APPEND simulator_libs
   simulator_interface)
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "kinematics/DiffDriveKinematics.h"
+#include "utils/time.h"
 #include "world_interface/data.h"
 
 #include <array>
@@ -184,3 +185,12 @@ constexpr frozen::unordered_map<robot::types::motorid_t, double, IK_MOTORS.size(
 
 constexpr double CONTROL_HZ = 10.0;
 } // namespace Constants
+
+namespace autonomous {
+constexpr double THETA_KP = 2.0;
+constexpr double DRIVE_VEL = 1.5;
+constexpr double SLOW_DRIVE_VEL = 1.0;
+constexpr double DONE_THRESHOLD = 0.5;
+// Duration long enough to confirm we are there, not so long that time is wasted
+constexpr util::dseconds CLOSE_TO_TARGET_DUR_VAL = std::chrono::seconds(1);
+} // namespace autonomous
--- a/src/autonomous/AutonomousTask.cpp
+++ b/src/autonomous/AutonomousTask.cpp
@@ -1,7 +1,6 @@
 #include "AutonomousTask.h"
 
 #include "../Constants.h"
-#include "../Globals.h"
 #include "../commands/DriveToWaypointCommand.h"
 #include "../world_interface/world_interface.h"
 
@@ -31,6 +30,8 @@ void AutonomousTask::navigate() {
 	commands::DriveToWaypointCommand cmd(_waypoint_coords, THETA_KP, DRIVE_VEL, SLOW_DRIVE_VEL,
 										 DONE_THRESHOLD, CLOSE_TO_TARGET_DUR_VAL);
 
+    DiffDriveKinematics diffDriveKinematics(Constants::EFF_WHEEL_BASE);
+
 	auto sleepUntil = std::chrono::steady_clock().now();
 	while (!cmd.isDone()) {
 		auto latestGPS = robot::readGPS();
@@ -41,7 +42,8 @@ void AutonomousTask::navigate() {
 			navtypes::pose_t latestPos(gpsData.x(), gpsData.y(), latestHeading.getData());
 			cmd.setState(latestPos, robot::types::dataclock::now());
 			commands::command_t output = cmd.getOutput();
-			robot::setCmdVel(output.thetaVel, output.xVel);
+            auto scaledVels = diffDriveKinematics.ensureWithinWheelSpeedLimit(DiffDriveKinematics::PreferredVelPreservation::PreferThetaVel, output.xVel, output.thetaVel, Constants::MAX_WHEEL_VEL);
+			robot::setCmdVel(scaledVels(2), scaledVels(0));
 		}
 
 		std::unique_lock autonomousTaskLock(_autonomous_task_mutex);
--- a/src/autonomous/AutonomousTask.cpp
+++ b/src/autonomous/AutonomousTask.cpp
@@ -30,7 +30,7 @@ void AutonomousTask::navigate() {
 	commands::DriveToWaypointCommand cmd(_waypoint_coords, THETA_KP, DRIVE_VEL, SLOW_DRIVE_VEL,
 										 DONE_THRESHOLD, CLOSE_TO_TARGET_DUR_VAL);
 
-    DiffDriveKinematics diffDriveKinematics(Constants::EFF_WHEEL_BASE);
+	DiffDriveKinematics diffDriveKinematics(Constants::EFF_WHEEL_BASE);
 
 	auto sleepUntil = std::chrono::steady_clock().now();
 	while (!cmd.isDone()) {
@@ -42,7 +42,9 @@ void AutonomousTask::navigate() {
 			navtypes::pose_t latestPos(gpsData.x(), gpsData.y(), latestHeading.getData());
 			cmd.setState(latestPos, robot::types::dataclock::now());
 			commands::command_t output = cmd.getOutput();
-            auto scaledVels = diffDriveKinematics.ensureWithinWheelSpeedLimit(DiffDriveKinematics::PreferredVelPreservation::PreferThetaVel, output.xVel, output.thetaVel, Constants::MAX_WHEEL_VEL);
+			auto scaledVels = diffDriveKinematics.ensureWithinWheelSpeedLimit(
+				DiffDriveKinematics::PreferredVelPreservation::PreferThetaVel, output.xVel,
+				output.thetaVel, Constants::MAX_WHEEL_VEL);
 			robot::setCmdVel(scaledVels(2), scaledVels(0));
 		}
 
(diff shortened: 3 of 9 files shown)


src/CMakeLists.txt
```diff
@@ -280,6 +283,7 @@ list(APPEND simulator_libs
 add_executable(Rover Rover.cpp)
 add_library(rover_common SHARED
   Autonomous.cpp
+  autonomous/AutonomousTask.cpp
```
abhaybd, line 286: This file is already included in the `autonomous` library, don't think we need to put it here as well.

src/CMakeLists.txt
```diff
@@ -303,6 +307,7 @@ list(APPEND rover_libs
   control
   commands
   utils
+  autonomous
```
abhaybd, line 310: Instead of linking `rover_libs` to `autonomous`, should we be linking `mission_control_interface` to `autonomous`? Since we're getting rid of the usages in `Rover.cpp`, that should be the only user I think.

src/Constants.h
```diff
@@ -120,3 +120,15 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 
 constexpr double CONTROL_HZ = 10.0;
 } // namespace Constants
+
+using namespace std::chrono_literals;
```
abhaybd, line 124: We should avoid `using` statements in header files. So the `1s` below should be `std::chrono::seconds(1)`.

src/Rover.cpp
```diff
@@ -131,7 +132,9 @@ int main(int argc, char** argv) {
 	parseCommandLine(argc, argv);
 	Globals::AUTONOMOUS = false;
 	Globals::websocketServer.start();
-	auto mcProto = std::make_unique<net::mc::MissionControlProtocol>(Globals::websocketServer);
+	autonomous::AutonomousTask autonomous_task;
+	auto mcProto = std::make_unique<net::mc::MissionControlProtocol>(Globals::websocketServer,
```
abhaybd: As we discussed, the autonomous task instantiation should happen in `MissionControlProtocol`.

src/autonomous/AutonomousTask.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void AutonomousTask::navigate() {
+	commands::DriveToWaypointCommand cmd(_waypoint_coords, THETA_KP, DRIVE_VEL, SLOW_DRIVE_VEL,
+										 DONE_THRESHOLD, CLOSE_TO_TARGET_DUR_VAL);
+
+	while (!cmd.isDone()) {
+		auto latestGPS = robot::readGPS();
+		if (latestGPS.isFresh(2000ms)) {
+			cmd.setState(latestGPS, robot::types::dataclock::now());
```
abhaybd, line 31: `latestGPS` is a `point_t`, so it has `[x, y, 1]`, but we want a `pose_t`, which has `[x, y, theta]`. Confusingly, they're both `Eigen::Vector3d`, so the compiler doesn't complain.

GPS only gives position information, so you'll need to get heading from the IMU. Call `robot::readIMUHeading()` to get that and assemble that into a single `pose_t`.

src/autonomous/AutonomousTask.cpp
```diff
@@ -0,0 +1,57 @@
...
+			commands::command_t output = cmd.getOutput();
+			robot::setCmdVel(output.thetaVel, output.xVel);
+		}
+		std::unique_lock autonomousTaskLock(_autonomous_task_mutex);
+		// Wait 1000ms or return out of while loop if kill called
+		if (_autonomous_task_cv.wait_until(autonomousTaskLock,
+										   std::chrono::system_clock::now() + 1000ms,
+										   [&] { return _kill_called; })) {
+			return;
+		}
```
abhaybd, line 41: This sleeps for 1000ms, so it assumes that the rest of the code runs in 0 time. That's not true, so we'll get an unsteady update rate.

Instead, initialize a `sleepUntil` time outside of the loop to the current time, and increment it by 1000ms and sleep every time. This will make it a more consistent update rate.

abhaybd: Also, we should use `std::chrono::steady_clock` instead of `std::chrono::system_clock`. Read about the difference [here](https://stackoverflow.com/questions/31552193/difference-between-steady-clock-vs-system-clock).

src/autonomous/AutonomousTask.cpp
```diff
@@ -0,0 +1,57 @@
...
+AutonomousTask::~AutonomousTask() {
+	if (_autonomous_task_thread.joinable()) {
+		_autonomous_task_thread.join();
+	}
+}
+
+void AutonomousTask::start(const navtypes::pose_t& waypointCoords) {
+	setWaypoint(waypointCoords);
+	_kill_called = false;
+	_autonomous_task_thread = std::thread(&autonomous::AutonomousTask::navigate, this);
```
abhaybd: Check if the task is currently running, and if it is, kill it and restart.

src/autonomous/AutonomousTask.cpp
```diff
@@ -0,0 +1,57 @@
...
+		if (_autonomous_task_cv.wait_until(autonomousTaskLock,
+										   std::chrono::system_clock::now() + 1000ms,
+										   [&] { return _kill_called; })) {
+			return;
+		}
+	}
+}
+
+void AutonomousTask::kill() {
+	_kill_called = true;
```
abhaybd, line 46: `_kill_called` should be protected with a mutex. Also, after tripping the flag, you should notify the CV. Something like:

```C++
{
    std::lock_guard lock(autonomous_mutex);
    _kill_called = true;
}
_autonomous_task_cv.notify_all();
```

src/autonomous/AutonomousTask.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+	if (_autonomous_task_thread.joinable()) {
+		_autonomous_task_thread.join();
+	}
+}
+
+void AutonomousTask::setWaypoint(const navtypes::pose_t& waypointCoords) {
+	_waypoint_coords = waypointCoords;
+}
```
abhaybd: Don't think we need this method

src/network/MissionControlProtocol.h
```diff
@@ -35,6 +36,7 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
 	void jointPowerRepeatTask();
 	void jointPosReportTask();
 	SingleClientWSServer& _server;
+	autonomous::AutonomousTask& _autonomous_task;
```
abhaybd: We already talked about this, but should be a value not a ref.

src/network/MissionControlProtocol.h
```diff
@@ -24,7 +25,7 @@ using websocket::WebSocketProtocol;
 
 class MissionControlProtocol : public WebSocketProtocol { // TODO: add documentation
 public:
-	MissionControlProtocol(SingleClientWSServer& server);
+	MissionControlProtocol(SingleClientWSServer& server, autonomous::AutonomousTask&);
```
abhaybd: As discussed, don't need to take it as an argument.

src/autonomous/AutonomousTask.h
```diff
@@ -0,0 +1,60 @@
...
+#include "../navtypes.h"
+
+#include <cmath>
+#include <condition_variable>
+#include <mutex>
+#include <thread>
+
+namespace autonomous {
+
+class AutonomousTask {
```
abhaybd, line 12: Can we add a documentation comment for the class?

abhaybd (changes requested):
Pretty good! Some comments about sleeping, using a CV, etc.

src/autonomous/AutonomousTask.cpp
huttongrabiel: Would appreciate if you could double check this carefully. I think I got it but not sure.

abhaybd (changes requested):
Pretty good, just some more minor fixes needed

src/network/MissionControlProtocol.cpp
```diff
@@ -176,6 +179,25 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
...
+}
+
+void MissionControlProtocol::handleWaypointNavRequest(const json& j) {
+	double latitude = j["latitude"];
+	double longitude = j["longitude"];
+	bool isApproximate = j["isApproximate"];
+	bool isGate = j["isGate"];
+
+	if (Globals::AUTONOMOUS && !isApproximate && !isGate) {
+		navtypes::pose_t waypointCoords(latitude, longitude, 1);
```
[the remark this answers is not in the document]
huttongrabiel: Should I then change the start method in AutonomousTask to take point_t instead of pose_t?

src/CMakeLists.txt
```diff
@@ -263,6 +263,10 @@ target_link_libraries(control utils)
 add_library(commands STATIC
   commands/DriveToWaypointCommand.cpp)
 
+add_library(autonomous SHARED
+  autonomous/AutonomousTask.cpp
+  commands/DriveToWaypointCommand.cpp)
```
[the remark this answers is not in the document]
huttongrabiel: Done but required changing commands to SHARED.

abhaybd (commented):
This looks good! Let's try to do some testing to ensure it works well before merging. Testing in sim would be fine, let's discuss how to do that on saturday.

src/network/MissionControlProtocol.cpp
```diff
@@ -221,6 +224,25 @@ void MissionControlProtocol::sendRoverPos() {
 	}
 }
 
+static bool validateWaypointNavRequest(const json& j) {
+	return util::validateKey(j, "latitude", val_t::number_unsigned) &&
+		   util::validateKey(j, "longitude", val_t::number_unsigned) &&
```
abhaybd, line 229: latitude and longitude aren't unsigned ints, this should probably be `val_t::number_float`?

src/network/MissionControlProtocol.cpp
```diff
@@ -221,6 +224,25 @@ void MissionControlProtocol::sendRoverPos() {
...
+		   util::validateKey(j, "approximate", val_t::boolean) &&
+		   util::validateKey(j, "gated", val_t::boolean);
```
abhaybd, line 231: According to the MC readme, these fields should be called `isGate` and `isApproximate`, is that not the case?

abhaybd (changes requested):
Looks good, just some questions about the MC protocol

src/network/MissionControlProtocol.cpp
huttongrabiel: It only validates if they are number_unsigned. I battled it for like an hour. I saw theyre floats in MissionControl but it won't validate as floats. If you have an idea that'd be great.

src/network/MissionControlProtocol.cpp
huttongrabiel: This was just a quick fix. They are not called isGate and isApproximate in the mission control code. They got changed somewhere along the line. The current message uses gate and approximate. Either the MC readme needs to change or the JS code on MC side

PR #224, opened 2023-03-01
Set dep versions

abhaybd:
Our build breaks whenever one of our team dependencies has a newer version available that we don't want to migrate to yet. Explicitly specifying the version will fix this.


huttongrabiel:
Is h264encoder 2.0.0 or 1.1.1? in https://github.com/huskyroboticsteam/CPP-H264-Encoder/blob/main/CMakeLists.txt, it is 2.0.0.

PR #225, opened 2023-03-02
Fix misc sim bugs

abhaybd:
Fix crashes for unsupported motors, and fixed units of joint position report.


DrDab (approved):
Don't see any issues. Looks good to me!

PR #226, opened 2023-03-10
Update prompt to handle invalid input and use modern language features

huttongrabiel:
Prompt will now continue to prompt as long as it reads invalid input. Invalid input is defined as input that cannot be converted to an integer. It still accepts any base integer (hex, octal, decimal, etc.)

### DIFF
--- a/src/CAN/FakeCANBoard.cpp
+++ b/src/CAN/FakeCANBoard.cpp
@@ -36,11 +36,22 @@ std::unordered_set<int> modes = {
 	static_cast<int>(TestMode::LimitSwitch),   static_cast<int>(TestMode::ScienceTelemetry),
 	static_cast<int>(TestMode::ScienceMotors), static_cast<int>(TestMode::ScienceServos)};
 
-int prompt(const char* msg) {
+int prompt(std::string_view message) {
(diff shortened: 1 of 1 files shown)


src/CAN/FakeCANBoard.cpp
```diff
@@ -36,11 +36,20 @@ std::unordered_set<int> modes = {
...
-	int val = std::stoi(str, nullptr, 0);
+	int val;
+	bool valid_input = false;
+	do {
+		std::cout << message << " > ";
+		std::getline(std::cin, str);
+		try {
+			val = std::stoi(str, nullptr, 0);
+			valid_input = true;
+		} catch (const std::exception&) {
```
abhaybd, line 49: Catching all exceptions is generally a code smell. Can we explicitly catch the errors we care about, i.e. `std::invalid_argument` and `std::out_of_range`?

abhaybd (changes requested):
Yeah this code is pretty anachronistic, so updating it is good. Just a minor fix needed about exception handling.

abhaybd (approved):
Looks good!

abhaybd (approved):

PR #228, opened 2023-04-18
negate hand pwm scales

huttongrabiel:
Agreed on Saturday that left trigger should open the hand and right trigger should close.


abhaybd (approved):
lgtm

PR #231, opened 2023-04-21
rover-config: Add script to automate file copying + update readme

huttongrabiel:
Easier to run rover-config.sh then copy each and every file manually.

### DIFF
--- a/rover-config/rover-config.sh
+++ b/rover-config/rover-config.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 cp -r ./.motion/ /home/$USER
-cp ./50-rover-cameras.rules /etc/udev/rules.d/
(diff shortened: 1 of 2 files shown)


rover-config/rover-config.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+cp -r ./.motion/ /home/$USER
+cp ./50-rover-cameras.rules /etc/udev/rules.d/
+cp ./50-usb-hokuyo-lidar.rules /etc/udev/rules.d/
```
abhaybd, line 4: I think these need `sudo`

abhaybd (changes requested):
sudo

abhaybd (approved):
Looks good!

PR #234, opened 2023-05-05
Send e-stop packet upon receiving e-stop message from mc

huttongrabiel:
Send an estop packet upon clicking estop in mission control


abhaybd:
This breaks the world interface abstraction (we shouldn't call can methods from the MC protocol), so #236 is the right way to do it.

PR #237, opened 2023-05-05
Initial Limit Switch zero calibration test.

AyushKulk:
Added a Limit Switch Calibration executable and modified world interface initialization function along with CMake

### DIFF
--- a/src/LimitCalib.cpp
+++ b/src/LimitCalib.cpp
@@ -1,0 +1,6 @@
+#include "world_interface/world_interface.h"
+
+int main() { 
+    bool runCalibration = true;
+    robot::world_interface_init(runCalibration);
+}
--- a/src/LimitCalib.cpp
+++ b/src/LimitCalib.cpp
@@ -1,6 +1,6 @@
 #include "world_interface/world_interface.h"
 
 int main() {
-        bool runCalibration = true;
-        robot::world_interface_init(runCalibration);
+       bool runCalibration = true;
+       robot::world_interface_init(runCalibration);
 }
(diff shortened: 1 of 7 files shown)


src/LimitCalib.cpp
```diff
@@ -0,0 +1,6 @@
+#include "world_interface/world_interface.h"
+
+int main() {
+	bool runCalibration = true;
+	robot::world_interface_init(runCalibration);
```
abhaybd, line 5: As we discussed in the discord this should add a limit switch callback, run the motors until it hits a limit switch, and then in the callback stop the motor power and exit.

abhaybd (changes requested):
Some changes needed, most of which we already talked about

src/LimitCalib.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+	// limit switch callbacks.
+	for (const auto& entry : encMotors) {
+		robot::addLimitSwitchCallback(entry.first, callback);
+	}
+
+	// run motors to get to limit switch.
+	for (const auto& runningMotor : encMotors) {
+		setMotorPower(runningMotor.first, -0.2);
```
abhaybd, line 36: Some motors may need positive or negative zero calib powers, so we should add fields to `real_world_constants.h` that specify this.

abhaybd (changes requested):
Some changes needed as we discussed, including adding the `initOnlyMotors` to `noop_world`

abhaybd (approved):
Looks good! I had to do some bug fixes, but overall this is great!

PR #240, opened 2023-05-16
Added handler for inverse kinematics special case joint request

Geeoon:
Added support for mission control joint requests specifically for inverse kinematic joint requests.  Still need to call the actual methods.

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -3,9 +3,12 @@
 #include "kinematics/DiffDriveKinematics.h"
 #include "world_interface/data.h"
 
+#include <array>
 #include <chrono>
 #include <cmath>
 #include <string>
+
+#include <frozen/unordered_map.h>
 
 using namespace kinematics;
 
@@ -121,7 +124,36 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
 // Video encoding
 namespace video {
 constexpr int H264_RF_CONSTANT = 40;
-}
+} // namespace video
+
+// Arm inverse kinematics
+namespace arm {
+
+constexpr double IK_SOLVER_THRESH = 0.001;
+
+constexpr int IK_SOLVER_MAX_ITER = 50;
+
+/**
+ * The motors used in IK. The ordering in this array is the canonical ordering of these motors
+ * for IK purposes.
+ */
+constexpr std::array<robot::types::motorid_t, 2> IK_MOTORS = {
+	robot::types::motorid_t::shoulder, robot::types::motorid_t::elbow};
+
+/**
+ * Map from motor ids to min and max joint limits
+ */
+constexpr frozen::unordered_map<robot::types::motorid_t, std::pair<int, int>, IK_MOTORS.size()>
+	JOINT_LIMITS{{robot::types::motorid_t::shoulder, {18200, 152500}},
+				 {robot::types::motorid_t::elbow, {-169100, 0}}};
+
+/**
+ * Map from motor ids to segment length
+ */
+constexpr frozen::unordered_map<robot::types::motorid_t, double, IK_MOTORS.size()>
+	SEGMENT_LENGTHS{{robot::types::motorid_t::shoulder, 0.3848608},
+					{robot::types::motorid_t::elbow, 0.461264}};
+} // namespace arm
 
 constexpr double CONTROL_HZ = 10.0;
 } // namespace Constants
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -166,14 +166,13 @@ constexpr std::array<robot::types::jointid_t, 2> IK_MOTOR_JOINTS = {
  * The motors used in IK. The i-th element in this array corresponds to the joint in the i-th
  * element of `IK_MOTOR_JOINTS`
  */
-constexpr std::array<robot::types::motorid_t, 2> IK_MOTORS = []() constexpr {
+constexpr std::array<robot::types::motorid_t, 2> IK_MOTORS = ([]() constexpr {
 	std::array<robot::types::motorid_t, IK_MOTOR_JOINTS.size()> ret{};
 	for (size_t i = 0; i < IK_MOTOR_JOINTS.size(); i++) {
 		ret[i] = JOINT_MOTOR_MAP.at(IK_MOTOR_JOINTS[i]);
 	}
 	return ret;
-}
-();
+})();
 
 /**
  * Map from motor ids to min and max joint limits in millidegrees
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -97,6 +97,10 @@ constexpr const char* SIM_PROTOCOL_NAME = "/simulator";
    WebSocket server endpoint for the DGPS protocol.
  */
 constexpr const char* DGPS_PROTOCOL_NAME = "/dgps";
+/**
+   Websocket server endpoint for ArduPilot protocol
+ */
+constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
 
 constexpr std::chrono::milliseconds JOINT_POWER_REPEAT_PERIOD(333);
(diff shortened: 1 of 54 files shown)


src/world_interface/kinematic_common_interface.cpp
```diff
@@ -170,9 +170,23 @@ double getJointPowerValue(types::jointid_t joint) {
 
 void setJointMotorPower(robot::types::jointid_t joint, double power) {
 	using robot::types::jointid_t;
-
 	if (jointMotorMap.find(joint) != jointMotorMap.end()) {
-		setMotorPower(jointMotorMap.at(joint), power);
+		// TODO: check if associated motor is in IK_MOTORS, don't check specifically these ones
+		if (joint == jointid_t::elbow || joint == jointid_t::shoulder) {
```
abhaybd, line 175: As said in the todo, we should check against IK_MOTORS, not hardcode.

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -170,9 +170,23 @@ double getJointPowerValue(types::jointid_t joint) {
...
+			if (!Globals::armIKEnabled) {
+				setMotorPower(jointMotorMap.at(joint), power);
+			}
+		} else {
+			setMotorPower(jointMotorMap.at(joint), power);
+		}
```
abhaybd, line 181: This set of nested if-else can just be simplified to "if ik not enabled or this isn't an ik joint", i.e.
```c++
if (!Globals::armIKEnabled || !std::find(IK_MOTORS.begin(), IK_MOTORS.end(), joint)) {
    setMotorPower(...);
}
```
which is cleaner

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -170,9 +170,23 @@ double getJointPowerValue(types::jointid_t joint) {
...
+	} else if (joint == jointid_t::ikForward || joint == jointid_t::ikUp) {
+		if (Globals::armIKEnabled) {
+			if (joint == jointid_t::ikForward) {
+				Globals::planarArmController.set_x_vel(dataclock::now(), power);
+			} else {
+				Globals::planarArmController.set_y_vel(dataclock::now(), power);
```
abhaybd, line 187: I realized we should also be scaling these. Add a constant in `Constants.h` called `MAX_EE_VEL` or something that is max EE velocity in m/s. I think 0.3m/s or something is good enough, though we can tune that.

Then just scale power by this max vel, so something like:
```c++
Globals::planarArmController.set_x_vel(dataclock::now(), power * Constants::arm::MAX_EE_VEL);
```
and the same for the y vel too.

src/Constants.h
```diff
@@ -96,6 +99,7 @@ constexpr const char* SIM_PROTOCOL_NAME = "/simulator";
 constexpr const char* DGPS_PROTOCOL_NAME = "/dgps";
 
 constexpr std::chrono::milliseconds JOINT_POWER_REPEAT_PERIOD(333);
+constexpr std::chrono::milliseconds ARM_IK_UPDATE_PERIOD(100);
```
abhaybd, line 102: Let's make this 50ms instead? Should make for smoother control

src/Constants.h
```diff
@@ -121,7 +125,36 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
...
+
+/**
+ * Map from motor ids to min and max joint limits
+ */
+constexpr frozen::unordered_map<robot::types::motorid_t, std::pair<int, int>, IK_MOTORS.size()>
+	JOINT_LIMITS{{robot::types::motorid_t::shoulder, {18200, 152500}},
+				 {robot::types::motorid_t::elbow, {-169100, 0}}};
+
+/**
+ * Map from motor ids to segment length
```
abhaybd, line 152: I know I wrote this but can you add that the seg lens are in meters.

src/Constants.h
```diff
@@ -121,7 +125,36 @@ constexpr uint32_t RPLIDAR_S1_BAUDRATE = 256000;
...
+
+/**
+ * The motors used in IK. The ordering in this array is the canonical ordering of these motors
+ * for IK purposes.
+ */
+constexpr std::array<robot::types::motorid_t, 2> IK_MOTORS = {
+	robot::types::motorid_t::shoulder, robot::types::motorid_t::elbow};
+
+/**
+ * Map from motor ids to min and max joint limits
```
abhaybd, line 145: Same as below, can you specify that units are in milldegrees

abhaybd (changes requested):
Great job with this! I left some comments below on some needed changes before we can merge. Lmk if you have any questions

src/Constants.h
```diff
@@ -133,26 +134,46 @@ constexpr int H264_RF_CONSTANT = 40;
 // Arm inverse kinematics
 namespace arm {
 
+constexpr double MAX_EE_VEL = 0.3; // m/s
```
abhaybd, line 137: Instead of putting `// m/s` at the end, can we add proper doxy comments saying "Maximum commanded end-effector velocity, in m/s" or something

src/network/MissionControlProtocol.cpp
```diff
@@ -121,6 +127,33 @@ void MissionControlProtocol::handleDriveRequest(const json& j) {
 	this->setRequestedCmdVel(dtheta, dx);
 }
 
+void MissionControlProtocol::handleSetArmIKEnabled(const json& j) {
+	bool enabled = j["enabled"];
+	Globals::armIKEnabled = enabled;
```
abhaybd, line 132: In the `true` case we set it to false and then back to true again, so this line is extraneous. It would be cleaner to just set `Globals::armIKEnabled = false` in the false case (before the `is_joinable` call) and delete this line.

abhaybd (changes requested):
Looks great functionality wise, just some quick comments

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -162,12 +162,11 @@ double getJointPowerValue(types::jointid_t joint) {
...
 	using robot::types::jointid_t;
-	if (Constants::arm::JOINT_MOTOR_MAP.find(joint) != Constants::arm::JOINT_MOTOR_MAP.end()) {
-		// TODO: check if associated motor is in IK_MOTORS, don't check specifically these ones
-		if (!Globals::armIKEnabled ||
-			!std::find(Constants::arm::IK_MOTOR_JOINTS.begin(),
-					   Constants::arm::IK_MOTOR_JOINTS.end(), joint)) {
-			setMotorPower(Constants::arm::JOINT_MOTOR_MAP.at(joint), power);
+	if (Constants::JOINT_MOTOR_MAP.find(joint) != Constants::JOINT_MOTOR_MAP.end()) {
+		bool isIKMotor = std::find(Constants::arm::IK_MOTOR_JOINTS.begin(),
+								   Constants::arm::IK_MOTOR_JOINTS.end(), joint);
```
abhaybd, line 167: I definitely lied to you earlier whoops, `std::find` doesn't return a `bool`, it returns an iterator. You should check `bool isIKMotor = std::find(...) == IK_MOTORS_JOINTS.end()`.

abhaybd (changes requested):
I was wrong earlier about `std::find`, one fix is needed for that

src/world_interface/kinematic_common_interface.cpp
huttongrabiel: @abhaybd `bool isIKMotor = std::find(...) == IK_MOTORS_JOINTS.end()` will return `true` if it doesn't find a motor. I think you meant `bool isIKMotor = std::find(..) != IK_MOTORS_JOINTS.end()`

abhaybd: Whoops yeah you're right, @geeoon do that instead

abhaybd (approved):
looks bueno

PR #246, opened 2023-05-20
Improve camera init

abhaybd:
Minor refactoring to reduce code duplication

### DIFF
--- a/src/world_interface/real_world_interface.cpp
+++ b/src/world_interface/real_world_interface.cpp
@@ -95,33 +95,25 @@ void initMotors() {
 	}
 }
 
+void openCamera(CameraID camID, const char* cameraPath) {
+	try {
+		auto cam = std::make_shared<cam::Camera>();
+		bool success = cam->openFromConfigFile(cameraPath);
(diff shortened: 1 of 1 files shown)


src/world_interface/real_world_interface.cpp
```diff
@@ -113,34 +113,26 @@ void initMotors() {
...
+		}
 	} catch (const std::exception& e) {
 		log(LOG_ERROR, "Error opening camera with id %s:\n%s\n",
-			util::to_string(Constants::AR_CAMERA_ID).c_str(), e.what());
+			util::to_string(camID).c_str(), e.what());
 	}
 }
+
+void setupCameras() {
+	// openCamera(Constants::AR_CAMERA_ID, Constants::AR_CAMERA_CONFIG_PATH);
```
huttongrabiel, line 133: Non-blocking: Just want to make sure this isn't meant to be called. The original code uses AR_CAMERA_ID but this one only uses HAND_CAMERA_ID.

huttongrabiel (approved):

PR #249, opened 2023-05-20
ArduPilotProtocol: Create protocol for ardupilot messages

huttongrabiel:
* Add orientation_t to navtypes
* Handle pixhawk messages for GPS, IMU, and heading

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -98,7 +98,7 @@ constexpr const char* SIM_PROTOCOL_NAME = "/simulator";
  */
 constexpr const char* DGPS_PROTOCOL_NAME = "/dgps";
 /**
-   Websocket server endpoint for Pixhawk protocol
+   Websocket server endpoint for ArduPilot protocol
  */
 constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
 
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -97,6 +97,10 @@ constexpr const char* SIM_PROTOCOL_NAME = "/simulator";
    WebSocket server endpoint for the DGPS protocol.
  */
 constexpr const char* DGPS_PROTOCOL_NAME = "/dgps";
+/**
+   Websocket server endpoint for Pixhawk protocol
+ */
+constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
 
 constexpr std::chrono::milliseconds JOINT_POWER_REPEAT_PERIOD(333);
 
--- a/src/ardupilot/ArduPilotInterface.cpp
+++ b/src/ardupilot/ArduPilotInterface.cpp
@@ -14,7 +14,7 @@ std::unique_ptr<net::ardupilot::ArduPilotProtocol> ardupilot_protocol;
 
 namespace ardupilot {
 void initArduPilotProtocol(net::websocket::SingleClientWSServer& websocketServer) {
-	std::make_unique<net::ardupilot::ArduPilotProtocol>(websocketServer);
+	ardupilot_protocol = std::make_unique<net::ardupilot::ArduPilotProtocol>(websocketServer);
 }
 } // namespace ardupilot
 
@@ -28,7 +28,7 @@ namespace robot {
 DataPoint<double> readIMUHeading() {
 	auto imu_heading = ardupilot_protocol->getIMU();
 	if (imu_heading.isValid()) {
-		return imu_heading.getData().yaw;
+		return DataPoint<double>(imu_heading.getTime(), imu_heading.getData().yaw);
 	}
 	return DataPoint<double>();
 }
--- a/src/ardupilot/ArduPilotInterface.cpp
+++ b/src/ardupilot/ArduPilotInterface.cpp
@@ -1,21 +1,35 @@
-#include "ArduPilotProtocol.h"
+#include "ArduPilotInterface.h"
+
+#include <memory>
 
 using namespace robot::types;
 using namespace navtypes;
 using namespace std::chrono_literals;
 
 namespace {
-net::ardupilot::ArduPilotProtocol ardupilot_protocol;
+
+std::unique_ptr<net::ardupilot::ArduPilotProtocol> ardupilot_protocol;
+
 } // namespace
+
+namespace ardupilot {
+void initArduPilotProtocol(net::websocket::SingleClientWSServer& websocketServer) {
+	std::make_unique<net::ardupilot::ArduPilotProtocol>(websocketServer);
+}
+} // namespace ardupilot
 
 namespace gps {
 DataPoint<gpscoords_t> readGPSCoords() {
-	return ardupilot_protocol.getGPS();
+	return ardupilot_protocol->getGPS();
 }
 } // namespace gps
 
 namespace robot {
 DataPoint<double> readIMUHeading() {
-	return ardupilot_protocol.getIMU().getData().yaw;
+	auto imu_heading = ardupilot_protocol->getIMU();
+	if (imu_heading.isValid()) {
+		return imu_heading.getData().yaw;
+	}
+	return DataPoint<double>();
 }
 } // namespace robot
--- a/src/ardupilot/ArduPilotInterface.cpp
+++ b/src/ardupilot/ArduPilotInterface.cpp
@@ -10,13 +10,12 @@ net::ardupilot::ArduPilotProtocol ardupilot_protocol;
 
 namespace gps {
 DataPoint<gpscoords_t> readGPSCoords() {
-    return ardupilot_protocol.getGPS();
+	return ardupilot_protocol.getGPS();
 }
 } // namespace gps
 
 namespace robot {
 DataPoint<double> readIMUHeading() {
-    return ardupilot_protocol.getIMU().getData().yaw;
+	return ardupilot_protocol.getIMU().getData().yaw;
 }
 } // namespace robot
-
--- a/src/ardupilot/ArduPilotInterface.cpp
+++ b/src/ardupilot/ArduPilotInterface.cpp
@@ -1,0 +1,22 @@
+#include "ArduPilotProtocol.h"
+
+using namespace robot::types;
+using namespace navtypes;
+using namespace std::chrono_literals;
+
+namespace {
+net::ardupilot::ArduPilotProtocol ardupilot_protocol;
+} // namespace
+
+namespace gps {
+DataPoint<gpscoords_t> readGPSCoords() {
+    return ardupilot_protocol.getGPS();
+}
+} // namespace gps
+
+namespace robot {
+DataPoint<double> readIMUHeading() {
+    return ardupilot_protocol.getIMU().getData().yaw;
+}
+} // namespace robot
+
--- a/src/ardupilot/ArduPilotProtocol.cpp
+++ b/src/ardupilot/ArduPilotProtocol.cpp
@@ -16,9 +16,12 @@ using namespace robot::types;
 using namespace navtypes;
 using std::placeholders::_1;
 
-ArduPilotProtocol::ArduPilotProtocol() {}
+ArduPilotProtocol::ArduPilotProtocol() {
+	initArduPilotServer(Globals::websocketServer);
+}
 
-void ArduPilotProtocol::initArduPilotServer() {
+void ArduPilotProtocol::initArduPilotServer(
+	net::websocket::SingleClientWSServer& websocketServer) {
 	auto ardupilot_protocol = std::make_unique<net::websocket::WebSocketProtocol>(
 		Constants::ARDUPILOT_PROTOCOL_NAME);
 	ardupilot_protocol->addMessageHandler(
@@ -35,7 +38,7 @@ void ArduPilotProtocol::initArduPilotServer() {
 	ardupilot_protocol->addDisconnectionHandler(
 		std::bind(&ArduPilotProtocol::clientDisconnected, this));
 
-	Globals::websocketServer.addProtocol(std::move(ardupilot_protocol));
+	websocketServer.addProtocol(std::move(ardupilot_protocol));
 }
 
 void ArduPilotProtocol::clientConnected() {
@@ -44,7 +47,6 @@ void ArduPilotProtocol::clientConnected() {
 		std::lock_guard<std::mutex> lock(_connectionMutex);
 		_arduPilotProtocolConnected = true;
 	}
-	_connectionCV.notify_one();
 }
 
 void ArduPilotProtocol::clientDisconnected() {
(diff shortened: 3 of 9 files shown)


src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,119 @@
...
+	return _lastGPS;
+}
+
+DataPoint<orientation_t> ArduPilotProtocol::getIMU() {
+	return _lastOrientation;
+}
+
+DataPoint<int> ArduPilotProtocol::getHeading() {
+	return _lastHeading;
+}
```
abhaybd, line 116: This is a race condition, since assignments and copies aren't atomic operations. The naive solution here is to acquire the relevant mutex before returning, but it might be simpler and better to just make `lastGPS`, `lastOrientation`, and `lastHeading` into `std::atomic`, and then use `store()` and `load()` for atomic copies and assignments. That way we can get rid of the associated mutexes and make the code a bit cleaner.

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,96 @@
...
+	robot::types::DataPoint<int> getHeading();
+
+	/* Validates GPS request
+	 *
+	 * @param json message
+	 *
+	 * @return True if message is valid, false otherwise
+	 */
+	static bool validateGPSRequest(const json& j);
+	/* Validates IMU request
```
abhaybd, line 50: For this and all the methods below, can we add newlines after a method declaration?

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,96 @@
+#include "../network/websocket/WebSocketProtocol.h"
+#include "../network/websocket/WebSocketServer.h"
+#include "../world_interface/data.h"
+
+#include <condition_variable>
+#include <mutex>
+
+using json = nlohmann::json;
```
abhaybd, line 8: `using` declarations in header files is bad because it results in namespace pollution. In header files everything should be referred to by the fully-qualified name

src/navtypes.h
```diff
@@ -14,6 +14,12 @@ struct gpscoords_t {
 	double lon;
 };
 
+struct orientation_t {
+	double roll;
+	double pitch;
+	double yaw;
+};
```
abhaybd, line 21: In some cases we represent orientation as a quaternion instead. Since RPY isn't the only way of representation orientation, maybe this should be called `eulerangles_t` or something?

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,119 @@
...
+	ardupilot_protocol->addDisconnectionHandler(
+		std::bind(&ArduPilotProtocol::clientDisconnected, this));
+
+	Globals::websocketServer.addProtocol(std::move(ardupilot_protocol));
+
+	{
+		log(LOG_INFO, "Waiting for ArduPilot to connect...\n");
+		std::unique_lock<std::mutex> lock(_connectionMutex);
+		_connectionCV.wait(lock, [this] { return _arduPilotProtocolConnected; });
+	}
```
abhaybd, line 43: We probably don't want to block everything until the ardupilot connects. The ardupilot isn't strictly required for the rover to run, so we don't want to add that dependency.

When the ardupilot isn't connected we can just return empty datapoints.

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,119 @@
...
+	log(LOG_INFO, "ArduPilot connected.\n");
+	{
+		std::lock_guard<std::mutex> lock(_connectionMutex);
+		_arduPilotProtocolConnected = true;
+	}
+	_connectionCV.notify_one();
+}
+
+void ArduPilotProtocol::clientDisconnected() {
+	log(LOG_INFO, "ArduPilot disconnected.");
```
abhaybd, line 56: Should probably be a `LOG_WARN`, since we don't expect this during normal operation

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,119 @@
...
+		_lastOrientation = orientation_t{roll, pitch, yaw};
+	}
+}
+
+bool ArduPilotProtocol::validateHeadingRequest(const json& j) {
+	return util::validateKey(j, "heading", val_t::number_integer);
+}
+
+void ArduPilotProtocol::handleHeadingRequest(const json& j) {
+	int heading = j["heading"];
```
abhaybd, line 99: Why is `heading` an int? Is it just heading in integer degrees? Also, what's the difference between heading and yaw in this code

src/ardupilot/ArduPilotProtocol.cpp
abhaybd: Also, missing a `\n` at the end

abhaybd (changes requested):
Mostly good! I left some comments below about some needed changes. Lmk if you have questions! Another required change is to update the `CMakeLists.txt` to add the new code to a build target.

Also we should discuss where to implement the world interface methods, let's do that today.

src/ardupilot/ArduPilotProtocol.cpp
huttongrabiel: ```heading``` is just an int between 0..359 with 0 being north. Heading is degrees, yaw is radians.

src/navtypes.h
```diff
@@ -14,6 +14,12 @@ struct gpscoords_t {
 	double lon;
 };
 
+struct eulerangles_t {
+	double roll;
+	double pitch;
+	double yaw;
+};
```
abhaybd, line 21: Even though this is self-explanatory, can we put doxy comments here? Could be useful to link here https://en.wikipedia.org/wiki/Euler_angles#Tait%E2%80%93Bryan_angles and mention that the rotation ordering is XYZ extrinsic rotations.

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,101 @@
...
+	~ArduPilotProtocol();
+	ArduPilotProtocol(const ArduPilotProtocol& other) = delete;
+	ArduPilotProtocol& operator=(const ArduPilotProtocol& other) = delete;
+
+	// Registers the protocol with the websocket server
+	void initArduPilotServer();
+	// Connection handler for websocket server
+	void clientConnected();
+	// Disconnect handler for websocket server
+	void clientDisconnected();
```
abhaybd, line 23: Do these methods need to be public? Client code probably doesn't need to be able to call these methods

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,101 @@
...
+
+class ArduPilotProtocol {
+public:
+	ArduPilotProtocol();
+	~ArduPilotProtocol();
+	ArduPilotProtocol(const ArduPilotProtocol& other) = delete;
+	ArduPilotProtocol& operator=(const ArduPilotProtocol& other) = delete;
+
+	// Registers the protocol with the websocket server
+	void initArduPilotServer();
```
abhaybd, line 19: Can you convert the `//` comment into proper doxy comments?

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,111 @@
...
+		validateIMURequest);
+	ardupilot_protocol->addMessageHandler(
+		"arduPilotHeadingReport",
+		std::bind(&ArduPilotProtocol::handleHeadingRequest, this, _1), validateHeadingRequest);
+	ardupilot_protocol->addConnectionHandler(
+		std::bind(&ArduPilotProtocol::clientConnected, this));
+	ardupilot_protocol->addDisconnectionHandler(
+		std::bind(&ArduPilotProtocol::clientDisconnected, this));
+
+	Globals::websocketServer.addProtocol(std::move(ardupilot_protocol));
```
abhaybd, line 38: Instead of hardcoding it to use the global websocket server, could we just take a reference to the server as a parameter of this method?

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,101 @@
...
+#include "../network/websocket/WebSocketServer.h"
+#include "../world_interface/data.h"
+
+#include <condition_variable>
+#include <mutex>
+
+namespace net {
+namespace ardupilot {
+
+class ArduPilotProtocol {
```
abhaybd, line 11: Can you add class documentation?

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void ArduPilotProtocol::clientConnected() {
+	log(LOG_INFO, "ArduPilot connected.\n");
+	{
+		std::lock_guard<std::mutex> lock(_connectionMutex);
+		_arduPilotProtocolConnected = true;
+	}
+	_connectionCV.notify_one();
```
abhaybd, line 47: Looks like we're notifying this cv but not waiting on it anywhere. Should we just remove it?

Keeping track of the connection state is a good idea though, can we add a getter `isArduPilotConnected()` or something?

src/Constants.h
```diff
@@ -94,6 +94,10 @@ constexpr const char* SIM_PROTOCOL_NAME = "/simulator";
    WebSocket server endpoint for the DGPS protocol.
  */
 constexpr const char* DGPS_PROTOCOL_NAME = "/dgps";
+/**
+   Websocket server endpoint for Pixhawk protocol
```
abhaybd, line 98: Pixhawk -> ArduPilot?

src/ardupilot/ArduPilotProtocol.h
abhaybd, line 19: This looks like UB if this method is called more than once. Is it cleaner to do this in the constructor instead, so we can guarantee that it's only called once?

abhaybd (changes requested):
Overall pretty good! Just some more fixes needed. Also, remember to add ArduPilot.cpp to the CMakeLists. I'd say create a new build target called `ardupilot` and link that as needed.

Also, I'd say create a new file called like `ardupilot_interface.cpp` or something and implement `readGPSCoords()` and `readIMUHeading()` there. That file can instantiate a `ArduPilotProtocol` as a global (in an anonymous namespace) and just call the appropriate getters in the world interface implementations. Remember to add an init function that initializes the protocol too.

src/ardupilot/ArduPilotProtocol.h
huttongrabiel: ya this works. its called in the constructor and initArduPilotServer is a private function now

huttongrabiel:
Let me know if this is what you were thinking with the ArduPilotInterface.

src/ardupilot/ArduPilotInterface.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace gps {
+DataPoint<gpscoords_t> readGPSCoords() {
+	return ardupilot_protocol.getGPS();
+}
+} // namespace gps
+
+namespace robot {
+DataPoint<double> readIMUHeading() {
+	return ardupilot_protocol.getIMU().getData().yaw;
```
abhaybd, line 19: We should check if the datapoint is valid before calling `getData()`, this will fail otherwise

src/ardupilot/ArduPilotProtocol.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+using json = nlohmann::json;
+using val_t = json::value_t;
+using namespace robot::types;
+using namespace navtypes;
+using std::placeholders::_1;
+using namespace net::websocket;
+
+ArduPilotProtocol::ArduPilotProtocol() {
+	initArduPilotServer(Globals::websocketServer);
```
abhaybd, line 21: Can we take the websocket server as a parameter? It doesn't need to be hardcoded here

src/ardupilot/ArduPilotProtocol.h
```diff
@@ -0,0 +1,111 @@
...
+	 */
+	robot::types::DataPoint<int> getHeading();
+
+	/* @brief Validates GPS request
+	 *
+	 * @param json message
+	 *
+	 * @return True if message is valid, false otherwise
+	 */
+	static bool validateGPSRequest(const nlohmann::json& j);
```
abhaybd, line 42: Should all these validate and handle methods be public? Seems like we can make em private

abhaybd (changes requested):
Looks better! Just some more changes needed.

As we discussed, add an init method to `ArduPilotInterface`. Also, make sure to link the real world interface against the ardupilot build target.

src/ardupilot/ArduPilotInterface.cpp
huttongrabiel: What do we want to return in the case that it is not valid?

abhaybd (changes requested):
Just a few more changes needed. Take a look at these comments, first. Then check the discord, as mentioned there #252 needs to be merged first. Then you'll have to redo some of the CMakeLists code, it'll be clearer how to do that when you look at the other pr.

Lmk if you have questions!

src/world_interface/real_world_interface.cpp
```diff
@@ -138,12 +139,12 @@ void setupCameras() {
 void world_interface_init(bool initOnlyMotors) {
 	if (!initOnlyMotors) {
 		setupCameras();
-		bool gps_success = gps::usb::startGPSThread();
 		bool lidar_success = lidar::initializeLidar();
 		bool landmark_success = AR::initializeLandmarkDetection();
 	}
 	can::initCAN();
 	initMotors();
+	ardupilot::initArduPilotProtocol(Globals::websocketServer);
```
abhaybd, line 147: Can you move this into the `if`? Outside the if should just be motor initialization stuff

src/world_interface/real_world_interface.cpp
```diff
@@ -242,10 +243,6 @@ DataPoint<points_t> readLidarScan() {
 	return lidar::readLidar();
 }
 
-DataPoint<double> readIMUHeading() {
-	return {}; // TODO: interface with IMU
-}
```
abhaybd, line 247: This doesn't have to be addressed in this PR but I'm just leaving this as a todo. This will probably break the GPS=USB build, since there won't be a `readIMUHeading()` method. We should add a dummy imu method, the same way we have a dummy gps method.

This isn't relevant right now though, so let's address that later.

abhaybd (changes requested):
One very minor fix needed now. Other than that, this is good! Let's try to test on the robot this week and merge it asap.

abhaybd (commented):
Looks good! Let's get some hardware testing done and then it's good to merge.

abhaybd (changes requested):
Looks good, one minor thing to fix. As we tested yesterday, the heading thing wasn't working for some reason, which we should investigate. Given that we don't have much time and it's not used anywhere though, let's just disable that packet from the python side for now.

abhaybd (approved):
yeehaw

PR #253, opened 2023-05-30
Make arm controller threadsafe

abhaybd:
We'll be calling this bad boy from multiple threads so let's make it thread safe


huttongrabiel (changes requested):

huttongrabiel (approved):
lgtm

PR #257, opened 2023-06-08
Cleanup for joint code

abhaybd:
This PR cleans up some joint-related code.
* It removes the drill arm joint, since we don't do anything with it
* It factors out some common code for getting the position of multiple joints at once

### DIFF
--- a/src/world_interface/world_interface.h
+++ b/src/world_interface/world_interface.h
@@ -294,7 +294,6 @@ types::DataPoint<navtypes::Vectord<N>>
 getMotorPositionsRad(const std::array<types::motorid_t, N>& motors) {
 	navtypes::Vectord<N> motorPositions;
 	std::optional<types::datatime_t> timestamp;
-	bool valid = true;
 	for (size_t i = 0; i < motors.size(); i++) {
 		types::DataPoint<int32_t> positionDataPoint = robot::getMotorPos(motors[i]);
 		if (positionDataPoint.isValid()) {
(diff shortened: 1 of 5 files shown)


src/world_interface/world_interface.h
```diff
@@ -276,6 +277,45 @@ void setJointPos(robot::types::jointid_t joint, int32_t targetPos);
...
+		} else {
+			valid = false;
+			break;
+		}
+	}
+	if (valid) {
+		return types::DataPoint<navtypes::Vectord<N>>(timestamp.value(), motorPositions);
+	} else {
+		return {};
+	}
```
huttongrabiel, line 316: Ignore if you want but there is not a need to explicitly track valid with a flag and then have to spend time evaluating the flag before exiting the function. You could just do:

```cpp
	for (size_t i = 0; i < motors.size(); i++) {
		types::DataPoint<int32_t> positionDataPoint = robot::getMotorPos(motors[i]);
		if (positionDataPoint.isValid()) {
			if (!timestamp.has_value() || timestamp > positionDataPoint.getTime()) {
				timestamp = positionDataPoint.getTime();
			}
			double position = static_cast<double>(positionDataPoint.getData());
			position *= M_PI / 180.0 / 1000.0;
			motorPositions(i) = position;
		} else {
			return {};
		}
	}
	return types::DataPoint<navtypes::Vectord<N>>(timestamp.value(), motorPositions);
```

huttongrabiel (changes requested):

src/world_interface/world_interface.h
abhaybd: Oh yeah good catch, I'll do that

huttongrabiel (approved):

PR #264, opened 2023-08-29
Report full attitude, fix lat/lon bug, report attitude even if no gps fix

abhaybd:
This PR adds `readIMU()` to the world interface, which allows reading the full attitude. It also includes some refactoring, since `readIMUHeading()` is now obviously a subset of `readIMU()`.

Also includes a bugfix where `lat` was repeated twice instead of `lat` and `lon`.

### DIFF
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -4,6 +4,7 @@
 #include "../Globals.h"
 #include "../base64/base64_img.h"
 #include "../log.h"
+#include "../utils/core.h"
 #include "../utils/json.h"
 #include "../world_interface/data.h"
 #include "../world_interface/world_interface.h"
@@ -187,7 +188,8 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
 void MissionControlProtocol::sendRoverPos() {
 	auto imu = robot::readIMU();
 	auto gps = gps::readGPSCoords();
-	log(LOG_DEBUG, "imu=%d, gps=%d\n", imu.isValid(), gps.isValid());
+	log(LOG_DEBUG, "imu_valid=%d, gps_valid=%d\n", util::to_string(imu.isValid()),
+		util::to_string(gps.isValid()));
 	if (imu.isValid()) {
 		auto rpy = imu.getData();
 		Eigen::Quaterniond quat(Eigen::AngleAxisd(rpy.roll, Eigen::Vector3d::UnitX()) *
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -185,24 +185,31 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
 }
 
 void MissionControlProtocol::sendRoverPos() {
-	auto heading = robot::readIMUHeading();
-	auto gps = robot::readGPS();
-	if (gps.isValid() && heading.isValid()) {
-		double orientX = 0.0;
-		double orientY = 0.0;
-		double orientZ = std::sin(heading.getData() / 2);
-		double orientW = std::cos(heading.getData() / 2);
-		double posX = gps.getData()[0];
-		double posY = gps.getData()[1];
+	auto imu = robot::readIMU();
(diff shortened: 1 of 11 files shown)


src/world_interface/simulator_interface.cpp
```diff
@@ -364,9 +368,9 @@ DataPoint<pose_t> readVisualOdomVel() {
 	return DataPoint<pose_t>{};
 }
 
-DataPoint<double> readIMUHeading() {
-	std::lock_guard<std::mutex> guard(headingMutex);
-	return lastHeading;
+types::DataPoint<navtypes::eulerangles_t> readIMU() {
```
huttongrabiel, line 371: Same as above. Only threw me off because the same file has `DataPoint` for `lastAttitude` that does not compile if you change it to `types::DataPoint` because it's not in the `robot` namespace.

FYI these are nitpicky and you can ignore if you want. They just throw off the consistency and requires reading more context around the code to see where you need to qualify and not qualify.

src/world_interface/noop_world.cpp
```diff
@@ -43,7 +43,7 @@ landmarks_t readLandmarks() {
 	return landmarks_t{};
 }
 
-DataPoint<double> readIMUHeading() {
+types::DataPoint<navtypes::eulerangles_t> readIMU() {
```
huttongrabiel, line 46: Same as other comments

src/network/MissionControlProtocol.cpp
```diff
@@ -185,24 +185,31 @@ void MissionControlProtocol::handleJointPowerRequest(const json& j) {
...
-	if (gps.isValid() && heading.isValid()) {
-		double orientX = 0.0;
-		double orientY = 0.0;
-		double orientZ = std::sin(heading.getData() / 2);
-		double orientW = std::cos(heading.getData() / 2);
-		double posX = gps.getData()[0];
-		double posY = gps.getData()[1];
+	auto imu = robot::readIMU();
+	auto gps = gps::readGPSCoords();
+	log(LOG_DEBUG, "imu=%d, gps=%d\n", imu.isValid(), gps.isValid());
```
huttongrabiel, line 190: It might be better to put `imu_valid` and `gps_valid` instead of just `imu` and `gps` so that you know what its indicating. 

You can also do ```log(LOG_DEBUG, "imu_valid=%s, gps=%s\n", imu.isValid() ? "true" : "false", gps.isValid() ? "true" : "false")``` which will give you true and false instead of 1 and 0.

So in total you  would get `imu_valid=true, gps_valid=true`. Which is a bit more clear than `imu=1, gps=1`

huttongrabiel (changes requested):

huttongrabiel:
Core code looks good! Everything that stuck out to me was just nitpicky things that you can choose to fix or not. I'm happy to approve as is.

src/utils/core.h
```diff
@@ -44,10 +44,20 @@ std::unordered_set<K> keySet(const std::unordered_map<K, V>& input) {
...
+template <typename T>
+std::string to_string(const T& val) {
 	return std::to_string(val);
 }
 
+/**
+ * @brief Converts a boolean to a string.
+ *
+ * @param val The value to get the string representation of.
+ * @return "true" iff val, otherwise "false".
```
huttongrabiel, line 56: typo: `iff` -> `if`

huttongrabiel (changes requested):

huttongrabiel:
Looks good. One minor typo in comment

src/utils/core.h
abhaybd: Not a typo, [iff](https://en.wikipedia.org/wiki/If_and_only_if) is short for "if and only if", which is pretty standard notation since it's taught in the early 300-level CSE courses. It's a mathematically stronger statement than "if", so imo it's preferrable when applicable

huttongrabiel: Good to know.

huttongrabiel (approved):

abhaybd:
I'll hold off on merging this until we just verify it works fine on the robot.

abhaybd:
Confirm that this works in sim!

PR #267, opened 2023-09-29
Add planar arm safety factor

DrDab:
Add overall length percentage bound to planar arm joints to prevent singularity lockup.


src/control/PlanarArmController.h
```diff
@@ -64,12 +69,12 @@ template <unsigned int N> class PlanarArmController {
 		}
 
 		// bounds check (new pos + vel vector <= sum of joint lengths)
-		double radius = kin.getSegLens().sum();
+		double radius = kin.getSegLens().sum() * safetyFactor;
 		if (pos.norm() > radius) {
 			// new position is outside of bounds
-			// TODO: will need to eventually shrink velocity vector until it is within radius
-			// instead of just normalizing it
+			// shrink velocity vector until it is within radius.
```
abhaybd, line 75: We should leave in the TODO comment since that's still an open task

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -9,5 +11,20 @@ using namespace std::chrono_literals;
...
+	navtypes::Vectord<2> minAngles({-M_PI, -M_PI});
+	navtypes::Vectord<2> maxAngles({M_PI, M_PI});
+	kinematics::PlanarArmKinematics<2> kin_obj(segLens, minAngles, maxAngles, 0.0, 0);
+
+	PlanarArmController<2> foo({0, 0}, kin_obj, Constants::arm::SAFETY_FACTOR);
+	foo.set_setpoint({10.0, 0.0});
+	REQUIRE(foo.get_setpoint(robot::types::dataclock::now()).norm() == 9.5);
+
+	foo.set_setpoint({2.0, 3.0});
+	REQUIRE(fabs(foo.get_setpoint(robot::types::dataclock::now()).norm() - 2.1166861277) <
```
abhaybd, line 28: Very minor, but we should prefer `std::abs()` over the C functions like `fabs()`

abhaybd, line 28: 2.1166861277 is a magic number, instead let's put the calculation instead.

abhaybd, line 28: 2.1166861277 is a magic number, instead let's put the calculation instead.

abhaybd, line 28: Very minor, but we should prefer `std::abs()` over the C functions like `fabs()`

src/control/PlanarArmController.h
abhaybd, line 75: We should leave in the TODO comment since that's still an open task

abhaybd (changes requested):
Mostly looks good, but one thing I think we missed in our discussion. The set point should always be within the allowed region, so we should perform the same check (check if outside the allowed radius and if so then correct it) in `set_setpoint()` as well. Since we'll be doing the same check twice we should factor out that logic into a method called `adjust_setpoint()` or something. Lmk if you have any questions!

src/control/PlanarArmController.h
```diff
@@ -42,7 +47,27 @@ template <unsigned int N> class PlanarArmController {
...
+	}
+
+	/**
+	 * @brief Normalize the input vector to have a set radius,
+	 *  	  while maintaining the same direction it did before if it exceeds that set radius.
+	 *
+	 * @param input The input vector to normalize.
+	 * @param radius The radius to normalize the vector to.
+	 */
+	Eigen::Vector2d normalizeVectorWithinRadius(Eigen::Vector2d input, double radius) {
```
abhaybd, line 61: This should be a private method

abhaybd, line 61: Also we should rename `input` to something more useful, like `eePos` or something, and rename `normalizeVectorWithinRadius` to something more indicative as well

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -9,5 +11,23 @@ using namespace std::chrono_literals;
...
+	PlanarArmController<2> foo({0, 0}, kin_obj, Constants::arm::SAFETY_FACTOR);
+	foo.set_setpoint({10.0, 0.0});
+	REQUIRE(foo.get_setpoint(robot::types::dataclock::now()).norm() == 9.5);
+
+	foo.set_setpoint({2.0, 3.0});
+	Eigen::Vector2d origSetPoint = kin_obj.jointPosToEEPos({2.0, 3.0});
+	double expectedNorm = origSetPoint.norm();
+
+	REQUIRE(std::abs(foo.get_setpoint(robot::types::dataclock::now()).norm() - expectedNorm) <
+			1e-6);
```
abhaybd, line 32: Instead of checking that the norms are equal, can we check for vector equality? Some of the other tests do the same thing, so you can copy-paste that code

abhaybd (changes requested):
Mostly good, just a few more edits needed

src/control/PlanarArmController.h
```diff
@@ -42,7 +47,8 @@ template <unsigned int N> class PlanarArmController {
 	void set_setpoint(const navtypes::Vectord<N>& targetJointPos) {
 		Eigen::Vector2d newSetPoint = kin.jointPosToEEPos(targetJointPos);
 		std::lock_guard<std::mutex> lock(mutex);
-		setpoint = newSetPoint;
+		double radius = kin.getSegLens().sum() * safetyFactor;
```
abhaybd, line 50: This line is duplicated in both usages of `normalizeEEWithinRadius()`, can we just put it in that function?

abhaybd (changes requested):
Looks good, just one last thing

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -1,13 +1,60 @@
...
+	if (thetaDiff > PI) {
+		thetaDiff -= 2 * PI;
+	}
+	bool angleEqual = abs(thetaDiff) <= angle;
+	if (distEqual && angleEqual) {
+		SUCCEED();
+	} else {
+		std::cout << "Expected: " << toString(p1) << ", Actual: " << toString(p2) << std::endl;
+		FAIL();
+	}
```
abhaybd, line 36: Instead of this, we should use `INFO` along with `REQUIRE`, which has all the functionality you need, and works simpler. So this could be replaced with:

```c++
std::stringstream ss;
ss << "Expected: " << toString(p1) << ", Actual: " << toString(p2);
INFO(ss.str());
REQUIRE(distEqual && angleEqual);
```

More info is [here](https://github.com/catchorg/Catch2/blob/devel/docs/logging.md), but note that the message given to `INFO` is only printed if a test fails, which is what we want.

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -1,13 +1,60 @@
...
+void assertApprox(const Eigen::Vector2d& p1, const Eigen::Vector2d& p2, double dist = 1e-5,
+				  double angle = 1e-5) {
+	Eigen::Vector2d diff = p1 - p2;
+	bool distEqual = diff.norm() <= dist;
+	double thetaDiff = std::fmod(abs(diff(1)), 2 * PI);
+	// change domain from [0, 2pi) to (-pi, pi]
+	if (thetaDiff > PI) {
+		thetaDiff -= 2 * PI;
+	}
+	bool angleEqual = abs(thetaDiff) <= angle;
```
abhaybd, line 30: I like that you're thinking about it, but I don't think checking the angle is necessary. At best it's redundant, and at worst it could cause false failures, since near the zero vector angle changes by quite a lot even with small absolute changes.

I think just asserting the norm is less than dist is good enough.

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -1,13 +1,60 @@
...
+
+TEST_CASE("Test Planar Arm Safety Factor", "[control][planararmcontroller]") {
+	navtypes::Vectord<2> segLens({4.0, 6.0});
+	navtypes::Vectord<2> minAngles({-M_PI, -M_PI});
+	navtypes::Vectord<2> maxAngles({M_PI, M_PI});
+	kinematics::PlanarArmKinematics<2> kin_obj(segLens, minAngles, maxAngles, 0.0, 0);
+
+	PlanarArmController<2> foo({0, 0}, kin_obj, Constants::arm::SAFETY_FACTOR);
+	foo.set_setpoint({10.0, 0.0});
+	REQUIRE(foo.get_setpoint(robot::types::dataclock::now()).norm() == 9.5);
```
abhaybd, line 53: Instead of just checking that the norm is 9.5, lets check that the whole vector equals (9.5,0), just to be more explicit.

abhaybd (changes requested):
Ok this should be the last thing! After fixing this let's get this tested and then merged

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -1,13 +1,61 @@
...
+	Eigen::Vector2d diff = p1 - p2;
+	bool distEqual = diff.norm() <= dist;
+
+	if (!distEqual) {
+		std::stringstream ss;
+		ss << "Expected: " << toString(p1) << ", Actual: " << toString(p2);
+		INFO(ss.str());
+	}
+
+	REQUIRE(distEqual);
```
abhaybd, line 32: Catch2 prints out the evaluated expression in the `REQUIRE`, so we'll get better logs if we do `REQUIRE(diff.norm() <= dist)`

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -1,13 +1,61 @@
...
+
+#include <iostream>
+
 #include <catch2/catch.hpp>
 
 using namespace Catch::literals;
 using namespace control;
 using namespace std::chrono_literals;
 
+constexpr double PI = M_PI;
```
abhaybd, line 13: Do we use `PI` anywhere? Seems like it's just `M_PI` that's used

abhaybd (changes requested):
Good comments and documentation! Realized we missed a minor logic bug in the constructor, and then some minor notes in the testcase

abhaybd (approved):
Looks good! Nice job on this

PR #268, opened 2023-10-03
implemented arm constraint formula

sarahfredrickson:
The formatting for the "a" value looks weird but that's what the formatter told me to do. Also will probably have git issues bc I pulled changes to origin before pushing but still says I'm 4 commits behind master.


src/control/PlanarArmController.h
```diff
@@ -65,13 +66,20 @@ template <unsigned int N> class PlanarArmController {
...
+			// new position is outside of bounds. Shrink the velocity vector until it is within
+			// radius setpoint = setpoint inside circle pos = new setpoint outside circle
+			double diffDotProd = (pos - setpoint).dot(setpoint);
+			double differenceNorm = (pos - setpoint).squaredNorm();
+			double a =
+				-1 * diffDotProd +
+				std::sqrt(std::pow(diffDotProd, 2) -
+						  differenceNorm * (setpoint.squaredNorm() - std::pow(radius, 2))) /
+					differenceNorm;
+			constrainedPos = (1 - a) * setpoint + a * pos;
```
abhaybd, line 80: Some minor math cleanup
- Can we replace `-1 * blah` with `-blah`
- Let's split up the definition of `a` into a few different declaration. For example we can take out everything inside the sqrt and call it `discriminant`, stuff like that

src/control/PlanarArmController.h
```diff
@@ -65,13 +66,20 @@ template <unsigned int N> class PlanarArmController {
 
 		// bounds check (new pos + vel vector <= sum of joint lengths)
 		double radius = kin.getSegLens().sum();
+		Eigen::Vector2d constrainedPos = pos;
```
abhaybd, line 69: Don't think there's any point in having `constrainedPos`, let's delete this and just modify `pos` inside the if statement.

abhaybd (changes requested):
overall looks good, just some minor stuff

src/control/PlanarArmController.h
```diff
@@ -66,10 +67,15 @@ template <unsigned int N> class PlanarArmController {
 		// bounds check (new pos + vel vector <= sum of joint lengths)
 		double radius = kin.getSegLens().sum();
 		if (pos.norm() > radius) {
-			// new position is outside of bounds
-			// TODO: will need to eventually shrink velocity vector until it is within radius
-			// instead of just normalizing it
-			pos.normalize();
+			// new position is outside of bounds. Shrink the velocity vector until it is within
+			// radius setpoint = setpoint inside circle pos = new setpoint outside circle
```
abhaybd, line 71: Like we talked about, we should briefly explain this math

abhaybd (changes requested):
Looks good, we just need some documentation. 2 things:

1. Victor will merge his PR soon, so be prepared to resolve those merge conflicts. Basically just move your code into the `normalizeEEPosWithinRadius` method (I forget the exact name)
2. You'll need to edit Victor's testcase for our updated logic. Basically it should just be changing the constants in the testcase

src/control/PlanarArmController.h
```diff
@@ -149,13 +150,19 @@ class PlanarArmController {
...
+			// new position is outside of bounds. Set new EE setpoint so it will follow the velocity
+			// vector instead of drifting along the radius.
+			// setpoint = setpoint inside circle
+			// eePos = new setpoint outside circle
+			double diffDotProd = (eePos - setpoint).dot(setpoint);
+			double differenceNorm = (eePos - setpoint).squaredNorm();
+			double discriminant =
+				std::pow(diffDotProd, 2) -
+				differenceNorm * (setpoint.squaredNorm() - std::pow(radius, 2));
+			double a = -diffDotProd + std::sqrt(discriminant) / differenceNorm;
```
abhaybd, line 163: Remember order of operations, the numerator should have parenthesis. It should be `(-diffDotProd + sqrt(...)) / differenceNorm`

src/control/PlanarArmController.h
```diff
@@ -149,13 +150,19 @@ class PlanarArmController {
...
-			// TODO: will need to eventually shrink velocity vector until it is within radius
-			// instead of just normalizing it.
-
-			eePos.normalize();
-			eePos *= radius;
+			double radius = kin.getSegLens().sum();
+			// new position is outside of bounds. Set new EE setpoint so it will follow the velocity
+			// vector instead of drifting along the radius.
+			// setpoint = setpoint inside circle
+			// eePos = new setpoint outside circle
```
abhaybd, line 157: I think we need a bit more math explanation. I'd say just state the equation we solve after these definitions, so at least readers can re-derive the solution if they want.

abhaybd (changes requested):
Took a look, I understand what the bug you were bumping into was. First, the math had a minor typo, so some parenthesis were missing.

Secondly, it wasn't a bug that you introduced necessarily. Your code was working properly, and normalizing the length to 9.5, but due to floating point rounding errors it was *just* over 9.5. So when `get_setpoint()` was called, it tries to normalize it again, but `eePos` is the same as `setpoint`, so the numerator becomes 0, so the whole thing becomes `nan`.

I'd say we could fix this by just adding a very small epsilon to the radius when we're checking if we should normalize. So, in `normalizeEEWithinRadius()`, change `if (eePos.norm() > radius)` to `if (eePos.norm() > radius + 1e-4)`, so it will only normalize if we exceed the radius by any perceptible amount.

src/control/PlanarArmController.h
```diff
@@ -148,14 +149,20 @@ class PlanarArmController {
...
 		double radius = kin.getSegLens().sum() * safetyFactor;
 
-		if (eePos.norm() > radius) {
-			// TODO: will need to eventually shrink velocity vector until it is within radius
-			// instead of just normalizing it.
-
-			eePos.normalize();
-			eePos *= radius;
+		if (eePos.norm() > (radius + 0.5)) {
+			double radius = kin.getSegLens().sum();
```
abhaybd, line 153: This line declares `radius` to an incorrect value (we need it to be multiplied by the safety factor). Just remove this line and it should be fixed. Not you'll have to change the expected value in the test case.

abhaybd (changes requested):

src/control/PlanarArmController.h
abhaybd: I think this explanation could still use some work. Like I mentioned, put the original equation we started with, which will allow others to rederive the solution if necessary.

abhaybd (changes requested):
Functionality-wise looks great! Just some clean-up needed on comments

abhaybd:
Great job, this looks good! The code looks fine, let's do some hardware testing before merge. Firmware should get ironed out very soon, which will enable this testing. Let's look into sim testing as well

src/control/PlanarArmController.h
```diff
@@ -41,7 +41,7 @@ class PlanarArmController {
 		// NOTE: currJointPos could extend beyond the safetyFactor, so safety factor
 		//       normalization logic is performed.
 		set_setpoint(currJointPos);
-		CHECK_F(safetyFactor > 0.0 && safetyFactor < 1.0);
+		assert(safetyFactor > 0.0 && safetyFactor < 1.0);
```
abhaybd, line 44: Why the change from `CHECK_F` to `assert`? In general we should usually prefer `CHECK_F`, since loguru has some nicer features, and still runs the checks in release mode.

sarahfredrickson:
Setpoint no longer goes out of bounds and arm doesn't crumple on itself

src/control/PlanarArmController.h
```diff
@@ -85,14 +86,18 @@ class PlanarArmController {
 	 * @return The new command, which is the new joint positions.
 	 */
 	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+		Eigen::Vector2d pos;
 		std::lock_guard<std::mutex> lock(mutex);
+		pos = setpoint;
 		if (velTimestamp.has_value()) {
 			double dt = util::durationToSec(currTime - velTimestamp.value());
-			setpoint += velocity * dt;
+			pos += velocity * dt;
```
abhaybd, line 94: If the `velTimestamp` doesn't have a value, there's no point in setting `setpoint` again, since we're not changing it. Can you change this to just `setpoint = normalizeEEWithinRadius(setpoint + velocity * dt)`?

After that you can remove the `pos` variable

src/control/PlanarArmController.h
```diff
@@ -103,14 +108,18 @@ class PlanarArmController {
 	 * @return The new command, which is the new joint positions.
 	 */
 	void set_y_vel(robot::types::datatime_t currTime, double targetVel) {
+		Eigen::Vector2d pos;
 		std::lock_guard<std::mutex> lock(mutex);
+		pos = setpoint;
 		if (velTimestamp.has_value()) {
 			double dt = util::durationToSec(currTime - velTimestamp.value());
-			setpoint += velocity * dt;
+			pos += velocity * dt;
```
abhaybd, line 116: Same comment as above

abhaybd (changes requested):
Super minor nitpicks

Geeoon (approved):

abhaybd (approved):
Looks good! Nice job on this, especially finding and fixing that bug

PR #270, opened 2023-10-10
Arm Inverse Kinematics Updates

Geeoon:
Changes setArmIKEnabled packet to requestArmIKEnabled.
Adds reportArmIKEnabled packet
No longer crashes if enabling arm IK doesn't succeed. Instead it sends a new packet, reportArmIKEnabled which will let MC know if arm IK is enabled or not.

### DIFF
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -86,10 +86,7 @@ void MissionControlProtocol::handleEmergencyStopRequest(const json& j) {
 	}
 	// TODO: reinit motors
 	Globals::E_STOP = stop;
-	Globals::armIKEnabled = false;
-
-	// inform MC that Ik has been disabled
-	sendArmIKEnabledReport();
+	this->setArmIKEnabled();
 }
 
 static bool validateOperationModeRequest(const json& j) {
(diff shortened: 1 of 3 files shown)


src/network/MissionControlProtocol.cpp
```diff
@@ -137,20 +138,20 @@ void MissionControlProtocol::handleSetArmIKEnabled(const json& j) {
...
-		Globals::armIKEnabled = true;
-		_arm_ik_repeat_thread =
-			std::thread(&MissionControlProtocol::updateArmIKRepeatTask, this);
+		// Rover responds with armIKEnabledReport after requestArmIKEnable is processed
+		if (armJointPositions.isValid()) {
+			Globals::planarArmController.set_setpoint(armJointPositions.getData());
+			Globals::armIKEnabled = true;
+			_arm_ik_repeat_thread =
+				std::thread(&MissionControlProtocol::updateArmIKRepeatTask, this);
+		}
```
abhaybd, line 147: If it's not valid, lets print a warning instead of doing nothing

abhaybd (changes requested):
Just one minor change needed, otherwise looks good

src/network/MissionControlProtocol.h
```diff
@@ -72,6 +73,7 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
 	void handleHeartbeatTimedOut();
 	void startPowerRepeat();
 	void stopAndShutdownPowerRepeat();
+	void setArmIKEnabled(const bool enabled);
```
abhaybd, line 76: There's no point in `const` pass-by-value arguments, so we should just make this `bool enabled`

src/network/MissionControlProtocol.cpp
```diff
@@ -129,25 +130,29 @@ void MissionControlProtocol::handleDriveRequest(const json& j) {
...
+			// Rover responds with armIKEnabledReport after requestArmIKEnable is processed
+			if (armJointPositions.isValid()) {
+				Globals::planarArmController.set_setpoint(armJointPositions.getData());
+				this->setArmIKEnabled(true);
+				_arm_ik_repeat_thread =
+					std::thread(&MissionControlProtocol::updateArmIKRepeatTask, this);
+			} else {
+				// unable to enable IK
+				LOG_F(WARNING, "Unable to enable IK");
+				this->sendArmIKEnabledReport();
```
abhaybd, line 151: I know `Globals::armIKEnabled` is already false, but could we change this to call `setArmIKEnabled(false)`? Otherwise, it looks weird that the `if` clause calls it but the `else` clause doesn't, since it's not immediately obvious what the difference between `setArmIKEnabled` and `sendArmIKEnabledReport` are.

abhaybd (changes requested):
Mostly good, just some minor nitpicks we should iron out

abhaybd (dismissed):
Looks good! Let's hold off on merging until the MC PR is ready, though.

abhaybd (approved):
Looks good! Just waiting on the MC pr now.

PR #272, opened 2023-10-20
Create watchdog upon connect instead of before

abhaybd:
The watchdog shouldn't be created until the first client connects, not when the protocol is registered. Otherwise the client won't respond to pings (duh, because it's not connected) which will cause the heartbeat to time out and E-stop the rover.

This fixes that bug, and was tested on hardware.


huttongrabiel (approved):
big brain

PR #273, opened 2023-10-22
Remove manual install instructions

abhaybd:
We don't really need it anymore, it just takes up space and makes things seem more complicated than it is.


huttongrabiel (approved):
lgtm. If we end up wanting them back in the future we can put them in a separate file so that it doesn't take up so much of the main readme.

PR #274, opened 2023-10-24
Updated Telemetry for Resurgence

seanzhao2021:
Replaced instances of posX and posY with longitude and latitude respectively


abhaybd (approved):
lgtm

PR #277, opened 2023-10-27
Delete and remove all references to Util.h 

QuinnP2910:
This removes all import references to Util.h and replaces them with the correct util file. Also, delete Util.h entirely.


abhaybd (approved):
In general I'd say to avoid formatting code that's untouched by your PR and instead create a separate formatting PR, but this should be fine. Nice job getting this done quickly!

PR #278, opened 2023-10-31
vfinal

AyushKulk:
aksudhas


AyushKulk:
Structurally complete code for limit calibration function returning to the standard position after calibration. At this time calls to an instance of TrapezoidalVelocityProfile from line 80 onwards do not take parameters in as expected. See encMotors in real_world_constants.h for a definition error with respect to the usage of the Frozen library.

PR #279, opened 2023-11-07
Reject enabling IK outside safety radius

DrDab:
This prevents MissionControlProtocol from setting Globals::armIKEnabled if end effector breaches safety radius.


src/network/MissionControlProtocol.cpp
```diff
@@ -452,7 +472,8 @@ void MissionControlProtocol::updateArmIKRepeatTask() {
 			robot::getMotorPositionsRad(Constants::arm::IK_MOTORS);
 		if (armJointPositions.isValid()) {
 			navtypes::Vectord<Constants::arm::IK_MOTORS.size()> targetJointPositions =
-				Globals::planarArmController.getCommand(dataclock::now(), armJointPositions);
+				Globals::planarArmController.value().getCommand(dataclock::now(),
+																armJointPositions);
```
abhaybd, line 476: We can replace `planarArmController.value().foo` with `planarArmController->foo`, which is more concise, so we should do that everywhere

src/network/MissionControlProtocol.cpp
```diff
@@ -138,14 +139,33 @@ void MissionControlProtocol::handleSetArmIKEnabled(const json& j) {
...
+			Globals::planarArmKinematics.jointPosToEEPos(armJointPositions.getData()).norm();
+		double maxRadius =
+			Globals::planarArmKinematics.getSegLens().sum() * Constants::arm::SAFETY_FACTOR;
+
+		if (eeRadius <= maxRadius) {
+			if (!Globals::planarArmController.has_value()) {
+				Globals::planarArmController.emplace(navtypes::Vectord<2>(0, 0),
+													 Globals::planarArmKinematics,
+													 Constants::arm::SAFETY_FACTOR);
+			}
```
abhaybd, line 162: This leaks implementation details a little bit, this should maybe be a factory function in `PlanarArmController`.

So maybe we could do:
```c++
if (!Globals::planarArmController.has_value()) {
    Globals::planarArmController = control::PlanarArmController::makeController(...)
}

if (Globals::planarArmController.has_value()) {
    // init thread, etc
}
```

and `makeController()` would check if the arm is in a valid position, and if it is, instantiate and return a controller, and if it isn't, return an empty optional.

src/network/MissionControlProtocol.cpp
```diff
@@ -138,14 +139,33 @@ void MissionControlProtocol::handleSetArmIKEnabled(const json& j) {
...
+		// If so, then enable IK by emplacing Globals::planarArmController and setting the
+		// setpoint. Otherwise, is a no-op.
+		double eeRadius =
+			Globals::planarArmKinematics.jointPosToEEPos(armJointPositions.getData()).norm();
+		double maxRadius =
+			Globals::planarArmKinematics.getSegLens().sum() * Constants::arm::SAFETY_FACTOR;
+
+		if (eeRadius <= maxRadius) {
+			if (!Globals::planarArmController.has_value()) {
+				Globals::planarArmController.emplace(navtypes::Vectord<2>(0, 0),
```
abhaybd, line 159: We should initialize it with the current arm position, not `{0,0}`

src/network/MissionControlProtocol.cpp
```diff
@@ -138,14 +139,33 @@ void MissionControlProtocol::handleSetArmIKEnabled(const json& j) {
...
+													 Globals::planarArmKinematics,
+													 Constants::arm::SAFETY_FACTOR);
+			}
+
+			Globals::planarArmController.value().set_setpoint(armJointPositions.getData());
```
abhaybd, line 164: If we do what I mentioned in the other review comment, we should add a method called `is_setpoint_valid()` or something to `PlanarArmController`, so we can check if the setpoint is valid before setting it, and reject otherwise.

src/Globals.cpp
```diff
@@ -4,6 +4,7 @@
 #include "world_interface/data.h"
 
 #include <atomic>
+#include <optional>
```
abhaybd, line 7: Shouldn't need this include, since we include it in the header file

abhaybd (changes requested):
Mostly look good, just some refactoring needed!

abhaybd (changes requested):
The changes since my last review look good, but the main architectural issues haven't been resolved. Take a look at my previous comments, the one about the factory function and such, and try to see what you can do there.  Let me know if you have any questions! We can talk about it at the next meeting if it helps.

src/Globals.h
```diff
@@ -5,6 +5,7 @@
 #include "world_interface/data.h"
 
 #include <atomic>
+#include <optional>
```
QuinnP2910, line 8: Is optional inclusion here necessary? It's included in world_interface/data.h.

src/network/MissionControlProtocol.cpp
```diff
@@ -12,6 +12,7 @@
 #include <functional>
 #include <loguru.hpp>
 #include <mutex>
+#include <optional>
```
QuinnP2910, line 15: Again here - is `optional` inclusion necessary? It's included in world_interface/data.h.

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -181,11 +181,11 @@ void setJointMotorPower(robot::types::jointid_t joint, double power) {
 	} else if (joint == jointid_t::ikForward || joint == jointid_t::ikUp) {
 		if (Globals::armIKEnabled) {
 			if (joint == jointid_t::ikForward) {
-				Globals::planarArmController.set_x_vel(dataclock::now(),
-													   power * Constants::arm::MAX_EE_VEL);
+				Globals::planarArmController.value().set_x_vel(
```
QuinnP2910, line 184: I believe you can use `planarArmController->set_x_vel()`. On line 187 too.

src/network/MissionControlProtocol.cpp
```diff
@@ -467,7 +468,8 @@ void MissionControlProtocol::updateArmIKRepeatTask() {
 			robot::getMotorPositionsRad(Constants::arm::IK_MOTORS);
 		if (armJointPositions.isValid()) {
 			navtypes::Vectord<Constants::arm::IK_MOTORS.size()> targetJointPositions =
-				Globals::planarArmController.getCommand(dataclock::now(), armJointPositions);
+				Globals::planarArmController->getCommand(dataclock::now(),
+																armJointPositions);
```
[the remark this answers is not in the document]
DrDab: Yes, as planarArmController is an optional

QuinnP2910 (approved):
lgtm, but get one final review from Abhay.

src/network/MissionControlProtocol.cpp
```diff
@@ -139,9 +139,16 @@ void MissionControlProtocol::handleRequestArmIKEnabled(const json& j) {
 			}
 			DataPoint<navtypes::Vectord<Constants::arm::IK_MOTORS.size()>> armJointPositions =
 				robot::getMotorPositionsRad(Constants::arm::IK_MOTORS);
-			// Rover responds with armIKEnabledReport after requestArmIKEnable is processed
-			if (armJointPositions.isValid()) {
-				Globals::planarArmController.set_setpoint(armJointPositions.getData());
+
+			auto o = control::PlanarArmController<2>::makeController(
```
huttongrabiel, line 143: Can you find a better name than o? Not really descriptive

src/control/PlanarArmController.h
```diff
@@ -44,6 +44,46 @@ class PlanarArmController {
...
+	PlanarArmController(PlanarArmController&& other)
+		: kin(std::move(other.kin)), safetyFactor(other.safetyFactor) {
+		std::lock_guard<std::mutex> lock(other.mutex);
+		setpoint = std::move(other.setpoint);
+		velocity = std::move(other.velocity);
+		velTimestamp = std::move(other.velTimestamp);
+	}
+
+	static std::optional<PlanarArmController>
+	makeController(const navtypes::Vectord<N>& currJointPos,
```
huttongrabiel, line 61: Can you add documentation for this `makeController` function?

huttongrabiel (changes requested):

DrDab:
Tested on simulator:
- Enabling the arm IK when the arm is within the safety radius is successful.
- Extending the arm beyond the safety radius prevents the arm IK from being enabled.
- Disabling the arm IK and extending the arm beyond the safety radius again again prevents the arm IK from being enabled. Arm IK is allowed to be anbled again once the arm is retracted to fit within the safety radius.

abhaybd (changes requested):
This is totally my bad, but I should have foreseen the data races this causes. We need to synchronize accesses to `Globals::planarArmController` since if we call `reset()` while another method is querying it, we could crash the robot.

Let's talk at the next meeting to iron out a plan on how to get rid of the data races.

src/control/PlanarArmController.h
```diff
@@ -36,15 +36,59 @@ class PlanarArmController {
...
+	 * @brief Instantiates and returns an std::optional containing a PlanarArmController.
+	 *
+	 * @param currJointPos The current joint positions of the arm.
+	 * @param kin_obj PlanarArmKinematics object for the arm (should have the same number of
+	 * arm joints).
+	 * @param safetyFactor the percentage factor to scale maximum arm extension radius by to
+	 * prevent singularity lock.
+	 * @return an std::optional containing a PlanarArmController if the joint positions are
+	 * within the robot's maximum arm extension radius, or an empty std::optional otherwise.
+	 */
```
abhaybd, line 65: Looks like this is the documentation for the old factory function, can you update?

src/control/PlanarArmController.h
```diff
@@ -36,15 +36,59 @@ class PlanarArmController {
...
+	/**
+	 * @brief Returns whether the target joint positions are within the arm controller's radius
+	 * limit.
+	 *
+	 * @param targetJointPos The target joint positions.
+	 * @return whether the target joint positions are within the arm controller's radius limit.
+	 */
+	static bool is_setpoint_valid(const navtypes::Vectord<N>& targetJointPos,
+								  kinematics::PlanarArmKinematics<N> kin_obj,
+								  const double safetyFactor) {
```
abhaybd, line 84: No point in marking non-reference/pointer parameters as `const`

src/control/PlanarArmController.h
```diff
@@ -36,15 +36,59 @@ class PlanarArmController {
...
+	 * @param kin_obj PlanarArmKinematics object for the arm (should have the same number of
+	 * arm joints).
+	 * @param safetyFactor the percentage factor to scale maximum arm extension radius by to
+	 * prevent singularity lock.
+	 * @return an std::optional containing a PlanarArmController if the joint positions are
+	 * within the robot's maximum arm extension radius, or an empty std::optional otherwise.
+	 */
+	bool tryInitController(const navtypes::Vectord<N>& currJointPos) {
+		if (is_setpoint_valid(currJointPos, kin, safetyFactor)) {
+			set_setpoint(currJointPos);
```
abhaybd, line 68: Don't we need to initialize `mutableFields` before we call `set_setpoint`? Otherwise we'll try accessing an empty optional. Remember to lock the mutex! In this case, it probably makes sense to initialize `mutableFields` and all its fields without calling `set_setpoint`, so we don't have to worry about double-locking of mutexes or anything.

src/control/PlanarArmController.h
```diff
@@ -36,15 +36,59 @@ class PlanarArmController {
...
+	 *
+	 * @param currJointPos The current joint positions of the arm.
+	 * @param kin_obj PlanarArmKinematics object for the arm (should have the same number of
+	 * arm joints).
+	 * @param safetyFactor the percentage factor to scale maximum arm extension radius by to
+	 * prevent singularity lock.
+	 * @return an std::optional containing a PlanarArmController if the joint positions are
+	 * within the robot's maximum arm extension radius, or an empty std::optional otherwise.
+	 */
+	bool tryInitController(const navtypes::Vectord<N>& currJointPos) {
```
abhaybd, line 66: We should consider the intended behavior if the controller is already initialized when this is called. If the joint positions are invalid, should we uninitialize the controller? e.g. by clearing the `mutableFields` optional?

tests/control/PlanarArmControllerTest.cpp
```diff
@@ -29,7 +29,8 @@ void assertApprox(const Eigen::Vector2d& p1, const Eigen::Vector2d& p2, double d
 TEST_CASE("Test Planar Arm Controller", "[control][planararmcontroller]") {
 	navtypes::Vectord<2> vec({0, 0});
 	kinematics::PlanarArmKinematics<2> kin_obj(vec, vec, vec, 0.0, 0);
-	PlanarArmController<2> foo({0, 0}, kin_obj, Constants::arm::SAFETY_FACTOR);
+	PlanarArmController<2> foo(kin_obj, Constants::arm::SAFETY_FACTOR);
+	REQUIRE(foo.tryInitController({0, 0}));
```
abhaybd, line 33: We should add a test case where we expect `tryInitController` to return false. Let's also check some edge-cases, e.g. double-init or something.

abhaybd (changes requested):
Mostly looks good, just left some comments on initialization logic!

Have you tested this in simulation? I suspect this doesn't work right now. Also, can you add to the docs in every method in the controller class to clarify what the behavior is if the controller is uninitialized? I.e. you should mention  if these methods should not be called on an uninitialized controller. Also, modify the class documentation to clarify what it means to be uninitialized, and clarify that the constructor creates it in that state.

tests/control/PlanarArmControllerTest.cpp
abhaybd, line 33: Also this test case doesn't do anything (it was initially added to force the template to be instantiated, otherwise it wouldn't be compiled) and since it initializes the segLens as 0, it will break a lot of the math, as you can see in the failing test case right now. I'd say just get rid of this test case, and then add some `CHECK_F`s in the constructor to make sure all the `segLens` are strictly positive.

huttongrabiel (approved):
lgtm

DrDab:
Tested on sim with test cases:
- Initialize IK within safety radius (succeeded as expected)
- Initialize IK outside safety radius (failed as expected)
- Re-initialize IK within safety radius after initial failure  (succeeded as expected)
- Re-initialize IK within safety radius after initial success (succeeded as expected)

PR #280, opened 2023-11-07
Implement Loguru

QuinnP2910:
Replace our old, custom logging system with Loguru. Delete log.h and log.cpp. Also, save logs to a file in the build directory.


src/Rover.cpp
```diff
@@ -94,32 +94,42 @@ void parseCommandLine(int argc, char** argv) {
...
+		.action([&](const auto&) { loguru::g_colorlogtostderr = false; })
 		.nargs(0);
 
 	try {
+		loguru::init(argc, argv);
+
+		const auto now = std::chrono::system_clock::now();
+		const std::time_t t_c = std::chrono::system_clock::to_time_t(now);
+		std::string logFileName = std::ctime(&t_c);
+		logFileName += ".log";
```
abhaybd, line 123: We should probably prefer `std::strftime` to `ctime`, so we get better control of the formatting. Note that `ctime()` appends a newline character, which we don't want.

It should probably be something like:

```c++
std::stringstream ss;
ss << std::put_time(std::localtime(&t_c), "%Y%m%d_%H%M%S.log");
std::string logFileName = ss.str();
```

src/Rover.cpp
```diff
@@ -94,32 +94,42 @@ void parseCommandLine(int argc, char** argv) {
...
+		loguru::add_file(logFileName.c_str(), loguru::Append, loguru::g_stderr_verbosity);
```
abhaybd, line 124: Ideally, we could also do log file rotation. I.e. we should have a log file called `latest.log` or something, so we don't have to type in the full date-time every time we want to see the latest log (which is what we'll usually want)

We can do this by having 2 `add_file` calls, one for `logFileName`, and another for `latest.log`, set to truncate mode.

src/LimitCalib.cpp
```diff
@@ -22,7 +22,7 @@ void callback(motorid_t motor, const types::DataPoint<LimitSwitchData>& data) {
 };
 
 void cleanup(int signum) {
-	log(LOG_ERROR, "Interrupted!\n");
+	LOG_F(ERROR, "Interrupted!\n");
```
abhaybd, line 25: I'll just put the note here instead of everywhere, but I believe loguru automatically appends a newline, so we can remove the newline at the end of all of our logs.

src/Rover.cpp
abhaybd, line 124: Any reason we shouldn't set verbosity to more? Maybe not max, but probably more than this. Maybe like Verbosity_INFO? What do you think?

abhaybd (changes requested):
This is pretty good, nice job getting it integrated! Left some notes about some additional functionality and verbosity. Plus, like we briefly talked about, we should change all asserts to `CHECK`s. If you want, you can do the `CHECK`s in a new PR after this one, whatever you prefer.

src/Rover.cpp
```diff
@@ -94,32 +95,82 @@ void parseCommandLine(int argc, char** argv) {
...
+					tm_struct.tm_sec = second;
+
+					// Current time
+					std::time_t unixTime = std::mktime(&tm_struct);
+
+					// Delete log if it's older than LOG_LIFESPAN
+					if (std::chrono::system_clock::to_time_t(
+							std::chrono::system_clock::now()) -
+							unixTime >
+						LOG_LIFESPAN) {
```
abhaybd, line 150: `std::time_t` has very little guarantees, i.e. it's not guaranteed to be in seconds, and it's technically not even guaranteed to count up. Let's prefer other time representations, so we could do this instead:

```c++
if (std::chrono::system_clock::now() - std::chrono::from_time_t(unixTime) > LOG_LIFESPAN) { ... }
```

src/Rover.cpp
```diff
@@ -94,32 +95,82 @@ void parseCommandLine(int argc, char** argv) {
...
 		.help("disables colors in console logging")
-		.action([&](const auto&) { setColorsEnabled(false); })
+		.action([&](const auto&) { loguru::g_colorlogtostderr = false; })
 		.nargs(0);
 
 	try {
+		loguru::init(argc, argv);
+		namespace fs = std::filesystem;
+
+		const int LOG_LIFESPAN = 604800; // 7 days
```
abhaybd, line 122: It's unclear what units this is in, and in general we should prefer `std::duration` to `int` of seconds or something. So, for clarity, this could instead be:

```c++
const auto LOG_LIFESPAN = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::days(7));
```

which is much clearer that the log lifespan is 7 days, in seconds.

src/Rover.cpp
```diff
@@ -94,32 +95,82 @@ void parseCommandLine(int argc, char** argv) {
...
+		try {
+			for (const auto& entry : fs::directory_iterator(fs::current_path())) {
+				std::cout << entry.path().extension();
+				if (entry.path().extension() == ".log" && entry.path().stem() != "latest") {
+					std::string dateString = entry.path().stem();
+
+					// Extract components from the date string
+					int year, month, day, hour, minute, second;
+					sscanf(dateString.c_str(), "%4d%2d%2d_%2d%2d%2d", &year, &month, &day,
+						   &hour, &minute, &second);
```
abhaybd, line 132: To be safe, check the return value is what we expect?

abhaybd (dismissed):
Mostly good, just some notes about time handling

src/Rover.cpp
```diff
@@ -94,32 +95,90 @@ void parseCommandLine(int argc, char** argv) {
...
+									std::chrono::system_clock::from_time_t(unixTime);
+
+					// Delete log if it's older than LOG_LIFESPAN
+					if (duration > LOG_LIFESPAN) {
+						fs::remove(entry.path());
+					}
+				}
+			}
+		} catch (const fs::filesystem_error& e) {
+			std::cerr << "Error accessing current directory!" << e.what() << std::endl;
```
huttongrabiel, line 165: Should probably add a space after the `!`. Might even want to do something like `std::cerr << "Error accessing current directory: " << e.what() << std::endl;`. Adding a space would be fine too. Just want it so that the string from `e.what()` is not just smashed up against the other string.

huttongrabiel (dismissed):
Mostly good! Just two really minor things.

src/Rover.cpp
```diff
@@ -94,32 +95,89 @@ void parseCommandLine(int argc, char** argv) {
...
+					sscanf(dateString.c_str(), "%4d%2d%2d_%2d%2d%2d", &year, &month, &day,
+						   &hour, &minute, &second);
+
+					// Ensure scan outputs are correct
+					CHECK_EQ_F(std::stoi(dateString.substr(0, 4)), year);
+					CHECK_EQ_F(std::stoi(dateString.substr(4, 2)), month);
+					CHECK_EQ_F(std::stoi(dateString.substr(6, 2)), day);
+					CHECK_EQ_F(std::stoi(dateString.substr(9, 2)), hour);
+					CHECK_EQ_F(std::stoi(dateString.substr(11, 2)), minute);
+					CHECK_EQ_F(std::stoi(dateString.substr(13, 2)), second);
```
huttongrabiel, line 139: This is probably fine, maybe better because it's more verbose, but you could also just check  that the return value from `sscanf` is equal to 6. @abhaybd any input on this?

src/Rover.cpp
```diff
@@ -94,32 +95,89 @@ void parseCommandLine(int argc, char** argv) {
...
 
 	try {
+		loguru::init(argc, argv);
+		namespace fs = std::filesystem;
+
+		const auto LOG_LIFESPAN = std::chrono::hours(24 * 7);
+		try {
+			for (const auto& entry : fs::directory_iterator(fs::current_path())) {
+				if (entry.path().extension() == ".log" && entry.path().stem() != "latest") {
+					std::string dateString = entry.path().stem();
```
huttongrabiel, line 126: Could you just add a comment above this to show the format of dateString? Something like `// Format of dateString: YYYYMMDD_HHMMSS`. I know you can get this from reading the sscanf below but its more obvious this way.

huttongrabiel (approved):
lgtm. I'm going to approve this because the couple more comments I left are non-blocking. I'll leave final say up to @abhaybd.

src/Rover.cpp
abhaybd: I agree, I think these `stoi` checks are redundant. I think it might be just simpler and clearer to just check that `sscanf(...) == 6`. Are there situations where `sscanf()` returns the right value but the parsed values are incorrect? I don't think so, and in fact, if the string is malformed these `stoi` calls would just throw uncaught exceptions.

abhaybd (changes requested):
See my reply on Hutton's comment, but I think that should be the last thing

src/Rover.cpp
QuinnP2910: For sure. Can do!

huttongrabiel (approved):
lgtm. Just a heads up, when you make changes, make sure you re-request a review from us so that we see we have something to review.

PR #283, opened 2023-11-14
Represent rover orientation as quaternion instead of euler angles

abhaybd:
Euler angles suck:
- they are ambiguous (granted, quaternions are a double-cover of SE3, but at least it's very easy to disambiguate, unlike euler angles)
- Euler angles are defined by 3 angles AND axis conventions, which will differ from code we may see in other places, and create hard-to-discover bugs, plus its hard to enforce. Quats don't have any additional associated conventions.
- Quaternions lend themselves more to linear algebra computations

The above reasons meant `readIMUHeading()` was wrong, since it wasn't giving the correct heading value, since the quat was being decomposed "incorrectly" into the euler angles. It was correct (technically), but wasn't what we wanted.

So this pr just represents rover orientation with quats instead of euler angles. And the heading is very easy to recover directly from the quat.

### DIFF
--- a/src/ardupilot/ArduPilotInterface.cpp
+++ b/src/ardupilot/ArduPilotInterface.cpp
@@ -25,7 +25,7 @@ DataPoint<gpscoords_t> readGPSCoords() {
 } // namespace gps
 
 namespace robot {
-DataPoint<eulerangles_t> readIMU() {
+DataPoint<Eigen::Quaterniond> readIMU() {
 	return ardupilot_protocol->getIMU();
 }
 } // namespace robot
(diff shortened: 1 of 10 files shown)


QuinnP2910 (dismissed):

huttongrabiel (dismissed):
Eventually we should probably change everything over to quaternions (see ArduPilotProtocol and ArduPilotInterface, maybe other places too I didn't check all). Whether you want to do that in this PR or later is up to you. Otherwise looks good.

abhaybd:
> Eventually we should probably change everything over to quaternions (see ArduPilotProtocol and ArduPilotInterface, maybe other places too I didn't check all). Whether you want to do that in this PR or later is up to you. Otherwise looks good.

Good point, just for consistency I replaced those with quats too. The ardupilot internally does rpy (in the ardupilot-driver code) so I don't know if there's a point in replacing that, but at least here all public interfaces deal with quaternions.

huttongrabiel (approved):
lgtm. and I don't think there is any point in replacing the ardupilot-driver side of things for now. Just nice to have the rover code all on the same page.

PR #284, opened 2023-11-18
Lock protocolMap with a mutex to prevent data race

QuinnP2910:


src/network/websocket/WebSocketServer.cpp
```diff
@@ -174,6 +179,7 @@ void SingleClientWSServer::onClose(connection_hdl hdl) {
 	LOG_F(INFO, "Server=%s, Endpoint=%s : Connection disconnected from %s", serverName.c_str(),
 		  path.c_str(), client.c_str());
 
+	std::lock_guard lock(protocolMapMutex);
```
abhaybd, line 182: Genuine question, should we only lock the protocol map while accessing it, or should we keep it locked for the whole method? What do you think is there opportunity for race conditions if we do the former?

abhaybd (changes requested):
mostly looks good!

src/network/websocket/WebSocketServer.cpp
QuinnP2910: I think it makes sense to only lock while accessing. Should I migrate to unique_lock and unlock post-access?

abhaybd: I think if we used `unique_lock` we'd have a lot of code duplication since we'd have to this every time we access the map:
```c++
std::unique_lock lock(protocolMapMutex);
auto& protocolData = protocolMap.at(path);
lock.unlock();
```

Instead, it might be a bit cleaner to just make a new private method `ProtocolData& getProtocol(const std::string& protocolPath)` that handles synchronized access to the map, so it reduces code duplication, which would look something like:

```c++
ProtocolData& getProtocol(const std::string& protocolPath) {
    std::lock_guard lock(protocolMapMutex);
    return protocolMap.at(protocolPath);
}
```

Thoughts?

src/network/websocket/WebSocketServer.cpp
```diff
@@ -91,6 +93,7 @@ bool SingleClientWSServer::addProtocol(std::unique_ptr<WebSocketProtocol> protoc
 
 void SingleClientWSServer::sendRawString(const std::string& protocolPath,
 										 const std::string& str) {
+	std::lock_guard lock(protocolMapMutex);
```
abhaybd, line 96: I think this is also a case where we shouldn't hold the lock of the protocol map for the entire method. What if we changed `getProtocol()` to return an `std::optional<ProtocolData&>`, and everywhere else (where we guarantee that the entry exists) we immediately call `value()`, while here we can check if the optional is filled or not?

abhaybd (changes requested):
Mostly good, just a note about locking in `sendRawString()`.

As an aside, it seems there might be a few other data races? `ProtocolData` defines a `mutex` that protects `ProtocolData::client` and `ProtocolData::heartbeatInfo`, but that mutex isn't necessarily locked everywhere it should be, for example in the `sendRawString()` method. Can you take a pass through `WebSocketServer.cpp` and ensure that the locking is correct for `ProtocolData`?

src/network/websocket/WebSocketServer.cpp
```diff
@@ -94,9 +94,9 @@ bool SingleClientWSServer::addProtocol(std::unique_ptr<WebSocketProtocol> protoc
 void SingleClientWSServer::sendRawString(const std::string& protocolPath,
 										 const std::string& str) {
 	std::lock_guard lock(protocolMapMutex);
-	auto entry = protocolMap.find(protocolPath);
-	if (entry != protocolMap.end()) {
-		auto& protocolData = entry->second;
+	auto protocolDataOpt = this->getProtocol(protocolPath);
+	if (protocolDataOpt.has_value()) {
+		auto& protocolData = protocolDataOpt.value().get();
```
abhaybd, line 100: This is super minor, but imo chaining getters is a little ugly. We could instead do:
```c++
ProtocolData& protocolData = protocolDataOpt.value();
```
Could you replace that everywhere?

abhaybd (changes requested):
looks good but tiny nitpick. After addressing, make sure you test in sim and make sure everything still works properly!

src/network/websocket/WebSocketServer.cpp
```diff
@@ -190,12 +197,13 @@ void SingleClientWSServer::onMessage(connection_hdl hdl, message_t message) {
...
-	auto it = protocolMap.find(path);
-	if (it != protocolMap.end()) {
+	auto protocolDataOpt = this->getProtocol(path);
+	if (protocolDataOpt.has_value()) {
+		ProtocolData& pd = protocolDataOpt.value();
 		std::string jsonStr = message->get_payload();
 		LOG_F(1, "Message on %s: %s", path.c_str(), jsonStr.c_str());
 		json obj = json::parse(jsonStr);
-		it->second.protocol->processMessage(obj);
+		pd.protocol->processMessage(obj);
```
abhaybd, line 206: Can you add a comment here saying that locking isn't necessary? Just so we don't in the future wonder why we didn't add a lock here.

abhaybd (changes requested):
This looks good! One last thing, can you replace all usages of `protocolMap.find()` with `getProtocol()` usages? Like in `onPong()` and methods like that.

abhaybd (approved):
Looks good, nice job on this!

PR #285, opened 2023-11-28
DiffDriveKinematics: Scale wheel velocity to stay below max speed

huttongrabiel:
Sometimes we want to scale the wheel velocity so that we focus more on rotational velocity or linear velocity. There are cases where the given theta velocity and linear velocity result in a wheel speed too great for the rover.

### DIFF
--- a/src/kinematics/DiffDriveKinematics.cpp
+++ b/src/kinematics/DiffDriveKinematics.cpp
@@ -1,7 +1,9 @@
 #include "DiffDriveKinematics.h"
 
+#include "../Constants.h"
 #include "../utils/transform.h"
 
+#include <cmath>
 #include <loguru.hpp>
 using namespace navtypes;
 using util::toTransformRotateFirst;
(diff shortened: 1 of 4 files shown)


src/kinematics/DiffDriveKinematics.cpp
```diff
@@ -49,4 +51,34 @@ pose_t DiffDriveKinematics::getNextPose(const wheelvel_t& wheelVel, const pose_t
...
+			case PreferredVelPreservation::Proportional:
+				xVel *= scaleFactor;
+				thetaVel *= scaleFactor;
+				break;
+			case PreferredVelPreservation::PreferXVel:
+				thetaVel *= (0.5 * scaleFactor);
+				break;
+			case PreferredVelPreservation::PreferThetaVel:
+				xVel *= (0.5 * scaleFactor);
+				break;
```
abhaybd, line 72: Don't think this math is right, since I don't think we can just do regular scaling here. Refer to the equation [here](https://github.com/huskyroboticsteam/Resurgence/pull/285/files#diff-b474567bf27f33d3c2f4b94d7af43bd3a081d63bb4017e7c2f0ca3b09aea37a3L16). 

If we're preferring xVel, then we should change thetaVel to satisfy the wheel speed constraint. Somewhat sure this pseudocode is correct.

```
if abs(xVel) > maxSpeed:
    return {copysign(maxSpeed, xVel), 0, 0}

if lVel < -maxSpeed:
    thetaVel = (-maxSpeed - xVel) * -2.0 / wheelBaseWidth
if lVel > maxSpeed:
    thetaVel = (maxSpeed - xVel) * -2.0 / wheelBaseWidth
if rVel < -maxSpeed:
    thetaVel = (-maxSpeed - xVel) * 2.0 / wheelBaseWidth
if rVel > maxSpeed:
    thetaVel = (maxSpeed - xVel) * 2.0 / wheelBaseWidth
```

And we do a similar thing for prefer theta vel, again somewhat sure about correctness:

```
if abs(wheelBaseWidth / 2 * thetaVel) > maxSpeed:
    return {0, 0, copysign(maxSpeed * 2 / wheelBaseWidth, thetaVel)}

if max(lVel, rVel) > maxSpeed:
    xVel -= max(lVel, rVel) - maxSpeed
if min(lVel, rVel) < -maxSpeed:
    xVel += -maxSpeed - min(lVel, rVel)
```

abhaybd (changes requested):

src/kinematics/DiffDriveKinematics.cpp
```diff
@@ -1,6 +1,10 @@
 #include "DiffDriveKinematics.h"
 
+#include "../Constants.h"
```
abhaybd, line 3: This class shouldn't rely on constants. I think it's just because you use `Constants::MAX_WHEEL_VEL` below instead of `maxWheelVel`, which is an easy fix.

abhaybd (changes requested):
Mostly looks good, just some quick fixes needed

tests/kinematics/DiffDriveKinematicsTest.cpp
```diff
@@ -60,4 +54,53 @@ TEST_CASE("Differential Drive Kinematics Test")
...
 				 kinematics.getNextPose(wheelVel(3 * PI / 4, PI / 4), {0, 0, 0}, 4));
-}
\ No newline at end of file
+
+	constexpr double max_wheel_vel = 0.75;
+
+	// Proportional Preservation Tests
+	assertApprox({0.5, 0, 0.5},
+				 kinematics.ensureWithinWheelSpeedLimit(
+					 kinematics::DiffDriveKinematics::PreferredVelPreservation::Proportional,
```
abhaybd, line 63: We have `using namespace kinematics` at the top, so you can simplify this to `DiffDriveKimatics::...`, both here and everywhere else further down.

tests/kinematics/DiffDriveKinematicsTest.cpp
```diff
@@ -60,4 +54,53 @@ TEST_CASE("Differential Drive Kinematics Test")
...
+					 1.0, 1.0, max_wheel_vel));
+
+	// xVel Preservation Tests
+	// Test #1: Test the xVel alone being too fast, therefore xVel should become the max and
+	// theta should
+	//          be zeroed
```
abhaybd, line 69: is this weird linebreak on purpose?

tests/kinematics/DiffDriveKinematicsTest.cpp
```diff
@@ -60,4 +54,53 @@ TEST_CASE("Differential Drive Kinematics Test")
...
+				 kinematics.ensureWithinWheelSpeedLimit(
+					 kinematics::DiffDriveKinematics::PreferredVelPreservation::PreferXVel,
+					 0.6, -0.5714, max_wheel_vel));
+	// Test #4: rVel < -max_wheel_vel: lVel = 0.2 rVel = -0.9
+	assertApprox({-0.35, 0, -0.8},
+				 kinematics.ensureWithinWheelSpeedLimit(
+					 kinematics::DiffDriveKinematics::PreferredVelPreservation::PreferXVel,
+					 -0.35, -1.1, max_wheel_vel));
+	// Test #5: rVel > max_wheel_vel: lVel = 0.2 rVel = 0.9
+	assertApprox({0.55, 0, 0.4},
```
abhaybd, line 90: Nice test coverage!

tests/kinematics/DiffDriveKinematicsTest.cpp
```diff
@@ -60,4 +54,53 @@ TEST_CASE("Differential Drive Kinematics Test")
...
+					 0.55, 0.7, max_wheel_vel));
+
+	// thetaVel Preservation Tests
+	// Test #1: std::max(lVel, rVel) > maxWheelSpeed: lVel = 0.9 r_vel = 0.35
+	assertApprox({0.475, 0, -0.55},
+				 kinematics.ensureWithinWheelSpeedLimit(
+					 kinematics::DiffDriveKinematics::PreferredVelPreservation::PreferThetaVel,
+					 0.625, -0.55, max_wheel_vel));
+	// Test #2: std::min(lVel, rVel) < -maxWheelSpeed: lVel = -0.9 rVel = -0.2
+	assertApprox({-0.4, 0, 0.7},
```
abhaybd, line 102: Can we also test the case where the thetaVel alone is too fast?

abhaybd (changes requested):
Almost there! Everything mostly looks good, just some notes about using `std::abs` instead of `abs`, and then some minor nitpicks beyond that.

src/kinematics/DiffDriveKinematics.cpp
```diff
@@ -49,4 +50,55 @@ pose_t DiffDriveKinematics::getNextPose(const wheelvel_t& wheelVel, const pose_t
...
+	PreferredVelPreservation preferredVelPreservation, double xVel, double thetaVel,
+	double maxWheelSpeed) const {
+	auto [lVel, rVel] = robotVelToWheelVel(xVel, thetaVel);
+	double calculatedMaxWheelVel = std::max(std::abs(lVel), std::abs(rVel));
+	if (calculatedMaxWheelVel > maxWheelSpeed) {
+		switch (preferredVelPreservation) {
+			case PreferredVelPreservation::Proportional: {
+				xVel *= maxWheelSpeed / calculatedMaxWheelVel;
+				thetaVel *= maxWheelSpeed / calculatedMaxWheelVel;
+			}
```
abhaybd, line 64: Missing a break?

abhaybd (changes requested):
Looks good, just I think a minor thing then should be good to go

src/kinematics/DiffDriveKinematics.cpp
huttongrabiel: fixed. also added const qualifiers to the variables declared at the top of the function that should not change

abhaybd (approved):
Looks good, nice job on this!

PR #286, opened 2023-12-05
Don't repeatedly send packets commanding 0 power to drivebase

abhaybd:
Minor improvement that only resends drivebase commands if they're nonzero. Note that this doesn't affect the first time it's sent, just the periodic resends.


QuinnP2910 (approved):
Looks good.

PR #290, opened 2023-12-09
Name threads for logging purposes

abhaybd:
Loguru prints thread names for log statements, so naming them something human-readable will make logs clearer to read, especially when debugging any weirdness with multithreading.


QuinnP2910 (approved):
Looks good!

src/utils/scheduler.h
```diff
@@ -190,6 +199,23 @@ class PeriodicScheduler : private impl::Notifiable {
...
+	 *
+	 * @param name The name of this Watchdog, for logging purposes.
+	 * @param duration The timeout duration. If not fed for at least this long, then the
+	 * callback is invoked.
+	 * @param callback The callback to invoke when the watchdog starves.
+	 * @param keepCallingOnDeath If true, keep invoking @p callback every @p duration
+	 * milliseconds until fed again. Otherwise, only call @p callback when starved, and do not
+	 * call again until being reset and subsequently starved again.
+	 */
+	Watchdog(const std::string& name, std::chrono::milliseconds duration,
```
huttongrabiel, line 213: Is there a reason to add another constructor? Can a default argument for `name` be used like in the above `PeriodicScheduler` constructor?

huttongrabiel (changes requested):

src/utils/scheduler.h
abhaybd: I feel like the `name` argument should be first, and default arguments always have to come at the end, so I had to make an overloaded constructor.

Do you think it should just be turned into a default argument at the end of the argument list?

huttongrabiel: Oh good point, oversight on my part that they have to come at end. I agree with you then on the overloaded function. It makes more sense to have `name` come first.

huttongrabiel (approved):

PR #292, opened 2023-12-09
Remove even more vestigial code

abhaybd:
Removed even more vestigial code:
- All factor graph code (it was completely undocumented and nobody knows how it works)
- Several ad-hoc tester scripts (CANTest, TestIMU, CameraTester, etc)
- Removed AR usages from sim interface. We're keeping the tag detection code around, but we're not using it since we need to revisit that later


src/filters/FilterPerfTest.cpp
```diff
@@ -30,14 +30,11 @@ void usage(const char* name);
 template <typename K, typename V>
 constexpr std::unordered_map<V, K> reverseMap(const std::unordered_map<K, V>& map);
 
-enum class FilterType
-{
-	EKF,
-	PoseGraph
+enum class FilterType {
+	EKF
 };
```
QuinnP2910, line 35: An example of the weird enum linting contradiction we were talking about, GitHub CI fails

QuinnP2910: I did some research into this, I think the GitHub check just calls the clang-format-11, which defaults to formatting based on just LLVM. Adding the parameter `-style=file` should use the .clang-format in out code.

QuinnP2910: The line:
`diff=$(git clang-format-11 --diff)`
Should be
`diff=$(git clang-format-11 -style=file --diff)`
In the formatted workflow

QuinnP2910 (approved):
Cool, CI passes now and changes look good.

src/filters/FilterPerfTest.cpp
abhaybd: Format check fixed by #295

PR #294, opened 2023-12-09
Upgrade format checker CI to clang-format-18

abhaybd:
We were previously using 11, which has a [bug](https://reviews.llvm.org/D106349) that was messing with our enums. Upgrading to 18 should fix this.

### DIFF
--- a/.github/workflows/format-checker.yml
+++ b/.github/workflows/format-checker.yml
@@ -12,14 +12,16 @@ jobs:
         submodules: true
     - name: Install dependencies from APT
       run: |
-        sudo apt install -y git clang-format-11
-        sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-11 100 \
-          --slave /usr/share/man/man1/clang-format.1.gz clang-format.1.gz /usr/share/man/man1/clang-format-11.1.gz
+        wget https://apt.llvm.org/llvm.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/format-checker.yml
```diff
@@ -12,14 +12,16 @@ jobs:
...
+        chmod +x llvm.sh
+        sudo ./llvm.sh 18
+        sudo apt install -y git clang-format-18
+        sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-18 100
     - name: Check Formatting
       run: |
         cd ${{ github.workspace }}
         git reset --soft $(git merge-base HEAD origin/master)
-        diff=$(git clang-format-11 --diff)
+        diff=$(git clang-format-18 --style=file --diff)
```
QuinnP2910, line 24: I think you were right about -`-style=file`, it should be the default, so I think no need for it here unless you want to be really clear that we're formatting from a file

abhaybd: Yeah, I figured we might as well be explicit about it. Saves us the trouble of going and finding the docs next time we forget if that's the default or not

QuinnP2910: Cool, looks good then!

QuinnP2910 (approved):

PR #295, opened 2023-12-09
Make CI work

abhaybd:
We installed the alternative but then didn't set it. This fixes that


QuinnP2910 (approved):
I saw that this change worked on #292 , so should be good!

PR #296, opened 2024-01-12
Improve compiler warning flags

QuinnP2910:
Right now, I've made this likely as verbose as we would possibly have it. It throws warnings for 
`-Wall
  -Wextra
  -Wconversion
  -Wpedantic`
See [here](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html).
  
Compile this and let me know which warnings make sense and which to suppress. Then, do we want to `-Werror` to really enforce these? This would require a lot of code cleanup to make our code compile without errors.

### DIFF
--- a/src/CAN/FakeCANBoard.cpp
+++ b/src/CAN/FakeCANBoard.cpp
@@ -20,7 +20,8 @@ using namespace std::chrono_literals;
 using can::motor::motormode_t;
 using namespace robot::types;
 
-enum class TestMode {
+enum class TestMode
+{
 	ModeSet,
 	PWM,
 	PID,
@@ -163,9 +164,11 @@ int main() {
 			[[maybe_unused]] double setPoint =
 				(targetVel * util::durationToSec(currTime - startTime)) +
 				initialMotorPos.getData();
+			LOG_F(INFO, "Set point: %f", setPoint);
 
 			// get y data: motor position
 			[[maybe_unused]] robot::types::DataPoint<int32_t> motorPos = motor->getMotorPos();
+			LOG_F(INFO, "Motor position: %d", motorPos.getData());
 
 			// check if time is up
 			double elapsedTime = util::durationToSec(currTime - startTime);
--- a/src/CAN/FakeCANBoard.cpp
+++ b/src/CAN/FakeCANBoard.cpp
@@ -20,8 +20,7 @@ using namespace std::chrono_literals;
 using can::motor::motormode_t;
 using namespace robot::types;
 
-enum class TestMode
-{
+enum class TestMode {
 	ModeSet,
 	PWM,
 	PID,
--- a/src/CAN/FakeCANBoard.cpp
+++ b/src/CAN/FakeCANBoard.cpp
@@ -160,13 +160,12 @@ int main() {
 			robot::types::datatime_t currTime = robot::types::dataclock::now();
 
 			// get y data: set point (target vel * time since set velocity call) + initial pos
-			[[maybe_unused]] double setPoint =
-				(targetVel * util::durationToSec(currTime - startTime)) +
-				initialMotorPos.getData();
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -360,6 +353,13 @@ if (WORLD_INTERFACE STREQUAL "REAL")
   target_link_libraries(LimitSwitchCalibration real_world_interface)
 endif()
 
+add_compile_options(
+  -Wall
+  -Wextra
+  -Wno-deprecated-copy
```
abhaybd, line 359: Instead of adding this flag we should fix the error that it's reporting. It looks like it caught a legitimate bug, we should have defined the copy assignment operator in `CameraParams`

src/network/MissionControlProtocol.cpp
```diff
@@ -228,8 +228,8 @@ static bool validateJointPositionRequest(const json& j) {
 static void handleJointPositionRequest(const json& j) {
 	// TODO: ignore this message if we are in autonomous mode.
 	std::string motor = j["joint"];
-	double position_deg = j["position"];
-	int32_t position_mdeg = std::round(position_deg * 1000);
+	[[maybe_unused]] double position_deg = j["position"];
+	[[maybe_unused]] int32_t position_mdeg = std::round(position_deg * 1000);
```
abhaybd, line 232: Instead of adding these hints, we can just comment out the body of this method and leave a `TODO` to implement at some point

src/CAN/FakeCANBoard.cpp
```diff
@@ -160,11 +160,12 @@ int main() {
 			robot::types::datatime_t currTime = robot::types::dataclock::now();
 
 			// get y data: set point (target vel * time since set velocity call) + initial pos
-			double setPoint = (targetVel * util::durationToSec(currTime - startTime)) +
-							  initialMotorPos.getData();
+			[[maybe_unused]] double setPoint =
+				(targetVel * util::durationToSec(currTime - startTime)) +
+				initialMotorPos.getData();
```
abhaybd, line 165: Instead of marking as unused, we should just print out `setPoint` and `motorPos`. It's an easy enough fix, can you add that?

abhaybd (changes requested):
Looks good, just a few changes needed

src/CAN/FakeCANBoard.cpp
```diff
@@ -160,11 +160,14 @@ int main() {
 			robot::types::datatime_t currTime = robot::types::dataclock::now();
 
 			// get y data: set point (target vel * time since set velocity call) + initial pos
-			double setPoint = (targetVel * util::durationToSec(currTime - startTime)) +
-							  initialMotorPos.getData();
+			[[maybe_unused]] double setPoint =
```
abhaybd, line 163: No point in marking these variables as unused if we immediately print it, so we should remove those attributes for `setPoint` and `motorPos`

src/network/MissionControlProtocol.cpp
```diff
@@ -225,11 +225,11 @@ static bool validateJointPositionRequest(const json& j) {
...
 
-static void handleJointPositionRequest(const json& j) {
+static void handleJointPositionRequest([[maybe_unused]] const json& j) {
 	// TODO: ignore this message if we are in autonomous mode.
-	std::string motor = j["joint"];
-	double position_deg = j["position"];
-	int32_t position_mdeg = std::round(position_deg * 1000);
+	// std::string motor = j["joint"];
+	// [[maybe_unused]] double position_deg = j["position"];
+	// [[maybe_unused]] int32_t position_mdeg = std::round(position_deg * 1000);
```
abhaybd, line 232: If we're commenting the line out we should just remove the unused attributes

abhaybd (changes requested):
Some minor fixes, and then this should be good!

abhaybd (approved):
Looks good!

PR #300, opened 2024-01-21
Add tasks to utils

abhaybd:
I'm in the process of cleaning up MissionControlProtocol, and factoring out the tasks from the big class into their own individual classes that encapsulate their own state.

This PR adds the required task classes and unit tests, in preparation of that.

I'm reasonably sure I caught any data races in `AsyncTask`, but let me know if you find any others!

### DIFF
--- a/src/utils/scheduler.h
+++ b/src/utils/scheduler.h
@@ -327,6 +327,9 @@ public:
 
 	virtual ~AsyncTask() {
 		stop();
+		if (thread.joinable()) {
+			thread.join();
+		}
 	}
 
 	AsyncTask& operator=(const AsyncTask&) = delete;
@@ -339,6 +342,7 @@ public:
 	virtual void start() {
 		std::lock_guard lock(mutex);
 		if (!running) {
+			std::lock_guard threadLock(threadMutex);
 			if (thread.joinable()) {
 				thread.join();
 			}
@@ -364,6 +368,7 @@ public:
 		}
 		if (isRunning) {
 			cv.notify_one();
+			std::lock_guard threadLock(threadMutex);
 			if (thread.joinable()) {
 				thread.join();
 			}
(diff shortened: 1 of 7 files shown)


src/utils/scheduler.h
```diff
@@ -299,6 +301,219 @@ class Watchdog : private impl::Notifiable {
...
+	AsyncTask(const std::optional<std::string>& name = std::nullopt)
+		: name(name), running(false), quitting(false) {}
+
+	AsyncTask(const AsyncTask&) = delete;
+
+	virtual ~AsyncTask() {
+		stop();
+		if (thread.joinable()) {
+			thread.join();
+		}
```
QuinnP2910, line 332: Is this necessary? `stop()` joins the thread and waits for the task to finish if it's running, and if it's not running, there's no need to join and wait, if my understanding is correct. Is this just for extra safety?

abhaybd: Good question, `stop()` only joins the thread if `running` is true, but there is the edge case where the thread has already exited. In that case, `running` is false, but the thread still needs to be joined to clean up its resources.

`running` tracks whether the task is logically running or not, not necessarily if the thread is active or not. So, in the destructor, we need to explicitly join the thread to make sure we clean it up.

src/utils/scheduler.h
```diff
@@ -299,6 +301,219 @@ class Watchdog : private impl::Notifiable {
...
+	 * @brief Wait for a given duration, or until the task has been stopped.
+	 *
+	 * @param lock The lock passed to AsyncTask::task.
+	 * @param tp The duration of time to wait for.
+	 * @return true iff the task was stopped while waiting.
+	 */
+	template <typename Rep, typename Period>
+	bool wait_for(std::unique_lock<std::mutex>& lock,
+				  const std::chrono::duration<Rep, Period>& dur) {
+		return wait_until(lock, Clock::now() + dur);
```
huttongrabiel, line 432: Missing lambda to check for quitting.

```cpp
return wait_until(lock, Clock::now() + dur, [&]() { return quitting; });
```

huttongrabiel (changes requested):
~~Mostly good. Just see one thing at the moment~~

huttongrabiel (dismissed):
lgtm

huttongrabiel:
@abhaybd Formatting errors on this

QuinnP2910 (approved):
Looks good!

PR #301, opened 2024-01-21
Fix format ci

abhaybd:
For some reason it seems like stderr wasn't being printed in the CI logs, so let's just print to stdout instead.


abhaybd:
Obsoleted by #303

PR #303, opened 2024-01-23
Downgrade format checker to clang-format-14

QuinnP2910:
For some reason, git clang-format-18 --style=file --diff would exit with error code 1. This version doesn't, but also fixes the enum bracket issue that caused us to upgrade to 18.

### DIFF
--- a/.github/workflows/format-checker.yml
+++ b/.github/workflows/format-checker.yml
@@ -14,15 +14,15 @@ jobs:
       run: |
         wget https://apt.llvm.org/llvm.sh
         chmod +x llvm.sh
-        sudo ./llvm.sh 18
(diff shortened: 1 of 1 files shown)


QuinnP2910:
I tested to make sure this really does solve the enum issue.
[This commit](https://github.com/huskyroboticsteam/Resurgence/pull/302/commits/7d71a39ed7b822df11129b5b9a7a9f699d5fe1ff) is version 11 and fails incorrectly
[This commit](https://github.com/huskyroboticsteam/Resurgence/pull/302/commits/1be73af3997d013bb19868c0b0c2e86534b3445f) is version 14 and passes

abhaybd (approved):
Looks good, nice fix!

PR #304, opened 2024-01-26
Handle IK failure gracefully

abhaybd:
Before, the velocity wouldn't get set to zero so the arm wouldn't move but the setpoint would keep moving at the current velocity. This PR ensure it gets handled properly, so the velocity is reset and the setpoint is made to be correct.


QuinnP2910 (approved):
lgtm

PR #305, opened 2024-01-28
Clean up MCP by consolidating long-running tasks using new task types

abhaybd:
This PR does a lot of refactoring and cleanup in MissionControlProtocol. Previously, that class had several long-running tasks (and their required data members) as fields, which made the files very long and the class definition was complicated.

This PR factors all that code out, so each task is an independent type. Now, the MCP object can just instantiate the task and control it as necessary, which makes it easier to edit and extend.


src/network/MissionControlTasks.cpp
```diff
@@ -0,0 +1,186 @@
...
+	_last_joint_power[id] = power;
+}
+
+void PowerRepeatTask::setCmdVel(double steerVel, double xVel) {
+	std::lock_guard lock(_mutex);
+	_last_cmd_vel = {steerVel, xVel};
+}
+
+void PowerRepeatTask::start() {
+	// lock the mutex before start() to prevent data races between start() and stop()
```
Geeoon, line 38: Would the mutex in the periodictask/asynctask class not prevent this already?

src/network/MissionControlTasks.cpp
```diff
@@ -0,0 +1,186 @@
...
+	std::lock_guard lock(_mutex);
```
Geeoon, line 39: Does the mutex in the AsyncTask/PeriodicTask class not prevent this?

abhaybd: That mutex only protects the members of the super class. Since this class adds its own members that need to be synchronized (since multiple threads could be accessing `_last_cmd_vel` and `_last_joint_power` at the same time, we need to add a mutex in this class to synchronize those accesses.

So, since we lock that mutex and modify the fields in `stop()`, we should correspondingly lock the mutex in `start()`, which ensures that `start()` can't be called while `stop()` is in the middle of executing. Otherwise, that could cause some invalid states.

src/network/MissionControlProtocol.h
```diff
@@ -29,51 +31,26 @@ class MissionControlProtocol : public WebSocketProtocol { // TODO: add documenta
...
+
 	void handleEmergencyStopRequest(const json& j);
 	void handleOperationModeRequest(const json& j);
 	void handleCameraStreamOpenRequest(const json& j);
 	void handleCameraStreamCloseRequest(const json& j);
 	void handleJointPowerRequest(const json& j);
 	void handleDriveRequest(const json& j);
 	void handleRequestArmIKEnabled(const json& j);
-	void sendArmIKEnabledReport();
-	void sendCameraStreamReport(const CameraID& cam,
```
huttongrabiel, line 68: How come we are getting rid of `sendCameraStreamReport`?

src/network/MissionControlMessages.h
```diff
@@ -0,0 +1,27 @@
...
+constexpr const char* CAMERA_STREAM_CLOSE_REQ_TYPE = "cameraStreamCloseRequest";
+
+// report keys
+constexpr const char* MOTOR_STATUS_REP_TYPE = "motorStatusReport";
+constexpr const char* CAMERA_STREAM_REP_TYPE = "cameraStreamReport";
+constexpr const char* LIDAR_REP_TYPE = "lidarReport";
+constexpr const char* MOUNTED_PERIPHERAL_REP_TYPE = "mountedPeripheralReport";
+constexpr const char* JOINT_POSITION_REP_TYPE = "jointPositionReport";
+constexpr const char* ROVER_POS_REP_TYPE = "roverPositionReport";
+constexpr const char* ARM_IK_ENABLED_REP_TYPE = "armIKEnabledReport";
```
QuinnP2910, line 25: MOTOR_POWER_REQ_TYPE, MOTOR_POSITION_REQ_TYPE, MOTOR_STATUS_REP_TYPE, and LIDAR_REP_TYPE are all unused. I know there's a TODO to implement motor position requests, but are the others necessary?

QuinnP2910 (commented):
Looks good, just a few questions.

src/network/MissionControlProtocol.cpp
```diff
@@ -2,60 +2,30 @@
...
-// report keys
-constexpr const char* MOTOR_STATUS_REP_TYPE = "motorStatusReport";
-constexpr const char* CAMERA_STREAM_REP_TYPE = "cameraStreamReport";
-constexpr const char* LIDAR_REP_TYPE = "lidarReport";
-constexpr const char* MOUNTED_PERIPHERAL_REP_TYPE = "mountedPeripheralReport";
-constexpr const char* JOINT_POSITION_REP_TYPE = "jointPositionReport";
-constexpr const char* ROVER_POS_REP_TYPE = "roverPositionReport";
-constexpr const char* ARM_IK_ENABLED_REP_TYPE = "armIKEnabledReport";
-// TODO: add support for missing report types
-// autonomousPlannedPathReport, poseConfidenceReport
```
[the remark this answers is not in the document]
abhaybd: Yeah, maybe that should have been in a different PR but we don't care about those reports. We should remove them from the MCP readme in the mc repo.

src/network/MissionControlMessages.h
abhaybd: That's a good point, we should probably re-evaluate which we care about and remove the others. I think we can maybe tackle that in another PR though, thoughts?

QuinnP2910: Sure. For the ones I mentioned other than motor pos, it should be as simple as just deleting the line in this file. But it's all good if you want that to be in some other MCP cleanup PR with the readme changes from the other comment.

src/network/MissionControlProtocol.cpp
```diff
@@ -403,137 +277,30 @@ MissionControlProtocol::MissionControlProtocol(SingleClientWSServer& server)
...
-			// break out of the loop if we should stop streaming
-			if (!this->_streaming_running) {
-				break;
-			}
-		}
-		std::this_thread::yield();
-	}
+void MissionControlProtocol::setRequestedJointPower(jointid_t joint, double power) {
+	_power_repeat_task.setJointPower(joint, power);
+	robot::setJointPower(joint, power);
```
quinnmp, line 303: Doesn't the `periodicTask` of the `PowerRepeatTask` already call this method?

src/network/MissionControlMessages.h
quinnmp, line 1: Maybe mark the currently unimplemented packet types with a `TODO` or something, just to be clear

quinnmp (commented):
Looks good, just a question and comment.

src/network/MissionControlProtocol.cpp
abhaybd: Yeah, but it's delayed by up to one loop period. We want the joint power to be set immediately, to avoid any lag.

src/network/MissionControlMessages.h
abhaybd: Done

quinnmp (approved):
Looks good

PR #308, opened 2024-02-12
Reset the arm IK setpoint when the velocity is 0

quinnmp:


src/control/PlanarArmController.h
```diff
@@ -85,12 +86,19 @@ class PlanarArmController {
 	 * @param targetVel The target x velocity.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_x_vel(robot::types::datatime_t currTime, double targetVel,
+				   robot::types::DataPoint<navtypes::Vectord<Constants::arm::IK_MOTORS.size()>>
```
abhaybd, line 90: The classes in the `control` folder (and most other folders) should be completely modular, so they should minimize dependencies on global variables. In this case, we have `N` as a template parameter for the class, which is the number of joints, which we should use instead of `Constants::arm::IK_MOTORS.size()`, so that should be replaced with `N`, and the include of `Constants.h` should be removed

src/control/PlanarArmController.h
```diff
@@ -104,12 +112,19 @@ class PlanarArmController {
 	 * @param targetVel The target y velocity.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_y_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_y_vel(robot::types::datatime_t currTime, double targetVel,
+				   robot::types::DataPoint<navtypes::Vectord<Constants::arm::IK_MOTORS.size()>>
```
abhaybd, line 116: Same comment as `set_x_vel`

abhaybd (dismissed):
Overall looks good, just some notes about modularity and code dependency

abhaybd (commented):
This looks good, can you just make sure you're up to date with master and resolve any merge conflicts?

src/control/PlanarArmController.h
```diff
@@ -85,12 +85,18 @@ class PlanarArmController {
 	 * @param targetVel The target x velocity.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_x_vel(robot::types::datatime_t currTime, double targetVel,
+				   robot::types::DataPoint<navtypes::Vectord<N>> jointPos) {
```
abhaybd, line 89: Why do we need it to be a `DataPoint`? Seems like we immediately just call `getData()`, so we might as well make this parameter a `const navtypes::Vectord<N>&`.

src/control/PlanarArmController.h
```diff
@@ -85,12 +85,18 @@ class PlanarArmController {
 	 * @param targetVel The target x velocity.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_x_vel(robot::types::datatime_t currTime, double targetVel,
```
abhaybd, line 88: We should update the doxy docs to reflect the new parameter

src/control/PlanarArmController.h
```diff
@@ -104,12 +110,18 @@ class PlanarArmController {
 	 * @param targetVel The target y velocity.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_y_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_y_vel(robot::types::datatime_t currTime, double targetVel,
+				   robot::types::DataPoint<navtypes::Vectord<N>> jointPos) {
```
abhaybd, line 114: Same as above, a) we should update the doxy docs and b) don't think there's a point in making this a `DataPoint`.

src/control/PlanarArmController.h
```diff
@@ -85,12 +85,18 @@ class PlanarArmController {
...
 	 */
-	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_x_vel(robot::types::datatime_t currTime, double targetVel,
+				   robot::types::DataPoint<navtypes::Vectord<N>> jointPos) {
 		std::lock_guard<std::mutex> lock(mutex);
 		if (velTimestamp.has_value()) {
-			double dt = util::durationToSec(currTime - velTimestamp.value());
-			// bounds check (new pos + vel vector <= sum of joint lengths)
-			setpoint = normalizeEEWithinRadius(setpoint + velocity * dt);
+			if (targetVel == 0.0 && velocity(1) == 0.0) {
```
abhaybd, line 92: We should also add a quick comment on why we do this, probably

abhaybd (dismissed):
Functionality-wise looks great, some design/cleanliness fixes we should look at though.

src/control/PlanarArmController.h
```diff
@@ -139,15 +139,25 @@ class PlanarArmController {
 	 *
 	 * @param currTime The current timestamp.
 	 * @param targetVel The target x velocity.
+	 * @param jointPos The joint angles of the arm.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_x_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_x_vel(robot::types::datatime_t currTime, double targetVel,
+				   const navtypes::Vectord<N> jointPos) {
```
abhaybd, line 146: Should be const ref

src/control/PlanarArmController.h
```diff
@@ -159,15 +169,25 @@ class PlanarArmController {
 	 *
 	 * @param currTime The current timestamp.
 	 * @param targetVel The target y velocity.
+	 * @param jointPos The joint angles of the arm.
 	 * @return The new command, which is the new joint positions.
 	 */
-	void set_y_vel(robot::types::datatime_t currTime, double targetVel) {
+	void set_y_vel(robot::types::datatime_t currTime, double targetVel,
+				   navtypes::Vectord<N> jointPos) {
```
abhaybd, line 176: Should be const ref

src/control/PlanarArmController.h
```diff
@@ -139,15 +139,25 @@ class PlanarArmController {
 	 *
 	 * @param currTime The current timestamp.
 	 * @param targetVel The target x velocity.
+	 * @param jointPos The joint angles of the arm.
```
abhaybd, line 142: Kinda nitpicky, but can you specify that this is the *current* joint angles? Both here and the set_y_vel method

abhaybd (changes requested):
Looks good! Just super minor changes needed, I think you just forgot to mark the params as const ref

abhaybd (approved):
Looks good!

PR #310, opened 2024-03-02
Remove vestigial estimators

abhaybd:
These estimators are technically correct but a) we don't use them and b) it's not very useful to fuse the kinematic model since the terrain is so shit anyway, and the kinematic parameters are likely misspecified anyway.


Geeoon (commented):
Looks good

quinnmp (approved):
Makes sense

PR #311, opened 2024-04-09
Refactor arm kinematics

abhaybd:
Now it's much more general to allow for different FK and IK algorithms, as well as non-planar kinematics. For convenience, all methods are wrapped in a general `ArmKinematics` class responsible for kinematic transformations.


Geeoon (commented):
I didn't notice anything besides what might be some commenting issues.
Looks good otherwise!

Geeoon (approved):
Looks good

PR #312, opened 2024-04-10
Swerve drive modes

quinnmp:
This is my basic implementation and is **definitely incorrect**, specifically in `kinematic_common_interface.cpp`. Also, there are not yet requests to change wheel angles.


src/world_interface/kinematic_common_interface.cpp
```diff
@@ -77,6 +84,100 @@ double setCmdVel(double dtheta, double dx) {
...
+
+	setMotorPower(motorid_t::frontLeftWheel, lfPWM);
+	setMotorPower(motorid_t::rearLeftWheel, lbPWM);
+	setMotorPower(motorid_t::frontRightWheel, rfPWM);
+	setMotorPower(motorid_t::rearRightWheel, rbPWM);
+
+	return maxAbsPWM > 1 ? maxAbsPWM : 1.0;
+}
+
+double setCrabCmdVel(double dtheta, double dy) {
```
quinnmp, line 149: This method does not handle rotation correctly at the moment. I wasn't sure how to, because we want this to essentially act as the normal drive mode but with a rotated reference frame

src/Constants.h
```diff
@@ -94,6 +94,17 @@ constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
...
 
+constexpr std::array<robot::types::motorid_t, 4> WHEEL_IDS = {
+	robot::types::motorid_t::frontLeftWheel,
+	robot::types::motorid_t::frontRightWheel,
+	robot::types::motorid_t::rearLeftWheel,
+	robot::types::motorid_t::rearRightWheel,
+};
+constexpr std::array<int32_t, 4> NORMAL_WHEEL_ROTS = {0, 0, 0, 0};
+constexpr std::array<int32_t, 4> TURN_IN_PLACE_WHEEL_ROTS = {45000, -45000, 45000, -45000};
+constexpr std::array<int32_t, 4> CRAB_WHEEL_ROTS = {90000, 90000, 90000, 90000};
```
abhaybd, line 105: These should be calculated from the swerve kinematics, not hardcoded, since the angles will change based on the wheelbase dimensions.

src/network/MissionControlProtocol.cpp
```diff
@@ -76,10 +76,73 @@ void MissionControlProtocol::handleOperationModeRequest(const json& j) {
...
+		Globals::driveMode = DriveMode::TurnInPlace;
+		for (int i = 1; i <= 4; i++) {
+			robot::setMotorPos(Constants::WHEEL_IDS[i],
+							   Constants::TURN_IN_PLACE_WHEEL_ROTS[i]);
+		}
+	} else if (mode == "crab") {
+		Globals::driveMode = DriveMode::Crab;
+		for (int i = 1; i <= 4; i++) {
+			robot::setMotorPos(Constants::WHEEL_IDS[i], Constants::CRAB_WHEEL_ROTS[i]);
+		}
```
abhaybd, line 101: I think you mean `for (int i = 0; i < 4; i++)`

src/network/MissionControlProtocol.cpp
```diff
@@ -76,10 +76,73 @@ void MissionControlProtocol::handleOperationModeRequest(const json& j) {
...
+		LOG_F(WARNING, "Drive mode set to %s, not Normal",
+			  driveModeStrings.at(Globals::driveMode).c_str());
+	}
+	return Globals::driveMode == DriveMode::Normal && util::hasKey(j, "straight") &&
+		   util::validateRange(j, "straight", -1, 1) && util::hasKey(j, "steer") &&
+		   util::validateRange(j, "steer", -1, 1);
+}
+
+static bool validateTankDriveRequest(const json& j) {
+	std::string msg = j.dump();
```
abhaybd, line 117: I know we do this in a lot of places, but why do we `dump()` and then don't use `msg`?

src/network/MissionControlProtocol.cpp
```diff
@@ -100,6 +163,33 @@ void MissionControlProtocol::handleDriveRequest(const json& j) {
...
+	double steer = j["steer"].get<double>();
+	double dtheta = Constants::MAX_DTHETA * steer;
+	LOG_F(1, "{steer=%.2f} -> setTurnInPlaceCmdVel(%.4f)", steer, dtheta);
+	this->setRequestedTurnInPlaceCmdVel(dtheta);
+}
+
+void MissionControlProtocol::handleCrabDriveRequest(const json& j) {
+	double crab = j["crab"];
+	double steer = -j["steer"].get<double>();
+	double norm = std::sqrt(std::pow(crab, 2) + std::pow(steer, 2));
```
abhaybd, line 186: Super minor, but `std::hypot` is more concise

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -59,6 +60,12 @@ double setCmdVel(double dtheta, double dx) {
 		return 0;
 	}
 
+	for (int i = 1; i <= 4; i++) {
```
abhaybd, line 63: Same as above, 0-indexing

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -77,6 +84,100 @@ double setCmdVel(double dtheta, double dx) {
...
+
+	return maxAbsPWM > 1 ? maxAbsPWM : 1.0;
+}
+
+double setTurnInPlaceCmdVel(double dtheta) {
+	if (Globals::E_STOP && (dtheta != 0)) {
+		return 0;
+	}
+
+	for (int i = 1; i <= 4; i++) {
```
abhaybd, line 121: Index out of bounds

abhaybd (commented):
Brief review, didn't get a chance to look at everything yet

src/network/MissionControlProtocol.cpp
quinnmp: I actually don't know - it was done in `validateDriveRequest` and so I emulated it in the others. I'll remove them for now and revert if needed

src/Constants.h
```diff
@@ -95,6 +96,16 @@ constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
...
 
+constexpr std::array<robot::types::motorid_t, 4> WHEEL_IDS = {
+	robot::types::motorid_t::frontLeftWheel,
+	robot::types::motorid_t::frontRightWheel,
+	robot::types::motorid_t::rearLeftWheel,
+	robot::types::motorid_t::rearRightWheel,
+};
+
+const extern std::unordered_map<DriveMode, std::array<int32_t, 4>> WHEEL_ROTS;
+constexpr double STEER_EPSILON = 1000;
```
abhaybd, line 107: Can we put this in a nested namespace, so maybe it would be `Constants::Drive` or something?

src/Globals.h
```diff
@@ -16,13 +16,25 @@ enum class RoverState {
...
+	{DriveMode::Crab, "Crab"}};
+
 // NOTE(sasha): To keep linker happy, declare globals with "extern" here and then
 //              provide variable definition in Globals.cpp
 namespace Globals {
 /**
    @deprecated
  */
 extern RoverState curr_state;
+extern std::tuple<DriveMode, bool> driveMode;
```
abhaybd, line 37: Can we add documentation on what the parts of the drive mode are? Particularly the `bool`, which is vague

abhaybd, line 37: Also, since there's just two, we should probably just use `std::pair`

src/Constants.h
```diff
@@ -95,6 +96,16 @@ constexpr const char* ARDUPILOT_PROTOCOL_NAME = "/ardupilot";
 constexpr std::chrono::milliseconds JOINT_POWER_REPEAT_PERIOD(333);
 constexpr std::chrono::milliseconds ARM_IK_UPDATE_PERIOD(50);
 
+constexpr std::array<robot::types::motorid_t, 4> WHEEL_IDS = {
+	robot::types::motorid_t::frontLeftWheel,
+	robot::types::motorid_t::frontRightWheel,
+	robot::types::motorid_t::rearLeftWheel,
+	robot::types::motorid_t::rearRightWheel,
```
abhaybd, line 103: Just a note, these motors are the actual wheels, we should add `motorid_t` values for the steer motors. I think Victor should handle that in his PR, though, so wait for that and you can merge it in

src/Constants.cpp
```diff
@@ -0,0 +1,20 @@
+#include "Constants.h"
+
+#include "Globals.h"
+
+namespace Constants {
+constexpr std::array<int32_t, 4> NORMAL_WHEEL_ROTS = {0, 0, 0, 0};
+const swervewheelvel_t TURN_IN_PLACE_VEL =
+	kinematics::SwerveDriveKinematics(Constants::WHEEL_BASE, Constants::ROBOT_LENGTH)
```
abhaybd, line 8: We should probably reuse `robot::swerveKinematics()` instead of re-constructing the swerve kinematics

abhaybd (changes requested):
Pretty good! Just some notes, and we should be able to start testing this after Victor's pr

src/Globals.cpp
```diff
@@ -37,6 +37,11 @@ net::websocket::SingleClientWSServer websocketServer("DefaultServer",
 													 Constants::WS_SERVER_PORT);
 std::atomic<bool> E_STOP = false;
 std::atomic<bool> AUTONOMOUS = false;
+
+// The boolean here is whether or not wheel rotation check should be ignored when doing
+// swerve-based driving
+std::tuple<DriveMode, bool> driveMode = {DriveMode::Normal, false};
```
huttongrabiel, line 43: Same thing here as Abhay said. `std::pair` instead of `std::tuple` if you make that change in `Globals.h`.

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -77,6 +86,106 @@ double setCmdVel(double dtheta, double dx) {
...
+	}
+
+	if (!std::get<1>(Globals::driveMode)) {
+		for (int i = 0; i < 4; i++) {
+			if (std::abs(robot::getMotorPos(Constants::WHEEL_IDS[i]) -
+						 Constants::WHEEL_ROTS.at(DriveMode::TurnInPlace)[i]) <
+				Constants::STEER_EPSILON)
+				return 0;
+		}
+	}
```
huttongrabiel, line 133: Is there an easy way to factor this out into a little function? It's used in 3 places throughout. No big deal if not

src/world_interface/world_interface.h
```diff
@@ -71,6 +78,36 @@ void emergencyStop();
...
+ *
+ * @param dtheta The heading velocity.
+ * @return double If the requested velocities are too high, they will be scaled down.
+ * The returned value is the scale divisor. If no scaling was performed, 1 is returned.
+ */
+double setTurnInPlaceCmdVel(double dtheta);
+
+/**
+ * @brief Request the robot to drive side to side and turn with given velocities using swerve.
+ * Essentially rotate the reference from of the robot by 90 degrees.
```
huttongrabiel, line 102: Did you mean frame instead of from?

huttongrabiel (changes requested):

src/Constants.h
```diff
@@ -9,6 +9,12 @@
 
 #include <frozen/unordered_map.h>
 
+namespace control {
+enum class DriveMode;
+}
+
+using control::DriveMode;
```
abhaybd, line 16: Looks unused. Also btw, `using` declarations in header files should usually be avoided, since it causes namespace pollution.

src/Globals.cpp
```diff
@@ -10,6 +10,7 @@
 
 using robot::types::motorid_t;
 using namespace Constants::arm;
+using control::DriveMode;
```
abhaybd, line 13: Unused?

src/Globals.h
```diff
@@ -17,6 +18,13 @@ enum class RoverState {
 	LookingForTennisball,
 };
 
+namespace control {
+enum class DriveMode;
+class SwerveController;
+} // namespace control
+
+using control::DriveMode;
```
abhaybd, line 26: Since we include the header file, we don't need the forward declaration. Also the other stuff is unused?

src/control/SwerveController.cpp
```diff
@@ -0,0 +1,84 @@
...
+			{static_cast<int32_t>(swerve_kinematics.robotVelToWheelVel(0, 0, 1).lfRot * 1000),
+			 static_cast<int32_t>(swerve_kinematics.robotVelToWheelVel(0, 0, 1).rfRot * 1000),
+			 static_cast<int32_t>(swerve_kinematics.robotVelToWheelVel(0, 0, 1).lbRot * 1000),
+			 static_cast<int32_t>(swerve_kinematics.robotVelToWheelVel(0, 0, 1).rbRot *
+								  1000)}},
+		   {DriveMode::Crab, {90000, 90000, 90000, 90000}}}) {}
+
+std::pair<double, swerve_commands_t>
+SwerveController::setTurnInPlaceCmdVel(double dtheta, swerve_rots_t wheel_rots) {
+	if (dtheta != 0) {
```
abhaybd, line 24: Do you mean `dtheta == 0`?

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+	 *
+	 * @param mode The drive mode to compare wheel rotations against.
+	 * @param wheel_rots The rotation of all four wheels in millidegrees
+	 * @return bool if the wheels are within `Constants::Drive::STEER_EPSILON` of their their
+	 * target position.
+	 */
+	bool checkWheelRotation(DriveMode mode, swerve_rots_t wheel_rots);
+
+	// The boolean here is whether or not wheel rotation check should be ignored when doing
+	// swerve-based driving
```
abhaybd, line 78: Can we make this doxy-style comments

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+enum class DriveMode {
+	Normal,
+	TurnInPlace,
+	Crab,
+};
+
+static const std::map<DriveMode, std::string> driveModeStrings = {
+	{DriveMode::Normal, "Normal"},
+	{DriveMode::TurnInPlace, "TurnInPlace"},
+	{DriveMode::Crab, "Crab"}};
```
abhaybd, line 18: Instead of a global map, I think it would be a bit cleaner to provide an overload of the [`to_string()`](https://github.com/huskyroboticsteam/Resurgence/blob/master/src/utils/core.h#L50-L53) function.

To do that, do something like
```c++
namespace util {
std::string to_string(control::DriveMode mode);
}
```

See an example [here](https://github.com/huskyroboticsteam/Resurgence/blob/master/src/world_interface/data.h#L305-L307)

PS: `static` globals and methods in header-files are kind of a code smell. It will just be compiled separately everywhere the header is included, and all it does is prevent a [ODR](https://en.cppreference.com/w/cpp/language/definition) violation. If we wanted to put this map here, we should declare it as `extern const` and provide the definition in `SwerveController.cpp`

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+	int lbRot;
+	int rbRot;
+};
+
+// The PWM commands to all four steer motors.
+struct swerve_commands_t {
+	double lfPWM;
+	double rfPWM;
+	double lbPWM;
+	double rbPWM;
```
abhaybd, line 33: Let's use `power` instead of `pwm`, since the exact method of setting the power is an implementation detail. For example, BLDC motors don't even use PWM.

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+	 * @param dy The side-to-side velocity.
+	 * @param wheel_rots The rotation of all four wheels in millidegrees
+	 * @return std::pair<double, swerve_commands_t> First: If the requested velocities are too
+	 * high, they will be scaled down. The returned value is the scale divisor. If no scaling
+	 * was performed, 1 is returned. Second: The PWM commands for all four wheels.
+	 */
+	std::pair<double, swerve_commands_t> setCrabCmdVel(double dtheta, double dy,
+													   swerve_rots_t wheels);
+
+	kinematics::SwerveDriveKinematics swerveKinematics();
```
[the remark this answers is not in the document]
abhaybd: Also, this (and most methods in this class) should be marked as `const`

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+	 * @param mode The drive mode to compare wheel rotations against.
+	 * @param wheel_rots The rotation of all four wheels in millidegrees
+	 * @return bool if the wheels are within `Constants::Drive::STEER_EPSILON` of their their
+	 * target position.
+	 */
+	bool checkWheelRotation(DriveMode mode, swerve_rots_t wheel_rots);
+
+	// The boolean here is whether or not wheel rotation check should be ignored when doing
+	// swerve-based driving
+	std::pair<DriveMode, bool> driveMode;
```
abhaybd, line 79: Now that this is encapsulated in a class, it lessens its readability to package this in a pair. We can just separate these out into class fields. Also, public fields are ugly and kind of leak implementation, so maybe make this private and add appropriate getters and setters.

abhaybd: Also, maybe the `setDriveMode` method can also return the corresponding `swerve_rots_t`, which makes client code a little more elegant

src/control/SwerveController.h
```diff
@@ -0,0 +1,88 @@
...
+
+	const std::unordered_map<DriveMode, std::array<int32_t, 4>> WHEEL_ROTS;
```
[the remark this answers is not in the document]
abhaybd: Also to be more in line with the rest of the class, the type should probably be `swerve_rots_t`

src/network/MissionControlTasks.cpp
```diff
@@ -60,7 +79,42 @@ void PowerRepeatTask::periodicTask() {
...
+					robot::getMotorPos(motorid_t::rearRightWheel)};
+				control::swerve_commands_t steer_PWM =
+					Globals::swerveController
+						.setCrabCmdVel(_last_cmd_vel->first, _last_cmd_vel->second,
+									   curr_wheel_rots)
+						.second;
+				robot::setMotorPower(motorid_t::frontLeftWheel, steer_PWM.lfPWM);
+				robot::setMotorPower(motorid_t::frontRightWheel, steer_PWM.rfPWM);
+				robot::setMotorPower(motorid_t::rearLeftWheel, steer_PWM.lbPWM);
+				robot::setMotorPower(motorid_t::rearRightWheel, steer_PWM.rbPWM);
```
abhaybd, line 116: This is a lot of duplicated code, maybe factor the common parts out?

src/world_interface/world_interface.h
```diff
@@ -18,6 +18,10 @@ namespace cam {
 class CameraParams;
 }
 
+namespace control {
+enum class DriveMode;
+}
```
abhaybd, line 23: Unused?

src/world_interface/world_interface.h
```diff
@@ -71,6 +80,16 @@ void emergencyStop();
...
 
+/**
+ * @brief Request the robot to drive in tank style, where each side is controlled individually.
+ *
+ * @param left The left velocity.
+ * @param right The right velocity.
+ * @return double If the requested velocities are too high, they will be scaled down.
+ * The returned value is the scale divisor. If no scaling was performed, 1 is returned.
+ */
+double setTankCmdVel(double dtheta, double dx);
```
abhaybd, line 91: Parameter names should be `left`, `right`

src/Globals.cpp
```diff
@@ -49,5 +50,6 @@ std::atomic<bool> AUTONOMOUS = false;
 robot::types::mountedperipheral_t mountedPeripheral = robot::types::mountedperipheral_t::none;
 control::PlanarArmController<2> planarArmController(createArmKinematics(),
 													Constants::arm::SAFETY_FACTOR);
+control::SwerveController swerveController(Constants::EFF_WHEEL_BASE, Constants::ROBOT_LENGTH);
```
abhaybd, line 53: /We should use the real wheel base instead of the effective wheel base. If `ROBOT_WIDTH` isn't already a constant, we should add it

src/CMakeLists.txt
```diff
@@ -152,6 +152,7 @@ add_library(utils SHARED
   kinematics/DiffDriveKinematics.cpp
   kinematics/DiffWristKinematics.cpp
   kinematics/SwerveDriveKinematics.cpp
+  control/SwerveController.cpp
```
abhaybd, line 155: This should probably be added to the `control` target instead of `utils`

abhaybd (changes requested):
Great job on this! I admittedly got a little carried away with the review comments, but this is a nice first draft, and the core functionality and design is all great, it's just cleanliness and QoL fixes at this point.

src/Constants.h
```diff
@@ -95,6 +90,7 @@ constexpr std::array<robot::types::motorid_t, 4> WHEEL_IDS = {
 	robot::types::motorid_t::rearLeftWheel,
 	robot::types::motorid_t::rearRightWheel,
 };
+const double STEER_EPSILON = 1000;
```
abhaybd, line 93: This should be defined as `extern`, in keeping with everything else. Also, can you document what this constant is? (what it is, what the units are, etc)

src/control/SwerveController.cpp
```diff
@@ -6,54 +6,72 @@
...
+	swerve_commands_t commands = setCrabCmdVel(dtheta, dy, wheel_rots, dummyScaleFactor);
+	return commands;
 }
 
-std::pair<double, swerve_commands_t>
-SwerveController::setCrabCmdVel(double dtheta, double dy, swerve_rots_t wheel_rots) {
+swerve_commands_t SwerveController::setCrabCmdVel(double dtheta, double dy,
+												  const swerve_rots_t& wheel_rots,
+												  double& scaleFactor) const {
 	if (Globals::E_STOP && (dtheta != 0 || dy != 0)) {
```
abhaybd, line 66: No point in checking the `E_STOP` value, since this method isn't responsible for policing that. Also means we can cut the dependency on globals

abhaybd: As an aside, we can just cut this whole special case, since if the inputs are zero then the outputs will be zero.

src/network/MissionControlProtocol.cpp
```diff
@@ -88,65 +97,57 @@ static bool validateDriveModeRequest(const json& j) {
...
+		target = swerveController.setDriveMode(DriveMode::TurnInPlace);
 	} else if (mode == "crab") {
-		swerveController.driveMode.first = DriveMode::Crab;
-		for (int i = 0; i < 4; i++) {
-			robot::setMotorPos(Constants::Drive::WHEEL_IDS[i],
-							   Globals::swerveController.WHEEL_ROTS.at(DriveMode::Crab)[i]);
-		}
+		target = swerveController.setDriveMode(DriveMode::Crab);
+	}
+	std::array<int, 4> target_rots = {target.lfRot, target.rfRot, target.lbRot, target.rbRot};
```
abhaybd, line 108: Can we log a warning and no-op if `target` is uninitialized? i.e. if the mode was the wrong value. In theory this shouldn't happen bc of the validator, but a buggy validator shouldn't be able to damage the hardware if some uninitialized value is passed as the pid target.

src/network/MissionControlProtocol.cpp
```diff
@@ -32,6 +35,16 @@ const std::chrono::milliseconds HEARTBEAT_TIMEOUT_PERIOD = 3000ms;
...
 
+/**
+ * Get the steer rotations of all wheels as a swerve_rots_t.
+ */
+static control::swerve_rots_t getAllSteerRotations();
+
+/**
+ * Set the steer power of all wheels with a swerve_commands_t.
+ */
+static void setAllSteerPowers(control::swerve_commands_t commands);
```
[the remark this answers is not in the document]
abhaybd: Makes sense, what if we just changed the methods of the power repeat task to also execute it now? That way we can just call `setTurnInPlaceCmdVel` or whatever and it will both schedule the power repeat *and* send the command now. That way the utility functions can only exist in MCPTasks

src/network/MissionControlTasks.cpp
```diff
@@ -22,6 +22,16 @@ namespace {
...
 
+/**
+ * Get the steer rotations of all wheels as a swerve_rots_t.
+ */
+static control::swerve_rots_t getAllSteerRotations();
+
+/**
+ * Set the steer power of all wheels with a swerve_commands_t.
+ */
+static void setAllSteerPowers(control::swerve_commands_t commands);
```
abhaybd, line 33: Pass structs by const ref!

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -36,6 +38,14 @@ double setCmdVel(double dtheta, double dx) {
 		return 0;
 	}
 
+	control::swerve_rots_t curr_wheel_rots = {
+		robot::getMotorPos(motorid_t::frontLeftWheel).getData(),
+		robot::getMotorPos(motorid_t::frontRightWheel).getData(),
+		robot::getMotorPos(motorid_t::rearLeftWheel).getData(),
+		robot::getMotorPos(motorid_t::rearRightWheel).getData()};
```
abhaybd, line 45: We should handle the case where any of these datapoints are invalid, otherwise it will crash if we haven't gotten an encoder packet yet. I added a PR (#324) that should make things easier so we can just add a try-catch around the whole block.

abhaybd: Note that this should be done everywhere else we do a `getData()` without checking if the datapoint is valid.

abhaybd (changes requested):
Looking better, just some notes and some minor stuff I missed before.

src/network/MissionControlTasks.cpp
```diff
@@ -185,4 +242,23 @@ void ArmIKTask::updateArmIK() {
...
 }
 
+static DataPoint<control::swerve_rots_t> getAllSteerRotations() {
+	try {
+		return DataPoint<control::swerve_rots_t>(
+			{robot::getMotorPos(motorid_t::frontLeftWheel).getData(),
+			 robot::getMotorPos(motorid_t::frontRightWheel).getData(),
+			 robot::getMotorPos(motorid_t::rearLeftWheel).getData(),
+			 robot::getMotorPos(motorid_t::rearRightWheel).getData()});
+	} catch (const std::runtime_error& e) {
```
quinnmp, line 252: Should this be  `bad_datapoint_access` instead of `std::runtime_error` once that PR is merged?

abhaybd: Yessir

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -38,11 +38,16 @@ double setCmdVel(double dtheta, double dx) {
...
-		robot::getMotorPos(motorid_t::frontRightWheel).getData(),
-		robot::getMotorPos(motorid_t::rearLeftWheel).getData(),
-		robot::getMotorPos(motorid_t::rearRightWheel).getData()};
+	control::swerve_rots_t curr_wheel_rots;
+	try {
+		curr_wheel_rots = {robot::getMotorPos(motorid_t::frontLeftWheel).getData(),
+						   robot::getMotorPos(motorid_t::frontRightWheel).getData(),
+						   robot::getMotorPos(motorid_t::rearLeftWheel).getData(),
+						   robot::getMotorPos(motorid_t::rearRightWheel).getData()};
+	} catch (const std::runtime_error& e) {
```
abhaybd, line 47: Same as above, remember to switch this to `bad_datapoint_access`

src/world_interface/kinematic_common_interface.cpp
```diff
@@ -68,11 +73,16 @@ double setTankCmdVel(double left, double right) {
...
```
abhaybd, line 82: Same as above you get the gist

abhaybd (commented):
Looks good! Let's get this tested asap (after setting up hardware) and then this should be good to merge

src/world_interface/world_interface.h
```diff
@@ -76,6 +81,16 @@ void emergencyStop();
...
 
+/**
+ * @brief Request the robot to drive in tank style, where each side is controlled individually.
+ *
+ * @param left The left velocity.
+ * @param right The right velocity.
+ * @return double If the requested velocities are too high, they will be scaled down.
+ * The returned value is the scale divisor. If no scaling was performed, 1 is returned.
+ */
+double setTankCmdVel(double left, double right);
```
abhaybd, line 92: This declaration should go in `control_interface.h` I think

abhaybd (changes requested):
This looks great! One minor nitpick but besides that we should be ready to test soon!

src/network/MissionControlProtocol.cpp
```diff
@@ -78,12 +81,73 @@ void MissionControlProtocol::handleOperationModeRequest(const json& j) {
...
+	} else if (mode == "turn-in-place") {
+		target = swerveController.setDriveMode(DriveMode::TurnInPlace);
+	} else if (mode == "crab") {
+		target = swerveController.setDriveMode(DriveMode::Crab);
+	} else {
+		LOG_F(WARNING, "Invalid requested Drive Mode");
+		return;
+	}
+
+	robot::setMotorPower(motorid_t::frontLeftSwerve, target.lfRot);
```
imisaacwu, line 104: These should be setMotorPos I think

quinnmp: Good catch!

abhaybd (approved):
Tested and approved! Great job on this!

PR #317, opened 2024-04-16
Unify CAN Motor implementation

DrDab:
Merge the functionality of `CANMotor_PSOC.cpp`  into `CANMotor.cpp` since `CANMotor_AVR.cpp` is no longer used


src/CAN/CANMotor.cpp
```diff
@@ -5,7 +5,11 @@
 
 #include <chrono>
 #include <cmath>
+#include <condition_variable>
+#include <mutex>
+#include <queue>
 #include <thread>
+#include <vector>
```
abhaybd, line 12: Are any of these new includes required? They seem like they might have been vestigial in `CANMotor_PSOC.cpp`, so we should probably remove them.

Note that we still want `<thread>`, so leave that in.

abhaybd (changes requested):
Just a note about includes, but looks good otherwise. Can you make sure it's up to date with master?

abhaybd (approved):
Looks good! Can you just ensure it's up to date with master, and then merge?

PR #320, opened 2024-04-29
Remove deprecated and vestigial methods from world interface

abhaybd:
We don't use or even implement these methods


huttongrabiel (approved):
lgtm

PR #321, opened 2024-04-30
Move constants definitions to impl file, add per-stream compression constants

abhaybd:
Moving definitions to a .cpp file makes tuning a lot faster, since only that file needs to be recompiled, otherwise editing the header files triggers a LOT of recompilation.

Besides that refactor, this PR adds a map allowing us to specify specific compression constants for streams, since we want the hand camera to be higher quality than the others.


huttongrabiel (approved):
lgtm

PR #323, opened 2024-05-12
Remove heading messages from ArdupilotProtocol

abhaybd:
Heading is encoded in the orientation, so no need for this.


huttongrabiel (approved):
Looks good!

PR #324, opened 2024-05-14
Improve DataPoint safety

abhaybd:
- Remove the implicit cast, which is probably a very easy way to introduce bugs
- Fix all usages to check if a datapoint is valid before accessing it
- Change behavior of accessing invalid datapoint from UB to throwing a specific exception


huttongrabiel (approved):
nice update. lgtm

huttongrabiel (approved):
lgtm

PR #326, opened 2024-05-24
Fix CMake circular dependencies

abhaybd:
Our CMake is becoming a bit of a mess, and circular dependencies were cropping up. This PR doesn't fully fix this problem, but is a partial solution. This PR focuses on the use of shared global state, particularly `Globals.h`, which had become way too pervasive in our codebase.
 - Moved E-Stop from globals to instead be owned by the world interface (which also makes more sense, since we should refuse to move motors if e-stopped) NOTE: there actually isn't code that checks for this, so that should be done in a follow-up PR
 - Moved `kinematic_common_interface` out of the world interface layer. It has been renamed to `control_interface`, and is not part of the world interface, it is only an abstraction around it.
 - Some refactoring to remove dependencies on global state, including creating a separate `constants` build target for use.

Overall, we need to focus on the distinction between "application code" and "library code". Library code should be the core functionality and modules of our codebase, which is instrumented and used by the application code. In general, library code should **not** use shared global state, while application code can.

### DIFF
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -3,6 +3,7 @@
 #include "../Constants.h"
 #include "../Globals.h"
 #include "../autonomous/AutonomousTask.h"
+#include "../control_interface.h"
 #include "../utils/core.h"
 #include "../utils/json.h"
 #include "../world_interface/data.h"
--- a/src/network/MissionControlProtocol.cpp
+++ b/src/network/MissionControlProtocol.cpp
@@ -56,7 +56,6 @@ void MissionControlProtocol::handleEmergencyStopRequest(const json& j) {
 		_power_repeat_task.start();
 	}
 	// TODO: reinit motors
-	Globals::E_STOP = stop;
 	this->setArmIKEnabled(false);
 }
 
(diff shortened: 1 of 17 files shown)


src/network/MissionControlProtocol.cpp
```diff
@@ -56,7 +57,6 @@ void MissionControlProtocol::handleEmergencyStopRequest(const json& j) {
 		_power_repeat_task.start();
```
quinnmp, line 57: Are we properly handling this case in which we leave e-stop?

quinnmp (commented):
Looks good, just a quick question and nitpick

src/network/MissionControlProtocol.cpp
abhaybd: Technically we can't leave e-stop, we're supposed to restart the rover code when we e-stop. (hence why there's no undo method for `robot::emergencyStop()`, so we should probably factor that into the logic here, but I think that's out of the scope of this PR, and should be done in the same PR that enforces the e-stop flag in the world interface.

quinnmp: Gotcha! Makes sense.

quinnmp (approved):

PR #331, opened 2024-07-18
Full-Resolution Camera Downloads

Geeoon:
Currently, clicking the download button will just download the client-side H264 compressed stream, which is fairly low quality.  This PR adds the ability to request full-resolution jpg images from the rover, and if the camera is available, a base64 encoded image will be sent to mission control at a much higher resolution than the normal client-side video stream.

Related with https://github.com/huskyroboticsteam/mission-control/pull/71


src/network/MissionControlTasks.cpp
```diff
@@ -141,6 +142,23 @@ void CameraStreamTask::closeStream(const CameraID& cam) {
...
+	std::string b64_data = "";
+	{
+		std::lock_guard lg(_mutex);
+		if (_open_streams.find(cam) != _open_streams.end()) {
+			auto camDP = robot::readCamera(cam);
+			if (camDP) {
+				auto data = camDP.getData();
+				cv::Mat frame = data.first;
+				b64_data = base64::encodeMat(frame, ".jpg");
+			}
```
abhaybd, line 155: If the `camDP` datapoint is empty, this will respond with an empty string. Instead, we should probably not send a response at all.

src/network/MissionControlTasks.cpp
```diff
@@ -141,6 +142,23 @@ void CameraStreamTask::closeStream(const CameraID& cam) {
 	_camera_encoders.erase(cam);
 }
 
+void CameraStreamTask::sendCurrentFrame(const robot::types::CameraID& cam) {
```
abhaybd, line 145: Why do we need this method? Why can't the callback handler just call `robot::readCamera()`? We don't necessarily need to check `_open_streams` to see if the given stream has been opened or not, we can just read the frame from the camera.

abhaybd:
This looks great, will merge once the corresponding MC pr has been tested and merged.

abhaybd:
Didn't get to this during CIRC, Hutton can handle review and merging.

src/network/MissionControlMessages.h
```diff
@@ -18,11 +18,13 @@ constexpr const char* MOTOR_POSITION_REQ_TYPE = "motorPositionRequest";
 constexpr const char* JOINT_POSITION_REQ_TYPE = "jointPositionRequest";
 constexpr const char* CAMERA_STREAM_OPEN_REQ_TYPE = "cameraStreamOpenRequest";
 constexpr const char* CAMERA_STREAM_CLOSE_REQ_TYPE = "cameraStreamCloseRequest";
+constexpr const char* CAMERA_FRAME_REQ_TYPE = "cameraFrameRequest";
```
huttongrabiel, line 21: Lets try to make these `frozen::string`

huttongrabiel (changes requested):
ignore this just a demo

huttongrabiel (approved):

imisaacwu (approved):
looks great!

PR #334, opened 2024-08-08
Update configurations for new jetson orin nx

abhaybd:
Reconfigure CAN to have the right datarate, update the CAN code to use can FD frames, update the camera udev rules to match the real camera device instead of the virtual metadata device (due to new video4linux kernel)


Asiadav (commented):
yeet my skeet. looks solid g

Asiadav (approved):

PR #335, opened 2024-08-14
DO NOT MERGE: Changes from CIRC competition

abhaybd:
This PR consists of hotfixes and changes implemented during the CIRC 2024 competition. As a result, much of this code consists of quick fixes/hacks that shouldn't be merged, but should rather be evaluated if we need it, and should be cleaned up and merged as a separate PR.

I will go through the code noting why sections of code were fixed, to help with the cleaning up process.


camera-config/PanoCameraCalibration.yml
```diff
@@ -0,0 +1,36 @@
...
+          6.4776804982533793e+02, 2.4785660889077488e+02, 0., 0., 1. ]
+   distortion_coefficients: !!opencv-matrix
+      rows: 5
+      cols: 1
+      dt: d
+      data: [ -4.5670452858419575e-01, 2.1512832208218641e-01,
+          -1.1179480533495888e-03, -5.7450430930239835e-04, 0. ]
+extrinsic_params: !!opencv-matrix
+  rows: 4
+  cols: 4
```
abhaybd, line 31: These are all unused. We should redo how we handle camera configuration, especially considering the changes in Camera.cpp

camera-config/DrillCameraCalibration.yml
```diff
@@ -0,0 +1,36 @@
...
+   distortion_coefficients: !!opencv-matrix
+      rows: 5
+      cols: 1
+      dt: d
+      data: [ -4.5670452858419575e-01, 2.1512832208218641e-01,
+          -1.1179480533495888e-03, -5.7450430930239835e-04, 0. ]
+extrinsic_params: !!opencv-matrix
+  rows: 4
+  cols: 4
+  dt: d
```
abhaybd, line 32: Same as other camera, unused. See note on other camera config file.

src/CAN/CANMotor.h
```diff
@@ -136,6 +136,8 @@ void setMotorPower(deviceserial_t serial, int16_t power);
  */
 void setMotorPIDTarget(deviceserial_t serial, int32_t target);
 
+void setMotorPIDMaxPower(deviceserial_t serial, uint16_t maxPower);
```
abhaybd, line 139: This method restricts the maximum power that the motor is allowed to apply when running PID. Add documentation! Also, maybe add an overload with a double parameter for power in the range `[0,1]`?

src/Constants.h
```diff
@@ -186,8 +183,8 @@ constexpr frozen::unordered_map<robot::types::motorid_t, std::pair<int, int>, IK
  * Map from motor ids to segment length in meters
  */
 constexpr frozen::unordered_map<robot::types::motorid_t, double, IK_MOTORS.size()>
-	SEGMENT_LENGTHS{{robot::types::motorid_t::shoulder, 0.3848608},
-					{robot::types::motorid_t::elbow, 0.461264}};
+	SEGMENT_LENGTHS{{robot::types::motorid_t::shoulder, 0.394},
+					{robot::types::motorid_t::elbow, 0.749}};
```
abhaybd, line 187: Note that the elbow measurement goes all the way from the elbow joint to the tip of the fingers of the hand.

src/TunePID.cpp
```diff
@@ -76,8 +77,11 @@ bool initMotor(motorid_t motor) {
 		robot::encparams_t enc_params = robot::encMotors.at(motor);
 		can::motor::initEncoder(serial, enc_params.isInverted, true, enc_params.ppjr,
 								robot::TELEM_PERIOD);
-		can::motor::setLimitSwitchLimits(serial, enc_params.limitSwitchLow,
-										 enc_params.limitSwitchHigh);
+		auto mDegToPulses = [&](int32_t mdeg) {
+			return static_cast<int32_t>((double)mdeg / 360000.0 * enc_params.ppjr);
+		};
+		can::motor::setLimitSwitchLimits(serial, mDegToPulses(enc_params.limitSwitchLow),
+										 mDegToPulses(enc_params.limitSwitchHigh));
```
abhaybd, line 84: The parameters to `setLimitSwitchLimits()` should be in encoder ticks, not mdegs. This used to work before, only because it was incorrect on both sides (firmware and software) but disagreed with documentation. This has been updated on both sides.

src/TunePID.cpp
```diff
@@ -132,13 +136,14 @@ int main(int argc, char** argv) {
 	std::getline(std::cin, str);
 	int d_coeff = std::stoi(str);
 
-	can::motor::setMotorMode(serial, can::motor::motormode_t::pid);
 	can::motor::setMotorPIDConstants(serial, p_coeff, i_coeff, d_coeff);
+	can::motor::setMotorPIDMaxPower(serial, 32767);
```
abhaybd, line 140: Hardcoded, probably remove.

src/TunePID.cpp
```diff
@@ -132,13 +136,14 @@ int main(int argc, char** argv) {
...
+	can::motor::setMotorMode(serial, can::motor::motormode_t::pid);
 
 	double period = 8.0;
 
 	std::this_thread::sleep_for(300ms); // wait for encoder position data to arrive
-	int32_t starting_angle = can::motor::getMotorPosition(serial);
+	int32_t starting_angle = can::motor::getMotorPosition(serial).getData();
```
abhaybd, line 146: This whole thing should probably be a while loop waiting until we get data (with a max timeout) instead of waiting for 300ms.

src/camera/Camera.cpp
```diff
@@ -23,7 +24,24 @@ bool Camera::open(int camera_id, CameraParams intrinsic_params, Mat extrinsic_pa
...
+	} else if (camera_id == 100) {
+		gstr_ss << "image/jpeg,width=1280,height=720,framerate=30/1 ! "
+				   "jpegdec ! videoflip method=rotate-180 ! videoconvert ! appsink";
+	} else if (camera_id == 110) {
+		gstr_ss << "image/jpeg,width=960,height=720,framerate=30/1 ! "
+				   "jpegdec ! videoflip method=rotate-180 ! videoconvert ! appsink";
+	}
+	LOG_F(INFO, "GSTR: %s", gstr_ss.str().c_str());
+	this->_capture = std::make_shared<cv::VideoCapture>(gstr_ss.str(), cv::CAP_GSTREAMER);
+	bool result = this->_capture->isOpened();
```
abhaybd, line 44: OpenCV on the jetson uses the GStreamer backend by default, which does not respect some parameter configurations such as `fourcc`, which is used to set the encoding. (So it was defaulting to yuyv instead of mjpg, which is what we were setting it to)

This code constructs [gstreamer pipelines](https://gstreamer.freedesktop.org/documentation/tools/gst-launch.html?gi-language=c) ([this ](https://gist.github.com/hum4n0id/cda96fb07a34300cdb2c0e314c14df0a) might be of some use) which set the resolution, framerate, and encoding as necessary, which are actually respected by GStreamer.

Note that you have to set the resolution/framerate/encoding to a combination actually supported by the device, which you can find by following [this](https://askubuntu.com/a/687140/1664152).

Observe how we can also do basic transformations, such as rotating the science cameras (id 100 and 110) by 180 degrees, since those cameras were mounted upside down. We can also do stuff like downsample framerate (which could be useful, for example, if a stream is available at 60fps but we only want 30fps)

The camera config files should be overhauled to support this format of specification, which is a lot better imo. I think we should also probably get rid of the `Camera::open()` method that accepts a filename, since it adds extra maintenance overhead and I don't think we'll be using it.

src/camera/Camera.cpp
```diff
@@ -37,6 +55,9 @@ bool Camera::open(string filename, CameraParams intrinsic_params, Mat extrinsic_
 	}
 	_capture_lock->lock();
 	bool result = this->_capture->open(filename);
+	_capture->set(cv::CAP_PROP_FOURCC, cv::VideoWriter::fourcc('M', 'J', 'P', 'G'));
+	_capture->set(cv::CAP_PROP_FRAME_WIDTH, 640);
+	_capture->set(cv::CAP_PROP_FRAME_HEIGHT, 480);
```
abhaybd, line 60: Remove, this stuff isn't respected by gstreamer

src/camera/Camera.cpp
abhaybd: Refactor as necessary to support these changes

src/camera/Camera.cpp
```diff
@@ -115,23 +136,26 @@ bool Camera::openFromConfigFile(std::string filename) {
...
+			cv::Ptr<cv::aruco::Dictionary> dictionary =
+				cv::aruco::getPredefinedDictionary(cv::aruco::DICT_4X4_100);
+			std::vector<std::vector<cv::Point2f>> markerCorners;
+			cv::Mat frameCopy;
+			std::vector<int> markerIds;
+			cv::aruco::detectMarkers(frame, dictionary, markerCorners, markerIds);
+			frame.copyTo(frameCopy);
+			if (!markerIds.empty()) {
+				cv::aruco::drawDetectedMarkers(frameCopy, markerCorners, markerIds);
+			}
```
abhaybd, line 155: This code does aruco tag detection (not localization!). We default to doing it on every frame of every camera, but maybe that should be configurable? Or maybe detection shouldn't be in the same file as camera reading. Maybe overhaul the entirety of the `AR` codebase and do detection/localization in there. In any case, this is a code sample that works so use as you see fit.

src/control/PlanarArmController.h
```diff
@@ -210,12 +210,15 @@ class PlanarArmController {
...
 									const navtypes::Vectord<N>& currJointPos) {
+		LOG_SCOPE_F(INFO, "IK getCommand");
 		Eigen::Vector2d newPos = get_setpoint(currTime);
 		// lock after calling get_setpoint since that internally locks the mutex
 		std::lock_guard<std::mutex> lock(mutex);
 
 		// get new joint positions for target EE
 		bool success = false;
+		LOG_F(INFO, "newPos=[%.2f, %.2f]", newPos(0), newPos(1));
+		LOG_F(INFO, "currJP=[%.2f, %.2f]", currJointPos(0), currJointPos(1));
```
abhaybd, line 221: Extraneous logging, remove

src/control/SwerveController.cpp
```diff
@@ -85,6 +85,7 @@ drive_commands_t SwerveController::setCrabCmdVel(double dtheta, double dy,
 
 bool SwerveController::checkWheelRotation(DriveMode mode,
 										  const swerve_rots_t& wheel_rots) const {
+    return true;
```
abhaybd, line 88: Hack due to broken swerve motors, remove after repair.

src/control_interface.cpp
```diff
@@ -219,6 +221,10 @@ void setJointMotorPower(robot::types::jointid_t joint, double power) {
 			Globals::wristKinematics.jointPowerToGearPower(jointPwr);
 		setMotorPower(motorid_t::wristDiffLeft, gearPwr.left);
 		setMotorPower(motorid_t::wristDiffRight, gearPwr.right);
+	} else if (joint == jointid_t::handActuator) {
+		if (std::abs(power) > 0.1) {
+			setMotorPos(motorid_t::handActuator, power > 0 ? 1 : 0);
+		}
```
abhaybd, line 227: This code just sets the hand linear actuator in or out. The firmware of the actuator board (id 0x16) is configured to accept position 0 as retracted and anything else as extended.

src/network/MissionControlProtocol.cpp
```diff
@@ -346,9 +343,6 @@ void MissionControlProtocol::handleConnection() {
 }
 
 void MissionControlProtocol::handleHeartbeatTimedOut() {
-	LOG_F(ERROR, "Heartbeat timed out! Emergency stopping.");
-	this->stopAndShutdownPowerRepeat(true);
-	robot::emergencyStop();
```
abhaybd, line 351: Hack due to bad connection, remove.

src/network/MissionControlTasks.cpp
```diff
@@ -127,28 +127,37 @@ CameraStreamTask::CameraStreamTask(websocket::SingleClientWSServer& server)
...
+			auto it = Constants::video::STREAM_RFS.find(cam);
+			int rf = it != Constants::video::STREAM_RFS.end()
+						 ? it->second
+						 : Constants::video::H264_RF_CONSTANT;
+			auto enc = std::make_shared<video::H264Encoder>(fps, rf);
+			auto cam_handle = robot::openCamera(cam);
+			if (cam_handle) {
+				_open_streams.insert_or_assign(cam, stream_data_t(enc, cam_handle));
+			}
+		}).detach();
```
abhaybd, line 143: The jetson can only open 2 cameras at a time (due to usb bandwidth issues? not sure) so this code opens cameras if and when we need them.

Reference counting makes sure that when we drop all references to an open camera it is automatically closed.

src/network/MissionControlTasks.h
```diff
@@ -92,11 +93,19 @@ class CameraStreamTask : public util::AsyncTask<> {
 	void task(std::unique_lock<std::mutex>& lock) override;
 
 private:
+	struct stream_data_t {
+		std::shared_ptr<video::H264Encoder> encoder;
+		std::shared_ptr<robot::types::CameraHandle> cam_handle;
```
abhaybd, line 98: The camera handle is a reference to the open camera. When all references get cleaned up, the camera is closed, allowing other ones to open.

src/network/websocket/WebSocketProtocol.cpp
```diff
@@ -65,9 +65,6 @@ void WebSocketProtocol::clientDisconnected() {
 }
 
 void WebSocketProtocol::heartbeatTimedOut() {
-	if (heartbeatInfo.has_value()) {
-		heartbeatInfo->second();
-	}
```
abhaybd, line 70: Hack, remove.

src/world_interface/real_world_constants.h
```diff
@@ -99,14 +100,16 @@ constexpr auto motorSerialIDMap = frozen::make_unordered_map<motorid_t, can::dev
 	 {motorid_t::shoulder, DEVICE_SERIAL_MOTOR_SHOULDER},
 	 {motorid_t::elbow, DEVICE_SERIAL_MOTOR_ELBOW},
 	 {motorid_t::forearm, DEVICE_SERIAL_MOTOR_FOREARM},
-	 {motorid_t::wristDiffLeft, DEVICE_SERIAL_MOTOR_WRIST_DIFF_LEFT},
-	 {motorid_t::wristDiffRight, DEVICE_SERIAL_MOTOR_WRIST_DIFF_RIGHT},
+	 {motorid_t::wristDiffLeft, 0xc},
+	 {motorid_t::wristDiffRight, 0x5},
 	 {motorid_t::hand, DEVICE_SERIAL_MOTOR_HAND},
+	 {motorid_t::handActuator, 0x16},
```
abhaybd, line 106: Hardcoded serial IDs, reconcile with HindsightCAN

src/world_interface/real_world_constants.h
```diff
@@ -99,14 +100,16 @@ constexpr auto motorSerialIDMap = frozen::make_unordered_map<motorid_t, can::dev
...
 	 {motorid_t::activeSuspension, DEVICE_SERIAL_LINEAR_ACTUATOR}});
 
 /** @brief A mapping of PID controlled motors to their pid coefficients. */
 constexpr auto motorPIDMap = frozen::make_unordered_map<motorid_t, pidcoef_t>(
-	{{motorid_t::shoulder, {70, 0, 0}},
+	{{motorid_t::shoulder, {200, 0, 0}},
+	 {motorid_t::elbow, {250, 0, 0}},
 	 {motorid_t::frontLeftSwerve, {2, 0, 0}},
 	 {motorid_t::frontRightSwerve, {2, 0, 0}},
 	 {motorid_t::rearLeftSwerve, {2, 0, 0}},
```
abhaybd, line 115: Major inconsistency with PID on swerve and non-swerve boards, related to PID constant scaling. Non-swerve boards have it correct, swerve boards need to fix how they scale PID power output. This is a firmware issue, and after it's fixed the swerve boards should be retuned. The associated PID constants will be much higher (in the 100s)

abhaybd: Notably, PID constants used to be specified in deci-units (i.e. divided by 10) now they will be specified in milli-units (divided by 1000). That is, if you want to set kP=1, you used to send the value 10, now you should use 1000.

src/world_interface/real_world_interface.cpp
```diff
@@ -53,7 +53,7 @@ const kinematics::DiffDriveKinematics& driveKinematics() {
 
 namespace {
 // map that associates camera id to the camera object
-std::unordered_map<CameraID, std::shared_ptr<cam::Camera>> cameraMap;
+std::unordered_map<CameraID, std::weak_ptr<cam::Camera>> cameraMap;
```
abhaybd, line 56: Maintain a weak pointer here so reference counting works. When the number of strong references (all held by client code) drops to 0 , the camera will be cleaned up.

src/world_interface/real_world_interface.cpp
```diff
@@ -84,8 +84,12 @@ void initMotors() {
...
 		can::motor::initEncoder(serial, enc_params.isInverted, true, enc_params.ppjr,
 								TELEM_PERIOD);
-		can::motor::setLimitSwitchLimits(serial, enc_params.limitSwitchLow,
-										 enc_params.limitSwitchHigh);
+		auto mDegToPulses = [&](int32_t mdeg) {
+			return static_cast<int32_t>((double)mdeg / 360000.0 * enc_params.ppjr);
+		};
+		// inverted because of weirdness
+		can::motor::setLimitSwitchLimits(serial, mDegToPulses(-enc_params.limitSwitchLow),
+										 mDegToPulses(-enc_params.limitSwitchHigh));
```
abhaybd, line 92: This is due to some quaddec inversion weirdness in firmware that I don't fully understand. Ask Casey and/or Michael. This should be fixed in firmware and we shouldn't have to negate the parameters.

src/world_interface/real_world_interface.cpp
```diff
@@ -96,33 +100,35 @@ void initMotors() {
 	}
 }
 
-void openCamera(CameraID camID, const char* cameraPath) {
+std::shared_ptr<cam::Camera> openCamera(CameraID camID, const std::string& cameraPath) {
```
abhaybd, line 103: As I said, the Jetson can only open 2 cameras, so we're opening cameras as necessary (instead of at the beginning) and using reference counting to figure out when to close it. This involves weak and shared pointers, which should be pretty intuitive.

src/world_interface/world_interface.h
```diff
@@ -12,17 +14,20 @@
...
  * @brief Collection of functions that allows interfacing with the hardware and the world.
  */
 namespace robot {
+namespace types {
+class CameraHandle {
+public:
+	CameraHandle(std::shared_ptr<cam::Camera> cam) : camera(cam) {}
+private:
+	std::shared_ptr<cam::Camera> camera;
+};
```
abhaybd, line 29: This is a dumb class that is used for reference counting without letting client code have access to the actual camera object. Is this the right place for it?

src/world_interface/world_interface.h
```diff
@@ -87,6 +92,8 @@ bool isEmergencyStopped();
  */
 std::unordered_set<types::CameraID> getCameras();
 
+std::shared_ptr<types::CameraHandle> openCamera(types::CameraID camera);
```
abhaybd, line 95: Document!

src/Constants.cpp
```diff
@@ -47,15 +47,19 @@ const double MAX_WHEEL_VEL = WHEEL_RADIUS * MAX_DRIVE_PWM / PWM_PER_RAD_PER_SEC;
...
+const robot::types::CameraID PANO_CAMERA_ID = "pano";
+const robot::types::CameraID DRILL_CAMERA_ID = "drill";
+
+const std::unordered_map<robot::types::CameraID, std::string> CAMERA_CONFIG_PATHS = {
+   {MAST_CAMERA_ID, "../camera-config/MastCameraCalibration.yml"},
+   {FOREARM_CAMERA_ID, "../camera-config/WristCameraCalibration.yml"},
+   {HAND_CAMERA_ID, "../camera-config/HandCameraCalibration.yml"},
+   {PANO_CAMERA_ID, "../camera-config/PanoCameraCalibration.yml"},
+   {DRILL_CAMERA_ID, "../camera-config/DrillCameraCalibration.yml"}
+};
```
Geeoon, line 62: Useless, as mentioned for PanoCameraCalibration.yml

src/world_interface/simulator_interface.cpp
```diff
@@ -290,6 +290,10 @@ bool isEmergencyStopped() {
 	return is_emergency_stopped;
 }
 
+std::shared_ptr<types::CameraHandle> openCamera(CameraID cameraID) {
+	return nullptr;
+}
```
Geeoon, line 295: Why is the old openCamera function returning nullptr?

huttongrabiel:
closing for abhays mental health

PR #340, opened 2024-11-05
Better test coverage

ricemaking:
running into an error when attempting to 'make tests'

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -337,6 +337,7 @@ if(WITH_TESTS)
     ../tests/util/CoreTest.cpp
     ../tests/util/DataTest.cpp
     ../tests/util/MathTest.cpp
+    ../tests/util/RandomTest.cpp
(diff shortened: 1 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -337,14 +337,18 @@ if(WITH_TESTS)
...
 	# Protocol/teleop tests
 	../tests/kinematics/DiffWristKinematicsTest.cpp)
 
   target_link_libraries(tests
     ${rover_libs}
-    stub_world_interface
+    stub_world_interface    double sample0 = stdn(0);
+    double sample1 = stdn(1);
+
+    REQUIRE(sample0 != sample1);
```
huttongrabiel, line 351: This is the reason CMake isn't working. It looks like some code may have accidentally been written here. Remove it and it should work

huttongrabiel (approved):
nice work!

PR #341, opened 2024-11-05
Change 300ms time delay to a while loop

abnv-goyal:


src/TunePID.cpp
```diff
@@ -137,8 +137,21 @@ int main(int argc, char** argv) {
 
 	double period = 8.0;
 
-	std::this_thread::sleep_for(300ms); // wait for encoder position data to arrive
-	int32_t starting_angle = can::motor::getMotorPosition(serial);
+	time_point<steady_clock> start = steady_clock::now();
+	int timeout = 300; // in milliseconds
```
huttongrabiel, line 141: Quick couple of things, since this `int` is constant it should really be `constexpr int timeout = 300ms`. Note the ms at the end so you don't have to specify it in comments.

src/TunePID.cpp
```diff
@@ -137,8 +137,21 @@ int main(int argc, char** argv) {
...
+	while (!starting_angle_data_point.isValid() &&
+		   steady_clock::now() - start < milliseconds(timeout)) {
+		starting_angle_data_point = can::motor::getMotorPosition(serial);
+	}
+
+	int32_t starting_angle;
+	if (starting_angle_data_point.isValid()) {
+		starting_angle = starting_angle_data_point.getData();
+	} else {
+		LOG_F(WARNING, "STARTING ANGLE DATA NOT FOUND");
```
huttongrabiel, line 152: I'm realizing that I said to log here but you can probably just exit with an error. This is the TunePID executable which isn't mission critical.

huttongrabiel: but still log

src/TunePID.cpp
```diff
@@ -137,8 +137,21 @@ int main(int argc, char** argv) {
...
+	int timeout = 300; // in milliseconds
+	DataPoint<int32_t> starting_angle_data_point = can::motor::getMotorPosition(serial);
+	while (!starting_angle_data_point.isValid() &&
+		   steady_clock::now() - start < milliseconds(timeout)) {
+		starting_angle_data_point = can::motor::getMotorPosition(serial);
+	}
+
+	int32_t starting_angle;
+	if (starting_angle_data_point.isValid()) {
+		starting_angle = starting_angle_data_point.getData();
```
huttongrabiel, line 150: also just do `int32_t starting_angle = ...` here instead of declaring it above. It's ok in this case since we will fail if we don't enter this block.

huttongrabiel (commented):
Really good! Just a couple minor things

huttongrabiel (changes requested):
Nice! Just a few things to fix

src/TunePID.cpp
```diff
@@ -137,8 +139,19 @@ int main(int argc, char** argv) {
...
-	int32_t starting_angle = can::motor::getMotorPosition(serial);
+	time_point<steady_clock> start = steady_clock::now();
+	constexpr milliseconds timeout = 300ms;
+	DataPoint<int32_t> starting_angle_data_point = can::motor::getMotorPosition(serial);
+	while (!starting_angle_data_point.isValid() && steady_clock::now() - start < timeout) {
+		starting_angle_data_point = can::motor::getMotorPosition(serial);
+	}
+
+	if (!starting_angle_data_point.isValid()) {
+		LOG_F(WARNING, "STARTING ANGLE DATA NOT FOUND");
```
huttongrabiel, line 150: can you make this `LOG_F(ERROR...)`

huttongrabiel (changes requested):

src/TunePID.cpp
```diff
@@ -17,8 +17,10 @@ extern "C" {
 #include "HindsightCAN/CANCommon.h"
 }
 
-enum class targetmode_t {
-	step, sinusoidal
+enum class targetmode_t
+{
```
huttongrabiel: can you move this onto the same line `enum class targetmode_t`? i thought clang-format would catch this but ig not

src/TunePID.cpp
```diff
@@ -159,7 +172,7 @@ int main(int argc, char** argv) {
 		if (mode == targetmode_t::step) {
 			prescaled_target = round(prescaled_target);
 		}
-		angle_target = (int32_t) round(amplitude * prescaled_target) + starting_angle;
+		angle_target = (int32_t)round(amplitude * prescaled_target) + starting_angle;
```
huttongrabiel: Can you use a `static_cast` instead of a C-style cast?

huttongrabiel (changes requested):

src/TunePID.cpp
abnv-goyal: so should it be like `enum class targetmode_t {step, sinusoidal);` ?

abnv-goyal: Oh you meant the left bracket?

huttongrabiel: i want it to be like
```c++
enum class targetmode_t {
   step,
   sinusoidal
};
```

but maybe that isn't what clang-format wants

PR #342, opened 2024-11-05
Remove c-style casts from codebase

jazonshou:
Not sure if I got all of them but I think I got most


huttongrabiel (approved):
beautiful. sorry for the mundane pr but it looks like you got the hang of interacting with the code base

huttongrabiel:
Also, there is merge conflicts so you'll have to handle those and then commit again. After than I can approve and merge :)

PR #343, opened 2024-11-08
Elevation data

PatrickRung:
Changed the GPS coords to contain altitude data and for the data from the simulator to take in altitude data and it sends over the altitude data as well to the mission control.

### DIFF
--- a/src/gps/usb_gps/read_usb_gps.cpp
+++ b/src/gps/usb_gps/read_usb_gps.cpp
@@ -60,15 +60,14 @@ void gps_loop() {
 			} else {
 				double lat = newdata->fix.latitude;
 				double lon = newdata->fix.longitude;
-				double alt = newdata->fix.altitude;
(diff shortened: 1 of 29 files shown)


huttongrabiel:
Make sure to run clang-format on your code. Then it'll pass CI.

src/gps/usb_gps/read_usb_gps.cpp
huttongrabiel, line 1: This file is obsolete so changes here are not required.

src/navtypes.h
```diff
@@ -12,6 +12,8 @@ struct gpscoords_t {
 	double lat;
 	/** the longitude of the gps coordinate, in degrees */
 	double lon;
+	/** the altitude of the gps coordinate */
```
huttongrabiel, line 15: See https://github.com/huskyroboticsteam/ardupilot-driver/pull/2/files and update this comment to specify the unit of data

huttongrabiel (changes requested):

huttongrabiel (approved):

PR #345, opened 2024-11-19
add readIMUHeading method in dummy_gps

jazonshou:
Not 100% sure if this is what you were talking about


src/gps/dummy/dummy_gps.cpp
```diff
@@ -5,3 +5,7 @@ using robot::types::DataPoint;
 DataPoint<navtypes::gpscoords_t> readGPSCoords() {
 	return {};
 }
+
+DataPoint<navtypes::eulerangles_t> readIMUHeading() {
```
huttongrabiel, line 9: This should be `DataPoint<double>`.

huttongrabiel (changes requested):

huttongrabiel (approved):

PR #346, opened 2024-11-22
Add code to ping motors on startup

abnv-goyal:


src/control_interface.cpp
```diff
@@ -168,6 +172,33 @@ types::DataPoint<int32_t> getJointPos(robot::types::jointid_t joint) {
...
+			return true;
+		}
+		std::this_thread::sleep_for(std::chrono::milliseconds(10));
+	}
+	return false;
+}
+
+void verifyAllMotorsConnected() {
+	for (auto motorMap : robot::motorSerialIDMap) {
+		can::pullDeviceTelemetry(std::make_pair(can::devicegroup_t::motor, motorMap.second),
```
huttongrabiel, line 192: You do `std::make_pair(can::devicegroup_t::motor, motorMap.second)` twice. Can you pull it out into a variable at the top and then pass it in?

src/control_interface.cpp
```diff
@@ -168,6 +172,33 @@ types::DataPoint<int32_t> getJointPos(robot::types::jointid_t joint) {
...
+								 can::telemtype_t::voltage);
+		if (!hasData(std::make_pair(can::devicegroup_t::motor, motorMap.second))) {
```
huttongrabiel, line 194: Second time doing `std::make_pair(can::devicegroup_t::motor, motorMap.second)` is here

huttongrabiel (changes requested):
This is a good start! I think we need to talk about if this is the right location for this function and then also implement calling it on start. Nice work!

huttongrabiel (commented):
nice!

take a look at rover.cpp and i think you can call this function in the start() function in there and it will ensure everything it set up correctly.

huttongrabiel:
also merge master into this branch so youre up to date

src/CAN/CANUtils.cpp
```diff
@@ -46,4 +49,32 @@ std::string packetToString(const CANPacket& packet) {
...
+
+void verifyAllMotorsConnected(
+	frozen::unordered_map<robot::types::motorid_t, deviceserial_t, 18UL> motor_id_map) {
+	for (auto motorMap : motor_id_map) {
+		auto motor_pair = std::make_pair(devicegroup_t::motor, motorMap.second);
+		pullDeviceTelemetry(motor_pair, telemtype_t::voltage);
+		if (!hasData(motor_pair)) {
+			LOG_F(ERROR, "Motor not connected!\nID: %s",
+				  std::to_string(motorMap.second).c_str());
+			throw std::runtime_error("Not all motors connected!");
```
imisaacwu, line 75: You could consider removing just this line for the function to go through all the pairs & not throw an error, but it's up to you how you want to implement it. We just want to be able to run this function and see a list of errors with the motors that aren't connected. It may also be nice to do an info log with "Motors are all connected!" if they are all connected

imisaacwu (dismissed):
lgtm, just needs testing!

PR #347, opened 2025-01-14
use frozen

clr-li:


huttongrabiel:
@clr-li It looks like this isn't building. Can you figure out why, fix, and then update this PR? It looks like frozen isn't being used correctly.

src/utils/json.cpp
```diff
@@ -11,12 +12,12 @@ bool validateKey(const json& j, const std::string& key, const val_t& type) {
 }
 
 bool validateKey(const json& j, const std::string& key,
-				 const std::unordered_set<val_t>& types) {
+				 const frozen::unordered_set<val_t>& types) {
```
huttongrabiel, line 15: It looks like this isn't quite the right usage. It seems like it wants a size being specified to it so maybe frozen is not the right use case here. Can you take a look into if that is correct that it wants a capacity?

src/utils/json.cpp
```diff
@@ -11,12 +12,12 @@ bool validateKey(const json& j, const std::string& key, const val_t& type) {
...
 	return hasKey(j, key) && types.find(j.at(key).type()) != types.end();
 }
 
 bool validateOneOf(const json& j, const std::string& key,
-				   const std::unordered_set<std::string>& vals) {
+				   const frozen::unordered_set<std::string>& vals) {
```
huttongrabiel, line 20: Same here

src/utils/json.cpp
clr-li: Yeah I checked the documentation and it seems like it does require a size. However after looking at validateKey's references one of them uses a set of size 2 and the other uses a set of size 3.

huttongrabiel: What do you think about templating it?

huttongrabiel:
Closing since we agreed that frozen wasn't necessary for this given that frozen wants the set to be of a known capacity.

PR #350, opened 2025-01-25
InstrumentationBox

PatrickRung:
Added the drill four bar linkage motor references and ID's.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/HindsightCAN"]
```
imisaacwu, line 1: Remove the submodule

imisaacwu (changes requested):
Will need to remove the HindsighCAN submodule, possibly rebase this branch onto master

PR #354, opened 2025-02-08
Insumentation box fixed

PatrickRung:
Added Instrumentation box four bar linkage motors.


.gitmodules
```diff
@@ -0,0 +1 @@
+
```
imisaacwu, line 1: remove this file

src/world_interface/real_world_constants.h
```diff
@@ -104,7 +104,9 @@ constexpr auto motorSerialIDMap = frozen::make_unordered_map<motorid_t, can::dev
 	 {motorid_t::hand, DEVICE_SERIAL_MOTOR_HAND},
 	 {motorid_t::activeSuspension, DEVICE_SERIAL_LINEAR_ACTUATOR},
 	 {motorid_t::drillActuator, DEVICE_SERIAL_DRILL_ARM_MOTOR},
-	 {motorid_t::drillMotor, DEVICE_SERIAL_DRILL_MOTOR}});
+	 {motorid_t::drillMotor, DEVICE_SERIAL_DRILL_MOTOR},
+	 {motorid_t::fourBarLinkage1, DEVICE_SERIAL_DRILL_FOUR_BAR_LINKAGE1},
```
imisaacwu, line 108: Super minimal, not sure if we want to classify the four bar linkages as a "drill" thing, but we can change this pretty easily if we really want to. This totally works for now though

imisaacwu (changes requested):
2 super small changes, looks great!

PR #356, opened 2025-02-23
Use the correct device groups

imisaacwu:


imisaacwu:
This includes the drill_control branch that implements the `motorPowerRequest` packet

PatrickRung:
For the simulator_interface there needs ot be mappings for fourbar1, and fourbar2.

PatrickRung (approved):
Looks good!

PR #363, opened 2025-05-03
Rework Cameras to use GStreamer Pipelines

imisaacwu:
Aims to simplify our camera library, allows for easier configuration of camera streams

### DIFF
--- a/src/CAN/CAN.cpp
+++ b/src/CAN/CAN.cpp
@@ -279,9 +279,9 @@ void sendCANPacket(const CANPacket& packet) {
 	}
 
 	if (!success) {
-		LOG_F(ERROR, "Failed to send CAN packet to group=%x, id=%x: %s",
-			  static_cast<int>(getDeviceGroup(packet)), getDeviceSerial(packet),
-			  std::strerror(errno));
+		// LOG_F(ERROR, "Failed to send CAN packet to group=%x, id=%x: %s",
+		// 	  static_cast<int>(getDeviceGroup(packet)), getDeviceSerial(packet),
+		// 	  std::strerror(errno));
 	}
 }
 
--- a/src/Constants.cpp
+++ b/src/Constants.cpp
@@ -52,9 +52,9 @@ const robot::types::CameraID WRIST_CAMERA_ID = 30;
 const robot::types::CameraID MAST_CAMERA_ID = 40;
 const robot::types::CameraID RAND_CAMERA_ID = 50;
 const robot::types::CameraID CJ1_CAMERA_ID = 510;
-const robot::types::CameraID CJ2_CAMERA_ID = 520;
+const robot::types::CameraID BOX_CAMERA_ID = 520;
 const robot::types::CameraID CJ3_CAMERA_ID = 530;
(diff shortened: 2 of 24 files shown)


src/Constants.cpp
```diff
@@ -47,15 +47,25 @@ const double MAX_WHEEL_VEL = WHEEL_RADIUS * MAX_DRIVE_PWM / PWM_PER_RAD_PER_SEC;
...
-const char* HAND_CAMERA_CONFIG_PATH = "../camera-config/HandCameraCalibration.yml";
-const robot::types::CameraID HAND_CAMERA_ID = "hand";
+const robot::types::CameraID HAND_CAMERA_ID = 20;
+const robot::types::CameraID WRIST_CAMERA_ID = 30;
+const robot::types::CameraID MAST_CAMERA_ID = 40;
+const robot::types::CameraID RAND_CAMERA_ID = 50;
+const robot::types::CameraID CJ1_CAMERA_ID = 510;
+const robot::types::CameraID CJ2_CAMERA_ID = 520;
+const robot::types::CameraID CJ3_CAMERA_ID = 530;
+const robot::types::CameraID CJ4_CAMERA_ID = 540;
```
Geeoon, line 57: We should document this somewhere on the mission control repo and possibly add instructions for adding new cameras.

src/CAN/CAN.cpp
```diff
@@ -279,9 +279,9 @@ void sendCANPacket(const CANPacket& packet) {
 	}
 
 	if (!success) {
-		LOG_F(ERROR, "Failed to send CAN packet to group=%x, id=%x: %s",
-			  static_cast<int>(getDeviceGroup(packet)), getDeviceSerial(packet),
-			  std::strerror(errno));
+		// LOG_F(ERROR, "Failed to send CAN packet to group=%x, id=%x: %s",
+		// 	  static_cast<int>(getDeviceGroup(packet)), getDeviceSerial(packet),
+		// 	  std::strerror(errno));
```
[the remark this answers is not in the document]
Geeoon: Is this still needed?

src/network/MissionControlTasks.cpp
```diff
@@ -81,28 +81,43 @@ CameraStreamTask::CameraStreamTask(websocket::SingleClientWSServer& server)
...
+	}).detach();
 }
 
 void CameraStreamTask::task(std::unique_lock<std::mutex>&) {
 	while (isRunningInternal()) {
 		{
 			std::lock_guard lg(_mutex);
 			// for all open streams, check if there is a new frame
-			for (const auto& stream : _open_streams) {
+			for (auto& stream : _open_streams) {
```
Geeoon, line 117: Is there any reason why this cant stay const?

Geeoon (changes requested):
Just look at the comments, everything else looks great.

src/network/MissionControlTasks.cpp
imisaacwu: Yeah, down below when we call `stream_data_t& stream_data = stream.second;`, .second isn't a const property, so we can't have stream be const

PR #371, opened 2025-06-22
Fix Simulator Interface + Optimize CMake

imisaacwu:
Added unimplemented world interface methods
Edited CMake to only build libraries relevant to the specified world interface

### DIFF
--- a/src/world_interface/simulator_interface.cpp
+++ b/src/world_interface/simulator_interface.cpp
@@ -384,17 +384,16 @@ void setMotorVel(robot::types::motorid_t motor, int32_t targetVel) {
 }
 
 void setServoPos(robot::types::servoid_t servo, int32_t position) {
-
+	// Implement when Servos are added to Simulator
 }
 
 void setRequestedStepperTurnAngle(robot::types::stepperid_t stepper, int16_t angle) {
-
+	// Implement when Steppers are added to Simulator
 }
 
 void setActuator(uint8_t value) {
-
+	// Implement when Actuators are added to Simulator
 }
-
 
 callbackid_t addLimitSwitchCallback(
 	robot::types::motorid_t motor,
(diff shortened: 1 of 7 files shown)


src/world_interface/simulator_interface.cpp
```diff
@@ -114,14 +110,6 @@ static std::shared_ptr<robot::types::CameraHandle> openCamera(CameraID cam, std:
 	return nullptr;
 }
 
-void initCameras() {
```
imisaacwu, line 117: Because of how we have the Cameras setup now where MissionControl will have to request which streams they want open, we don't need this `initCameras`

src/world_interface/simulator_interface.cpp
```diff
@@ -392,6 +383,19 @@ void setMotorVel(robot::types::motorid_t motor, int32_t targetVel) {
...
+
+}
+
+void setRequestedStepperTurnAngle(robot::types::stepperid_t stepper, int16_t angle) {
+
+}
+
+void setActuator(uint8_t value) {
+
+}
```
imisaacwu, line 396: Need to implement servos, steppers, and actuators in the sim first

src/world_interface/simulator_interface.cpp
```diff
@@ -298,6 +285,10 @@ std::unordered_set<CameraID> getCameras() {
 	return util::keySet(cameraLastFrameIdxMap);
 }
 
+std::shared_ptr<robot::types::CameraHandle> openCamera(CameraID cam) {
+	return nullptr;
+}
+
```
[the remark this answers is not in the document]
PatrickRung: I believe that we need to use the code in the openCamera method in the namespace above, so that we tell the simulator to send the correct camera frames. I did some testing with this and ran into issues with the data being sent between the sim and Resurgence. However I may have tested camera incorrectly (I opened up everything as normal and then pressed the open button next to camera stream on mission control)?

src/world_interface/simulator_interface.cpp
PatrickRung: Looks good!

PR #372, opened 2025-06-29
Fix simulator camera stream

PatrickRung:


src/world_interface/simulator_interface.cpp
```diff
@@ -302,8 +302,9 @@ std::shared_ptr<robot::types::CameraHandle> openCamera(CameraID cam) {
 			{"height", static_cast<int>(fs[KEY_IMAGE_HEIGHT])},
 			{"intrinsics", nullptr}};
 	sendJSON(msg);
-
-	return nullptr;
+	auto camObject = std::make_shared<cam::Camera>();
```
PatrickRung, line 305: I was thinking for this camera object should we try to give it more information about camera name or description or can we leave it as is since it is a fake camera object. Let me know if u have any thoughts on a better way to implement this : )

imisaacwu (approved):
lgtm, if it runs as intended, go ahead and merge👍

